PR #175, opened 2014-04-16
Refactor node.rb template

igreg:
Please provide feedback only.
This PR just include the new template.

I spend a couple of days tidying up the code in external_node.rb.
The aim was to clean up some code that was hard to understand and make the file more maintainable.
Some command line options have been altered although it is still very much backward compatible.
And since the file is also linked to from various web pages, I have provided a way to configure it without altering the source code so it's easier to upgrade to a more recent version of the script manually.

* Refactor / clean up existing code
* Store configuration in `/etc/puppet/foreman.yaml` instead of requiring changes to the template itself
* Add command line help
* Replace `--push-facts-parallel` with `--push-facts --parallel`
* Use the gem `rb-inotify` to watch file changes (simpler to use)

### DIFF
--- a/templates/external_node_v3.rb.erb
+++ b/templates/external_node_v3.rb.erb
@@ -1,0 +1,453 @@
+#!/usr/bin/env ruby
+
+# Script usually acts as an ENC for a single host, with the certname supplied as argument
+#   if 'facts' is true, the YAML facts for the host are uploaded
+#   ENC output is printed and cached
+#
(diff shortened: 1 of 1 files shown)


ekohl:
So a long time has passed. Should we revisit this?

igreg:
Indeed, that was a long time ago!
I'm no longer using Foreman so can't contribute anymore to this pull request.

ekohl:
Thanks for your effort anyway.

PR #182, opened 2014-04-30
Add puppet-syntax and future parser tests

domcleal:
There are a couple of strange failures to look into.


ekohl:
Now rspec-puppet 2.0 has been released, think we could rebase this?

ekohl:
Since puppetlabs_spec_helper now already solves these problems and in #284 I intend to use those, I'm closing this one.

PR #196, opened 2014-06-08
Add beaker tests

ekohl:


domcleal:
Sorry for not getting to this at all, but maybe we should do it?  I hadn't noticed that it was using the Docker support, which is handy.

ekohl:
I suppose we could have a go at it again. I'll try to see how much of it still applies.

Issue #221, opened 2014-08-28
Ability to increase max pool size in ActiveRecord?


mstrzele:
There is no option to set size of connection pool. Currently it's set to 5 (default value). Under heavy load, using the PostgreSQL adapter, I often get the following error:

```
ActiveRecord::ConnectionTimeoutError (could not obtain a database connection within 5 seconds (waited 5.000178992 seconds). The max pool size is currently 5; consider increasing it.):
  /usr/lib/ruby/gems/1.8/gems/passenger-4.0.18/lib/phusion_passenger/rack/thread_handler_extension.rb:77:in `process_request'
  /usr/lib/ruby/gems/1.8/gems/passenger-4.0.18/lib/phusion_passenger/request_handler/thread_handler.rb:140:in `accept_and_process_next_request'
  /usr/lib/ruby/gems/1.8/gems/passenger-4.0.18/lib/phusion_passenger/request_handler/thread_handler.rb:108:in `main_loop'
  /usr/lib/ruby/gems/1.8/gems/passenger-4.0.18/lib/phusion_passenger/request_handler.rb:441:in `block (3 levels) in start_threads'
```

PR #222, opened 2014-08-28
add pool parameter

mstrzele:
closes GH-221

### DIFF
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -78,6 +78,9 @@
 #
 # $db_sslmode::               Database 'production' ssl mode
 #
+# $db_pool::                  Database 'production' size of connection pool
+#                             type:integer
+#
 # $app_root::                 Name of foreman root directory
 #
 # $user::                     User under which foreman will run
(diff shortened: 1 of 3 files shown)


mmoll:
After some googling, I think sqlite and mysql2 can also use pools in Rails and we should just set the default to 5 and insert that value for all DB types in database.yml.
Also, this needs a rebase. Ping @mstrzele

mstrzele:
Rebase done. I've added a default value for connection pool.

manifests/init.pp
```diff
@@ -78,6 +78,8 @@
 #
 # $db_sslmode::               Database 'production' ssl mode
 #
+# $db_pool::                  Database 'production' size of connection pool
+#
```
mmoll: please add ```type:integer``` here, as seen for db_port

mmoll:
Apart from the nitpick: tested, works for me, :+1:.
@mstrzele Please squash the commits together into one.

mstrzele:
OK, all commits are squashed into a single commit.

ekohl:
:+1: 

/cc @domcleal

mmoll:
:+1:

domcleal:
Thanks @mstrzele!

PR #235, opened 2014-10-01
Add parameters to foreman::config::puppetdb

ekohl:
This replaces https://github.com/theforeman/puppet-foreman/pull/184 since I no longer have access to that repository.

### DIFF
--- a/manifests/plugin/puppetdb.pp
+++ b/manifests/plugin/puppetdb.pp
@@ -1,5 +1,15 @@
 # Installs puppetdb_foreman plugin
-class foreman::plugin::puppetdb {
+#
+# === Parameters:
+#
+# $enabled::                Whether to enable the plugin
+#
+# $address::                URL of the puppetdb plugin should use to connect
+#
+class foreman::plugin::puppetdb(
+  $enabled = true,
+  $address = "https://puppetdb.${::domain}:8081/v2/commands",
+) {
   case $::osfamily {
     'RedHat': {
       case $::operatingsystem {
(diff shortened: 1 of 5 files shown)


ekohl:
Rebased. Also includes #281 which should be merged first.

manifests/plugin/puppetdb.pp
```diff
@@ -1,5 +1,15 @@
 # Installs puppetdb_foreman plugin
-class foreman::plugin::puppetdb {
+#
+# === Parameters:
+#
+# $enabled::                Whether to enable the plugin
```
domcleal: type:boolean, and validate it

manifests/plugin/puppetdb.pp
```diff
@@ -1,5 +1,15 @@
...
+#
+# $address::                URL of the puppetdb plugin should use to connect
```
domcleal: "URL of the PuppetDB instance" perhaps?

templates/puppetdb_plugin.yaml.erb
```diff
@@ -0,0 +1,4 @@
+---
+:puppetdb:
```
domcleal: Check the format of this file.. there's the old hash style used by the plugin which is discouraged, and there's the new style which is DB settings.  The latter would be outside of the 'puppetdb' hash, just top-level puppetdb_enabled/address.

domcleal:
I like the look of it!

ekohl:
Closed in favour of https://github.com/theforeman/puppet-foreman/issues/383

PR #242, opened 2014-10-21
Add foreman enc subdirs with modes

mhjacks:
Here is a followup pull request.

I am specifying the mode because on my test Ubuntu 14.04 system, those directories are created 755.

I cannot find in defaults.rb where any subdir of $yamldir is managed, so all the ones needed by foreman's enc (node,facts,foreman) functionality are specified.

Thanks for your consideration.


domcleal:
I just realised there was a whole reply in @mhjacks' last comment that GitHub hid, thinking it was quoted material.

The thing I'm worried about is that Puppet itself manages this directory: https://github.com/puppetlabs/puppet/blob/3.7.3/lib/puppet/defaults.rb#L1198-L1204, so on startup of the master it's going to try and chmod it back to 0750.  I think it's best to enforce the mode by setting it in puppet.conf.

ekohl:
Can't we do both? Set it in puppet.conf and through puppet code?

domcleal:
True.

mmoll:
@mhjacks could you rebase and implement the changes to puppet.conf also?

mhjacks:
@mmoll where would you like me to make the changes to puppet.conf?  Would that be in https://github.com/theforeman/puppet-puppet?

mmoll:
@mhjacks that would be in puppet-puppet, yeah.

mmoll:
works for me :+1:

mmoll:
I wanted to cherry-pick this and do some minor modifications, mmoll/puppet-foreman@72cf291 was the outcome... is `selinux_ignore_defaults => true,` also needed for these?

domcleal:
It might be, since they're Puppet-managed directories and probably have a different default context to the one that Puppet tries to apply.

mmoll:
@lzap could you have a look regarding `selinux_ignore_defaults`?

domcleal:
I think leave it out for now.  I looked at #40 where it was originally added and if the Puppet master isn't managing these directories through defaults.rb then it _probably_ isn't relevant.  I think the issue in #40 was that the master run in unconfined_u (this is only on EL6 too), managed the yaml/ dir and caused that part of the label to reset.

I think more modern versions of Puppet don't try to manage that part of the label anyway, as well as these directories not being managed.

mmoll:
@domcleal so it should be OK for me to push mmoll/puppet-foreman@72cf291?

domcleal:
Yes, I think so.

mmoll:
perfect, thanks @mhjacks! and also thanks to @domcleal for clarifying!

PR #245, opened 2014-10-30
Allows sub-uri deployment

ares:
With this patch you can install foreman to sub-uri like this
```
foreman-installer --foreman-foreman-url="https://example.com/foreman" --foreman-proxy-foreman-base-url="https://example.com/foreman"
```


domcleal:
Nope, you're spot-on, you can only test the end resources.

Merged as 4cad220, thanks @ares!

PR #253, opened 2014-11-20
Added support for tar-based discovery image 2.0

lzap:
Tests are green, but I struggle testing this. When I do

`foreman-installer -v --foreman-plugin-discovery-install-images=true`

nothing happens. I am not sure if kafo caches the parameters or something (they
changed a bit). @ares any tips for me?

Thanks

### DIFF
--- a/manifests/plugin/discovery.pp
+++ b/manifests/plugin/discovery.pp
@@ -4,44 +4,44 @@
 #
 # === Parameters:
 #
-# $version::         version string of discovery image, in form of x.y.z-r
-#
-# $source::          mirror url from which the image files should be obtained, you
-#                    can use http(s):// or file://
-#
-# $initrd::          name of initrd image file
-#
-# $kernel::          name of kernel file
-#
 # $install_images::  should the installer download and setup discovery images
 #                    for you? the average size is few hundreds of MB
 #                    type:boolean
 #
+# $tftp_root::       tftp root to install image into
+#
+# $source_url::      source URL to download from
+#
+# $image_name::      tarball with images
+#
 class foreman::plugin::discovery (
-  $version        = $foreman::plugin::discovery::params::version,
-  $source         = $foreman::plugin::discovery::params::source,
-  $initrd         = $foreman::plugin::discovery::params::initrd,
-  $kernel         = $foreman::plugin::discovery::params::kernel,
   $install_images = $foreman::plugin::discovery::params::install_images,
+  $tftp_root      = $foreman::plugin::discovery::params::tftp_root,
+  $source_url     = $foreman::plugin::discovery::params::source_url,
+  $image_name     = $foreman::plugin::discovery::params::image_name,
 ) inherits foreman::plugin::discovery::params {
 
+  $tftp_root_clean = regsubst($tftp_root, '/$', '')
   validate_bool($install_images)
+  validate_absolute_path($tftp_root_clean)
+  validate_string($source_url)
+  validate_string($image_name)
 
   foreman::plugin {'discovery':
   }
 
   if $install_images {
-    include ::tftp::params
 
-    foreman::remote_file {"${::tftp::params::root}boot/${kernel}":
-      remote_location => "${source}${kernel}",
+    foreman::remote_file {"${tftp_root_clean}/boot/${image_name}":
+      remote_location => "${source_url}${image_name}",
       mode            => 0644,
-      require         => File["${::tftp::params::root}boot"],
-    }
-    foreman::remote_file {"${::tftp::params::root}boot/${initrd}":
-      remote_location => "${source}${initrd}",
-      mode            => 0644,
-      require         => File["${::tftp::params::root}boot"],
+      require         => File["${tftp_root_clean}/boot"],
+      notify          => Exec["untar ${image_name}"],
+    } ~> exec { "untar ${image_name}":
+      command => "tar xf ${image_name}",
+      path    => '/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin',
+      cwd     => "${tftp_root_clean}/boot",
+      creates => "${tftp_root_clean}/boot/fdi-image/initrd0.img",
     }
   }
 }
--- a/manifests/plugin/discovery/params.pp
+++ b/manifests/plugin/discovery/params.pp
@@ -1,8 +1,9 @@
 # Default parameters for foreman::plugin::discovery
 class foreman::plugin::discovery::params {
-  $version         = 'latest'
-  $source          = 'http://downloads.theforeman.org/discovery/releases/latest/'
-  $initrd          = "foreman-discovery-image-${version}.el6.iso-img"
-  $kernel          = "foreman-discovery-image-${version}.el6.iso-vmlinuz"
-  $install_images  = false
+  include ::tftp::params
+
+  $install_images = false
+  $tftp_root      = $::tftp::params::root
+  $source_url     = 'http://downloads.theforeman.org/discovery/releases/latest/'
+  $image_name     = 'fdi-image-latest.tar'
 }
--- a/spec/classes/foreman_plugin_discovery_spec.rb
+++ b/spec/classes/foreman_plugin_discovery_spec.rb
@@ -11,10 +11,6 @@ describe 'foreman::plugin::discovery' do
   context 'with enabled image installation' do
     let :params do
       {
-          :version => 'latest',
-          :source => 'http://yum.theforeman.org/discovery/releases/latest/',
-          :initrd => 'foreman-discovery-image-latest.el6.iso-img',
-          :kernel => 'foreman-discovery-image-latest.el6.iso-vmlinuz',
           :install_images => true
       }
     end
@@ -23,24 +19,25 @@ describe 'foreman::plugin::discovery' do
       should contain_foreman__plugin('discovery')
     end
 
-    it 'should download and install kernel file' do
-      should contain_foreman__remote_file("/var/lib/tftpboot/boot/foreman-discovery-image-latest.el6.iso-vmlinuz").
-                 with_remote_location('http://yum.theforeman.org/discovery/releases/latest/foreman-discovery-image-latest.el6.iso-vmlinuz')
+    it 'should download and install tarball' do
+      should contain_foreman__remote_file("/var/lib/tftpboot/boot/fdi-image-latest.tar").
+        with_remote_location('http://downloads.theforeman.org/discovery/releases/latest/fdi-image-latest.tar')
+    end
+    
+    it 'should extract the tarball' do
+      should contain_exec('untar fdi-image-latest.tar').with({
+        'command'     => "tar xf fdi-image-latest.tar",
+        'path'        => '/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin',
+        'cwd'         => '/var/lib/tftpboot/boot',
+        'creates'     => "/var/lib/tftpboot/boot/fdi-image/initrd0.img",
+      })
     end
 
-    it 'should download and install initrd file' do
-      should contain_foreman__remote_file("/var/lib/tftpboot/boot/foreman-discovery-image-latest.el6.iso-img").
-                 with_remote_location('http://yum.theforeman.org/discovery/releases/latest/foreman-discovery-image-latest.el6.iso-img')
-    end
   end
 
   context 'with disabled image installation' do
     let :params do
       {
-          :version => 'latest',
-          :source => 'http://yum.theforeman.org/discovery/releases/latest/',
-          :initrd => 'foreman-discovery-image-latest.el6.iso-img',
-          :kernel => 'foreman-discovery-image-latest.el6.iso-vmlinuz',
           :install_images => false
       }
     end
@@ -49,12 +46,8 @@ describe 'foreman::plugin::discovery' do
       should contain_foreman__plugin('discovery')
     end
 
-    it 'should not download and install kernel file' do
-      should_not contain_foreman__remote_file("/var/lib/tftpboot/boot/foreman-discovery-image-latest.el6.iso-vmlinuz")
-    end
-
-    it 'should not download and install initrd file' do
-      should_not contain_foreman__remote_file("/var/lib/tftpboot/boot/foreman-discovery-image-latest.el6.iso-img")
+    it 'should not download and install tarball' do
+      should_not contain_foreman__remote_file("/var/lib/tftpboot/boot/fdi-image-latest.tar")
     end
   end
 end


manifests/plugin/discovery/params.pp
```diff
@@ -1,8 +1,8 @@
 # Default parameters for foreman::plugin::discovery
 class foreman::plugin::discovery::params {
-  $version         = 'latest'
-  $source          = 'http://downloads.theforeman.org/discovery/releases/latest/'
-  $initrd          = "foreman-discovery-image-${version}.el6.iso-img"
-  $kernel          = "foreman-discovery-image-${version}.el6.iso-vmlinuz"
+  $major_version   = 'latest'
+  $minor_version   = 'latest'
+  $source          = "http://downloads.theforeman.org/discovery/releases/${major_version}/"
+  $image_name      = "fdi-image-${minor_version}.tar"
```
domcleal: using the versions inside source and image_name here will make it hard to override them, it'd be better to concatenate the filenames/URLs inside the class?

Else remove the version parameters as they don't perform any function right now.

manifests/plugin/discovery.pp
```diff
@@ -33,15 +33,19 @@
...
-    foreman::remote_file {"${::tftp::params::root}boot/${initrd}":
-      remote_location => "${source}${initrd}",
-      mode            => 0644,
-      require         => File["${::tftp::params::root}boot"],
+
+    exec { "untar ${image_name}":
+      command     => "tar xf ${image_name}",
+      path        => '/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin',
+      cwd         => "${::tftp::params::root}boot",
+      refreshonly => true,
```
domcleal: can we use `creates =>` with a filename, or does the filename inside the tarball vary?

manifests/plugin/discovery.pp
```diff
@@ -33,15 +33,19 @@
   if $install_images {
     include ::tftp::params
 
-    foreman::remote_file {"${::tftp::params::root}boot/${kernel}":
-      remote_location => "${source}${kernel}",
+    foreman::remote_file {"${::tftp::params::root}boot/${image_name}":
```
domcleal: It might be good to add $tftp_root as a class parameter which defaults to tftp::params::root, so it can be overridden like https://github.com/theforeman/puppet-foreman_proxy/commit/837bfccbf6318d259d861838fbfec9a9d20d7e03?  (Include tftp::params in foreman::plugin::discovery::params.)

domcleal:
We'll pull this into 3.0.0 if you can fix the issues noted above, as we can accept incompatible parameter changes for the major release.

lzap:
Note for myself: this PR is from my master.

manifests/plugin/discovery/params.pp
lzap: Hmmm how about to remove everything else except versions and install_images?

manifests/plugin/discovery.pp
lzap: Oh sure, created this:

      creates     => "${::tftp::params::root}boot/fdi-image/initrd0.img"],

manifests/plugin/discovery.pp
lzap: Thanks for the example, it makes our lifes *much* more easier :-)

Gonna try that now.

manifests/plugin/discovery/params.pp
ekohl: Why leave versions? I'd agree with just leaving ```source``` and ```image_name```.

lzap: Yeah since the web server is under our control, we can always symlink/redirect in case of incompatible change on the server side.

domcleal: Well, I think we either just expose versions and the (base) source URL, then concatenate the full URL, or we supply the full URL as a parameter.  I suppose how often are people going to want to override the version, and would it better to expose just that, or to override the whole URL?

manifests/plugin/discovery.pp
lzap: I think I don't get the example. Is this what you mean?

```diff
diff --git a/manifests/plugin/discovery.pp b/manifests/plugin/discovery.pp
index 334f63e..6191dbc 100644
--- a/manifests/plugin/discovery.pp
+++ b/manifests/plugin/discovery.pp
@@ -23,6 +23,7 @@ class foreman::plugin::discovery (
   $source         = $foreman::plugin::discovery::params::source,
   $image_name     = $foreman::plugin::discovery::params::image_name,
   $install_images = $foreman::plugin::discovery::params::install_images,
+  $tftp_root      = $tftp::params::root,
 ) inherits foreman::plugin::discovery::params {
```

In the example above it's different.

domcleal: <pre>
class foreman::plugin::discovery::params {
  include tftp::params
  $tftp_root = $tftp::params::root
}
</pre>

and here:

<pre>
class foreman::plugin::discovery (
  $tftp_root = $foreman::plugin::discovery::params::tftp_root,
)
</pre>

ekohl: Btw, since https://github.com/theforeman/puppet-tftp/commit/c0bc540d542ea2f2e037aad4926b32dc64d685f1 it can be overridden and before that you could already refer to ```$tftp::root``` due to inheritance so let's use that instead.

lzap:
Amended, pushed. I guess you want me to write a new test case now :-)

manifests/plugin/discovery.pp
```diff
@@ -4,44 +4,50 @@
...
 #
 class foreman::plugin::discovery (
-  $version        = $foreman::plugin::discovery::params::version,
-  $source         = $foreman::plugin::discovery::params::source,
-  $initrd         = $foreman::plugin::discovery::params::initrd,
-  $kernel         = $foreman::plugin::discovery::params::kernel,
+  $major_version  = $foreman::plugin::discovery::params::major_version,
+  $minor_version  = $foreman::plugin::discovery::params::minor_version,
   $install_images = $foreman::plugin::discovery::params::install_images,
+  $tftp_root      = $foreman::plugin::discovery::params::tftp_root,
```
ekohl: I think this should be documented for kafo.

manifests/plugin/discovery.pp
```diff
@@ -4,44 +4,50 @@
...
+
   foreman::plugin {'discovery':
   }
 
   if $install_images {
-    include ::tftp::params
 
-    foreman::remote_file {"${::tftp::params::root}boot/${kernel}":
-      remote_location => "${source}${kernel}",
+    foreman::remote_file {"${tftp_root}boot/${image_name}":
```
ekohl: Can we assume tftp_root has a trailing slash?

manifests/plugin/discovery.pp
ekohl: Also, can be validated using ```validate_absolute_path```.

manifests/plugin/discovery.pp
```diff
@@ -4,44 +4,50 @@
...
-  $kernel         = $foreman::plugin::discovery::params::kernel,
+  $major_version  = $foreman::plugin::discovery::params::major_version,
+  $minor_version  = $foreman::plugin::discovery::params::minor_version,
   $install_images = $foreman::plugin::discovery::params::install_images,
+  $tftp_root      = $foreman::plugin::discovery::params::tftp_root,
 ) inherits foreman::plugin::discovery::params {
 
   validate_bool($install_images)
 
+  $source          = "http://downloads.theforeman.org/discovery/releases/${major_version}/"
```
ekohl: I think this should be configurable and then join the two parameters. Surely someone will want their own mirror for this.

lzap: I am not so keen about this. If someone sets own mirror, then he/she can just download it. I don't expect this.

manifests/plugin/discovery.pp
lzap: Right, fixed.

manifests/plugin/discovery.pp
lzap: Yeah double `//` won't hurt on UNIX. Fixed.

manifests/plugin/discovery.pp
lzap: I mean - I think this is extra effort that it isn't worth it... I can do that if you insist.

ekohl: I suppose we can leave it for the future.

lzap: Hmmm so what do you expect? Can you list me inputs you want to see? Because I have just withdrawn this from options. I don't understand what you mean. Thanks.

ekohl: I thought either:
```puppet
class foreman::plugin::discovery (
  $base_source_url = "http://downloads.theforeman.org/discovery/releases/",
  $major_version = "latest",
) {
  $source_url = "${base_source_url}${major_version}/"
}
```

or

```puppet
class foreman::plugin::discovery (
  $source_url = "http://downloads.theforeman.org/discovery/releases/latest/",
) {}
```

I'm leaning to the latter.

lzap:
Ok amended the source_url change.

manifests/plugin/discovery.pp
```diff
@@ -4,44 +4,51 @@
...
   if $install_images {
-    include ::tftp::params
 
-    foreman::remote_file {"${::tftp::params::root}boot/${kernel}":
-      remote_location => "${source}${kernel}",
+    foreman::remote_file {"${tftp_root_clean}/boot/${image_name}":
+      remote_location => "${source_url}${image_name}",
       mode            => 0644,
-      require         => File["${::tftp::params::root}boot"],
+      require         => File["${tftp_root_clean}/boot"],
```
ekohl: I wonder we should solve this straight away in foreman::remote_file, but can be fixed later without breaking the API.

manifests/plugin/discovery.pp
```diff
@@ -4,44 +4,51 @@
...
-    foreman::remote_file {"${::tftp::params::root}boot/${initrd}":
-      remote_location => "${source}${initrd}",
-      mode            => 0644,
-      require         => File["${::tftp::params::root}boot"],
+
+    exec { "untar ${image_name}":
+      command     => "tar xf ${image_name}",
+      path        => '/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin',
+      cwd         => "${tftp_root_clean}/boot",
+      refreshonly => true,
```
ekohl: With creates this is redundant.

manifests/plugin/discovery/params.pp
```diff
@@ -1,8 +1,9 @@
...
-  $version         = 'latest'
-  $source          = 'http://downloads.theforeman.org/discovery/releases/latest/'
-  $initrd          = "foreman-discovery-image-${version}.el6.iso-img"
-  $kernel          = "foreman-discovery-image-${version}.el6.iso-vmlinuz"
+  include ::tftp::params
+
+  $major_version   = 'latest'
+  $minor_version   = 'latest'
   $install_images  = false
+  $tftp_root       = $tftp::params::root
```
ekohl: I think $::tftp::params::root is cleaner here.

manifests/plugin/discovery.pp
```diff
@@ -4,44 +4,51 @@
...
+# $tftp_root::       tftp root to install image into (optional)
+#
+# $source_url::      source URL to download from (optional)
+#
 class foreman::plugin::discovery (
-  $version        = $foreman::plugin::discovery::params::version,
-  $source         = $foreman::plugin::discovery::params::source,
-  $initrd         = $foreman::plugin::discovery::params::initrd,
-  $kernel         = $foreman::plugin::discovery::params::kernel,
+  $major_version  = $foreman::plugin::discovery::params::major_version,
```
ekohl: This is no longer used.

manifests/plugin/discovery.pp
```diff
@@ -4,44 +4,51 @@
...
   $install_images = $foreman::plugin::discovery::params::install_images,
+  $tftp_root      = $foreman::plugin::discovery::params::tftp_root,
+  $source_url     = "http://downloads.theforeman.org/discovery/releases/latest/",
 ) inherits foreman::plugin::discovery::params {
 
+  $tftp_root_clean = regsubst($tftp_root, '/$', '')
   validate_bool($install_images)
+  validate_absolute_path($tftp_root_clean)
+
+  $image_name      = "fdi-image-${minor_version}.tar"
```
ekohl: Should we move this to a class parameter as well so we can drop *_version from the parameters?

manifests/plugin/discovery.pp
lzap: > I wonder we should solve this straight away in foreman::remote_file, but can be fixed later without breaking the API.

Not following. Please keep in mind I am not the one who should do
refactoring of any puppet code. Let's proceed with bare-minimal changes
:-)

-- 
Later,
 Lukas #lzap Zapletal

manifests/plugin/discovery.pp
lzap: > With creates this is redundant.

Removed.

-- 
Later,
 Lukas #lzap Zapletal

manifests/plugin/discovery/params.pp
lzap: > I think $::tftp::params::root is cleaner here.

Not a problem.

-- 
Later,
 Lukas #lzap Zapletal

manifests/plugin/discovery.pp
lzap: > > +  $major_version  = $foreman::plugin::discovery::params::major_version,
> 
> This is no longer used.

Right, I am tempted to remove both versions now because in each
directory (started with version 2.0) we have

http://downloads.theforeman.org/discovery/releases/

the fdi-image-latest.tar symlink.

And since we default to "latest" anyway and offer user to change the
base URL, there is no need for this. Much cleaner and more simple.

What you think?

-- 
Later,
 Lukas #lzap Zapletal

manifests/plugin/discovery.pp
lzap: > > +  $image_name      = "fdi-image-${minor_version}.tar"
> 
> Should we move this to a class parameter as well so we can drop *_version from the parameters?

Or this way is even better. Will do.

-- 
Later,
 Lukas #lzap Zapletal

manifests/plugin/discovery.pp
ekohl: Fair enough. What I meant was that ```foreman::remote_file``` could have this require line somewhere so you need to worry less when using it. However, I think this is the only use of it so it wouldn't be a big win.

ekohl:
I think it's good, but travis found a few lint issues which should be solved first.

manifests/plugin/discovery.pp
```diff
@@ -4,44 +4,47 @@
...
-#
-# $initrd::          name of initrd image file
-#
-# $kernel::          name of kernel file
-#
 # $install_images::  should the installer download and setup discovery images
 #                    for you? the average size is few hundreds of MB
 #                    type:boolean
 #
+# $tftp_root::       tftp root to install image into (optional)
```
domcleal: I think "(optional)" is superfluous, as these have default values.

manifests/plugin/discovery.pp
```diff
@@ -4,44 +4,47 @@
...
-    include ::tftp::params
 
-    foreman::remote_file {"${::tftp::params::root}boot/${kernel}":
-      remote_location => "${source}${kernel}",
+    foreman::remote_file {"${tftp_root_clean}/boot/${image_name}":
+      remote_location => "${source_url}${image_name}",
       mode            => 0644,
-      require         => File["${::tftp::params::root}boot"],
+      require         => File["${tftp_root_clean}/boot"],
+      notify          => Exec["untar ${image_name}"],
```
domcleal: Nitpick, you can use `~>` between the two resources if you like.

manifests/plugin/discovery.pp
```diff
@@ -4,44 +4,47 @@
...
-    foreman::remote_file {"${::tftp::params::root}boot/${initrd}":
-      remote_location => "${source}${initrd}",
-      mode            => 0644,
-      require         => File["${::tftp::params::root}boot"],
+
+    exec { "untar ${image_name}":
+      command     => "tar xf ${image_name}",
+      path        => '/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin',
+      cwd         => "${tftp_root_clean}/boot",
+      require     => File["${tftp_root_clean}/boot/${image_name}"],
```
domcleal: This is superfluous given the notify above, best just replace both lines with the `~>`

spec/classes/foreman_plugin_discovery_spec.rb
```diff
@@ -49,12 +46,9 @@
...
 
-    it 'should not download and install kernel file' do
-      should_not contain_foreman__remote_file("/var/lib/tftpboot/boot/foreman-discovery-image-latest.el6.iso-vmlinuz")
-    end
-
-    it 'should not download and install initrd file' do
-      should_not contain_foreman__remote_file("/var/lib/tftpboot/boot/foreman-discovery-image-latest.el6.iso-img")
+    it 'should not download and install tarball' do
+      should_not contain_foreman__remote_file("/var/lib/tftpboot/boot/fdi-image-latest.tar").
+        with_remote_location('http://yum.theforeman.org/discovery/releases/latest/fdi-image-latest.tar')
```
domcleal: best to remove this line, so you're just checking for the resource not to exist, rather than the resource with a specific parameter

domcleal:
Looks about right, yep!

lzap:
Incorporated all your comments, tests are green! Thanks.

manifests/plugin/discovery.pp
domcleal: That means you can drop this notify line now.

manifests/plugin/discovery.pp
```diff
@@ -4,44 +4,44 @@
...
-    foreman::remote_file {"${::tftp::params::root}boot/${initrd}":
-      remote_location => "${source}${initrd}",
+    foreman::remote_file {"${tftp_root_clean}/boot/${image_name}":
+      remote_location => "${source_url}${image_name}",
       mode            => 0644,
-      require         => File["${::tftp::params::root}boot"],
+      require         => File["${tftp_root_clean}/boot"],
+      notify          => Exec["untar ${image_name}"],
+    } ~> exec { "untar ${image_name}":
+      command     => "tar xf ${image_name}",
```
ekohl: puppet-lint doesn't like the extra whitespace in between here.

manifests/plugin/discovery/params.pp
```diff
@@ -1,8 +1,9 @@
...
-  $version         = 'latest'
-  $source          = 'http://downloads.theforeman.org/discovery/releases/latest/'
-  $initrd          = "foreman-discovery-image-${version}.el6.iso-img"
-  $kernel          = "foreman-discovery-image-${version}.el6.iso-vmlinuz"
-  $install_images  = false
+  include ::tftp::params
+
+  $install_images = false
+  $tftp_root      = $::tftp::params::root
+  $source_url     = "http://downloads.theforeman.org/discovery/releases/latest/"
```
ekohl: puppet-lint wants these using ```'``` instead of ```"``` since there's no variable here.

manifests/plugin/discovery/params.pp
```diff
@@ -1,8 +1,9 @@
...
+  $image_name     = "fdi-image-latest.tar"
```
ekohl: same

ekohl:
Once the puppet-lint issues are fixed, travis should be green.

manifests/plugin/discovery/params.pp
lzap: this rule is insane, why should we waste our time with this?

lzap:
Wait, why do we error out on *WARNINGS*? Jeeeez, nobody reads warnings...

```
manifests/plugin/discovery.pp:41:WARNING: indentation of => is not properly aligned
manifests/plugin/discovery.pp:42:WARNING: indentation of => is not properly aligned
manifests/plugin/discovery.pp:43:WARNING: indentation of => is not properly aligned
manifests/plugin/discovery.pp:44:WARNING: indentation of => is not properly aligned
manifests/plugin/discovery/params.pp:7:WARNING: double quoted string containing no variables
manifests/plugin/discovery/params.pp:8:WARNING: double quoted string containing no variables
```

lzap:
Amended.

domcleal:
Keeps our modules in good shape, easier to read and maintain.  The alignment doesn't seem to have been fixed in your latest commit?

ekohl:
By keeping in line with puppet-lint we also get a higher score on the quality rating that the forge gives. Not sure why there's one lint failure on https://forge.puppetlabs.com/theforeman/foreman/scores though.

domcleal:
@ekohl that'll be due to using inline lint ignore comments (manifests/rake.pp, due to a lint bug).  The Forge's lint scanner deliberately ignores them: https://tickets.puppetlabs.com/browse/FORGE-233

lzap:
Pushing my last idea. If that's red, please tell me how do I properly align this.

Let me rate quality of the Puppet language design here: **F** ;-)

domcleal:
Merged as 23c5c10 with the notify fixed, thanks @lzap.

One thought is that once somebody adds smart_proxy_discovery to puppet-foreman_proxy, this image downloading ought to move to that module.

PR #255, opened 2014-12-03
Move from interface-based listen to ip-based parameter

roidelapluie:
See theforeman/puppet-puppet#200

### DIFF
--- a/manifests/config/passenger.pp
+++ b/manifests/config/passenger.pp
@@ -4,8 +4,8 @@
 #
 # $app_root::               Root of the application.
 #
-# $listen_on_interface::    Specify which interface to bind passenger to.
-#                           Defaults to all interfaces.
+# $ip::                     Specify which IP address to bind passenger to.
+#                           Defaults to any IP addresses.
 #
 # $ruby::                   Path to Ruby interpreter
 #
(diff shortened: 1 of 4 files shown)


manifests/config/passenger.pp
```diff
@@ -4,8 +4,8 @@
 #
 # $app_root::               Root of the application.
 #
-# $listen_on_interface::    Specify which interface to bind passenger to.
-#                           Defaults to all interfaces.
+# $ip::                     Specify which ip address to bind passenger to.
```
domcleal: Nitpick, "ip" should be "IP" in docs

domcleal:
So it looks like we're prepping a major release of the module, so can make this change if we want.  I think I'm for it, could you rebase it please @roidelapluie?

roidelapluie:
@domcleal here you have :smile:

domcleal:
Seems that this triggers a check in pl-apache about mixing different listen configurations, needs a bit more investigation about what to change in the vhost parameters I think.

https://github.com/puppetlabs/puppetlabs-apache/blob/master/manifests/vhost.pp#L336-L338

spec/classes/foreman_config_spec.rb
```diff
@@ -84,7 +84,7 @@
 
       it 'should contain foreman::config::passenger' do
         should contain_class('foreman::config::passenger').
-          with_listen_on_interface('').
+          with_ip(:undef).
```
domcleal: `nil` rather than :undef here, that's what it ends up evaluating to.  Most of the time, sigh.

spec/classes/foreman_config_spec.rb
```diff
@@ -259,8 +259,8 @@
       end
 
       it { should contain_class('foreman::config::passenger').with({
-        :listen_on_interface => '',
-        :ruby                => '',
+        :ip   => :undef,
```
domcleal: nil too

roidelapluie:
@ekohl I have updated this PR

manifests/config/passenger.pp
```diff
@@ -36,25 +36,25 @@
...
-  $ssl_cert            = $::foreman::server_ssl_cert,
-  $ssl_key             = $::foreman::server_ssl_key,
-  $ssl_crl             = $::foreman::server_ssl_crl,
-  $use_vhost           = $::foreman::use_vhost,
-  $user                = $::foreman::user,
-  $prestart            = $::foreman::passenger_prestart,
-  $min_instances       = $::foreman::passenger_min_instances,
-  $start_timeout       = $::foreman::passenger_start_timeout,
+  $app_root      = $::foreman::app_root,
+  $ip            = $::foreman::passenger_ip,
```
ekohl: lint fails on trailing whitespace here.

ekohl:
I still think this is a good change, but I'm a bit unsure how it will impact existing installations. This will I will try to get my test environment updated again and have a go at it.

roidelapluie:
Closing after inactivity

PR #263, opened 2014-12-12
Use rspec-puppet-facts

mcanevet:
Work in Progress. Could be merged as it though.

### DIFF
--- a/spec/classes/foreman_cli_spec.rb
+++ b/spec/classes/foreman_cli_spec.rb
@@ -1,93 +1,97 @@
 require 'spec_helper'
 
 describe 'foreman::cli' do
-  context 'standalone with parameters' do
-    let(:params) do {
-      'foreman_url' => 'http://example.com',
-      'username'    => 'joe',
-      'password'    => 'secret',
-    } end
 
-    it { should contain_package('foreman-cli').with_ensure('installed') }
-
-    describe '/etc/hammer/cli.modules.d/foreman.yml' do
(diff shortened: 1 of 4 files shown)


spec/classes/foreman_config_passenger_spec.rb
```diff
@@ -1,168 +1,132 @@
...
-        })
-      end
-
-      it 'should not include start timeout fragment on Squeeze' do
-        should contain_apache__vhost('foreman-ssl').without({
-          :custom_fragment => %r{^PassengerStartTimeout},
-        })
-      end
-
-      it 'should not include the Ruby interpreter' do
```
domcleal: Any way we can keep this example for wheezy and trusty?

mcanevet:
@domcleal I removed `foreman::config::passenger`'s specs refactoring from the PR as it is not ready yet.

domcleal:
Ah okay, that's fine for now, we can have another go later.  Could you give it a rebase now I've done the modulesync?  I think we're ready to go.

ekohl:
:+1: when the tests pass

domcleal:
They're taking a very long time!

ekohl:
That's caused by running the tests on many operatingsystems since there are 11 in metadata.json. I suspect that ```include foreman``` as precondition in many tests makes it a lot slower than needed.

domcleal:
Thanks @mcanevet, great work!

mcanevet:
\o/

PR #269, opened 2015-01-09
fixes #8880 yum plugins repo, removed rc

endyman:
This PR adds yum plugins repo and removes (according @domcleal) outdated rc versions and tests.

### DIFF
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -35,7 +35,7 @@
 #                             set it to some custom location.
 #                             type:boolean
 #
-# $repo::                     This can be stable, rc, or nightly
+# $repo::                     This can be stable, nightly or a specific version i.e. 1.7
 #
 # $configure_epel_repo::      If disabled the EPEL repo will not be configured on RedHat family systems.
 #                             type:boolean
--- a/manifests/install/repos/yum.pp
+++ b/manifests/install/repos/yum.pp
@@ -1,8 +1,15 @@
 # Install a yum repo
 define foreman::install::repos::yum ($repo, $yumcode, $gpgcheck) {
   $repo_path = $repo ? {
-    'stable' => 'releases/latest',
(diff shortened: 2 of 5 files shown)


manifests/install/repos/yum.pp
```diff
@@ -1,8 +1,13 @@
 # Install a yum repo
 define foreman::install::repos::yum ($repo, $yumcode, $gpgcheck) {
   $repo_path = $repo ? {
-    'stable' => 'releases/latest',
-    default  => $repo,
+    'stable'  => 'releases/latest',
+    'nightly' => $repo,
+    default   => "releases/${repo}",
```
domcleal: We might need to handle the case where somebody's already passing "releases/1.7" or similar in, so we don't prepend "releases/" again.

manifests/init.pp
```diff
@@ -35,7 +35,7 @@
 #                             set it to some custom location.
 #                             type:boolean
 #
-# $repo::                     This can be stable, rc, or nightly
+# $repo::                     This can be stable, nightly or a apecific version i.e. 1.7
```
domcleal: specific

mmoll:
:+1:

manifests/install/repos/yum.pp
```diff
@@ -1,8 +1,14 @@
...
-    'stable' => 'releases/latest',
-    default  => $repo,
+    'stable'      => 'releases/latest',
+    /^releases\// => $repo,
+    'nightly'     => $repo,
+    default       => "releases/${repo}",
+  }
+  $plugins_repo_path = $repo ? {
+    'stable' => 'plugins/latest',
+    default  => "plugins/${repo}",
```
domcleal: Maybe handle `releases/` here too?

domcleal:
Thanks @endyman!

PR #270, opened 2015-01-14
fixes #8944 - renaming facts parameter to receive_facts

gpetras:
This is just a rename of the ```$facts``` parameter, as it's a reserved keyword when ```trusted_node_data = true```.

### DIFF
--- a/manifests/params.pp
+++ b/manifests/params.pp
@@ -10,8 +10,8 @@ class foreman::params {
   $enc          = true
   # Should foreman receive reports from puppet
   $reports      = true
-  # Should foreman recive facts from puppet
-  $facts        = true
+  # Should foreman receive facts from puppet
+  $receive_facts = true
   # should foreman manage host provisioning as well
   $unattended   = true
   # Enable users authentication (default user:admin pw:changeme)
--- a/manifests/puppetmaster.pp
+++ b/manifests/puppetmaster.pp
@@ -6,7 +6,7 @@ class foreman::puppetmaster (
   $foreman_password = $foreman::params::foreman_password,
   $reports          = $foreman::params::reports,
   $enc              = $foreman::params::enc,
-  $facts            = $foreman::params::facts,
+  $receive_facts    = $foreman::params::receive_facts,
   $puppet_home      = $foreman::params::puppet_home,
   $puppet_user      = $foreman::params::puppet_user,
   $puppet_group     = $foreman::params::puppet_group,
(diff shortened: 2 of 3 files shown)


mmoll:
:+1:, but also needs a change here: https://github.com/theforeman/puppet-puppet/blob/master/manifests/server/config.pp#L30

domcleal:
Fair enough - but just thinking out loud to see if we can avoid a major version bump, is the error a hard one that fails the whole Puppet run when the variable is defined?  If not, we should probably keep the parameter and handle it for compatibility..

gpetras:
Thanks guys. I am running Puppet Enterprise 3.7.1, which bundles puppet 3.7.3. This is the error received:

```
Error: Could not retrieve catalog from remote server: Error 400 on SERVER: Attempt to assign to a reserved variable name: 'facts' at modules/foreman/manifests/params.pp:14 on node foreman.example.com
Warning: Not using cache on failed catalog
Error: Could not retrieve catalog; skipping run
```

ekohl:
https://docs.puppetlabs.com/puppet/latest/reference/lang_reserved.html#reserved-variable-names
> If enabled, the $trusted and $facts variables are reserved for facts and cannot be reassigned at local scopes.

domcleal:
Ah, so it's the params declaration that's the issue.  For compatibility, can I suggest:

params.pp
    $facts_old = undef
    $receive_facts = true

puppetmaster.pp
    class puppetmaster(
      $facts = $facts_old,
      $receive_facts = $foreman::params::receive_facts,
    )

and then pick the first used variable:

    $real_receive_facts = pick($facts, $receive_facts)

Then we avoid removing the parameter altogether right now.

domcleal:
Oh, I'm talking rubbish.  It still doesn't seem to allow it, and $facts will actually contain data.  Sigh, I guess the rename is inevitable.

gpetras:
Well, always good to try to make backwards compatible changes. I didn't see an obvious way to do that when I looked at it. Feel free to hold onto this PR until you think it's appropriate to merge. Or if you come up with an idea I'd be happy to implement and test.

domcleal:
Yeah, I thought I'd just tested it successfully and then realised I'd not done it correctly.  Unless @ekohl has any better ideas, we should merge it and the next version will be bumped, c'est la vie.

ekohl:
I'm afraid there's no other solution and a bump it is. :+1:

mmoll:
merged as 022b819, thanks @gpetras :tada:

PR #271, opened 2015-01-15
Remove deprecated passenger_scl parameter

domcleal:


ekohl:
:+1: it's good to clean up

mmoll:
merged as a6d46ed, thanks @domcleal
While I did test this only on Debian, it didn't break and looks :ok:

PR #273, opened 2015-01-19
Add openscap plugin support

ares:


manifests/plugin/openscap.pp
```diff
@@ -0,0 +1,66 @@
...
+            $repo = 'epel'
+          }
+        }
+
+        if $::osreleasemajor == undef {
+          if $::operatingsystemmajrelease == undef {
+            $versions_array = split($::operatingsystemrelease, '\.') # facter 1.6
+            $major = $versions_array[0]
+          } else {
+            $major = $::operatingsystemmajrelease # facter 2.2
```
ekohl: Actually facter 1.7+

manifests/plugin/openscap.pp
```diff
@@ -0,0 +1,66 @@
...
+        case $::operatingsystem {
+          'fedora': {
+            $repo = 'fedora'
+          }
+          default: {
+            $repo = 'epel'
+          }
+        }
+
+        if $::osreleasemajor == undef {
```
ekohl: Don't we have this logic somewhere else already?

manifests/plugin/openscap.pp
```diff
@@ -0,0 +1,66 @@
...
+#                             puppet module into this directory for you
+#                             e.g. /etc/puppet/environments/production/modules
+#
+class foreman::plugin::openscap (
+  $configure_openscap_repo = false,
+  $scap_client_module_dir  = undef,
+) {
+  case $::osfamily {
+    'RedHat': {
+      validate_bool($configure_openscap_repo)
```
ekohl: I'd pull this out of the case statement.

spec/classes/foreman_plugin_openscap_spec.rb
```diff
@@ -0,0 +1,63 @@
+require 'spec_helper'
+
+describe 'foreman::plugin::openscap' do
+  let :facts do
```
ekohl: We have sets of facts now to automatically test on all supported platforms. See https://github.com/theforeman/puppet-foreman/blob/master/spec/classes/foreman_config_ipa_spec.rb for more examples.

manifests/plugin/openscap.pp
```diff
@@ -0,0 +1,66 @@
...
+            $versions_array = split($::operatingsystemrelease, '\.') # facter 1.6
+            $major = $versions_array[0]
+          } else {
+            $major = $::operatingsystemmajrelease # facter 2.2
+          }
+        } else {
+          $major = $::osreleasemajor
+        }
+
+        yumrepo { 'isimluk-openscap':
```
domcleal: same comment, should be in our repos, not copr

manifests/plugin/openscap.pp
```diff
@@ -0,0 +1,66 @@
+# = Foreman Openscap plugin
+#
+# This class installs openscap plugin
```
domcleal: "OpenSCAP"

manifests/plugin/openscap.pp
ekohl: For FreeIPA we already have a one copr in https://github.com/theforeman/puppet-foreman/blob/b5d8e4f40bf7ecabe0d11ddacc048b25b1e04558/manifests/install/repos/extra.pp#L32-L39

domcleal: Yeah, but this also appears to be shipping the plugins themselves, which isn't best.  (Our repos are laid out better, separated by major version, plus we have Debian variants.)

I understood from the thread with Simon on foreman-dev that future EL versions will ship the required packages, so it's likely this repo can then be dropped or disabled if we ship the plugins.

manifests/plugin/openscap.pp
ares: originally I thought I could [reuse this](https://github.com/theforeman/puppet-foreman/blob/master/manifests/install/repos/extra.pp#L9) but after lint changes, it does not make sense, I'll remove it

manifests/plugin/openscap.pp
ekohl: Fair enough. So the policy is plugins in our repos, core packages preferably in base OS.

domcleal: That's my preference, yep.

ares: so this repo must remain here because of openscap package, moving gems into foreman-packaging is separate issue right?

domcleal: I'd like to see it fixed first, before we add a repo to the installer that contains an assortment of packages.

ares:
We have all plugin packages built, is there anything else I could do for this to be merged?

manifests/plugin/openscap.pp
```diff
@@ -0,0 +1,63 @@
...
+#
+# $configure_openscap_repo::  Enable custom yum repo with packages needed for foreman_openscap,
+#                             type:boolean
+#
+# $scap_client_module_dir::   if you set a directory here, installer will upload foreman_scap_client
+#                             puppet module into this directory for you
+#                             e.g. /etc/puppet/environments/production/modules
+#
+class foreman::plugin::openscap (
+  $configure_openscap_repo = false,
```
domcleal: Should this be true, as it won't work at the moment without it?  We can perhaps change it to false later, once it's generally available.

manifests/plugin/openscap.pp
```diff
@@ -0,0 +1,63 @@
...
+          before   => [ Foreman::Plugin['openscap'] ],
+        }
+      }
+
+      foreman::plugin {'openscap': }
+
+      if $scap_client_module_dir {
+        file { "${scap_client_module_dir}/foreman_scap_client":
+          ensure  => directory,
+          source  => '/usr/share/foreman-installer/modules/foreman_scap_client',
```
domcleal: Where does this come from?

manifests/plugin/openscap.pp
```diff
@@ -0,0 +1,63 @@
...
+
+        yumrepo { 'isimluk-openscap':
+          enabled  => 1,
+          gpgcheck => 0,
+          baseurl  => "http://copr-be.cloud.fedoraproject.org/results/isimluk/OpenSCAP/${repo}-${major}-\$basearch/",
+          before   => [ Foreman::Plugin['openscap'] ],
+        }
+      }
+
+      foreman::plugin {'openscap': }
```
domcleal: Shouldn't this and the bit below be outside of the 'RedHat' osfamily conditional?

manifests/plugin/openscap.pp
ares: https://github.com/theforeman/foreman-installer/pull/140/files#diff-53e02589c79aa03f1974d2c3012523e2R17

manifests/plugin/openscap.pp
```diff
@@ -0,0 +1,63 @@
...
+  $configure_openscap_repo = false,
+  $scap_client_module_dir  = undef,
+) {
+  validate_bool($configure_openscap_repo)
+
+  case $::osfamily {
+    'RedHat': {
+
+      if $configure_openscap_repo {
+        case $::operatingsystem {
```
ekohl: I prefer the shorthand:
```puppet
$repo = $::operatingsystem ? {
  'Fedora' => 'fedora',
  default  => 'epel',
}
```

manifests/plugin/openscap.pp
ares: I wanted to keep consistency with other custom repos (e.g. https://github.com/theforeman/puppet-foreman/blob/master/manifests/params.pp#L218). Also I think user should explicitly allow extra repo. Maybe a better parameter documentation would help?

manifests/plugin/openscap.pp
ares: We currently support only RedHat family AFAIK.

domcleal: What do you mean by "we"?

manifests/plugin/openscap.pp
domcleal: I'm not sure I understand, the installer PR links to some client gem?

manifests/plugin/openscap.pp
ares: Foreman packagers, who didn't manage to build deb packages yet. Or do you think we'll manage to build them in near future? Note that there are some openscap libs in Debian wheezy, version is 0.8.0 though and 1.0.9 in Jessie.

manifests/plugin/openscap.pp
ares: Oh sorry, there was wrong URL. Updated

manifests/plugin/openscap.pp
ekohl: IMHO that doesn't matter. The case statement will ensure it's not reachable for debian anyway, but when it is, it's just a matter of adding debian to the case statement.

domcleal: I kind of prefer to keep the plugin classes OS-agnostic, so once the packages are available in our plugin repos, we don't need to change the installer..

manifests/plugin/openscap.pp
domcleal: This should be a module dependency then, not added via the Puppetfile.

domcleal: And it's not really reliable, it's only going to work with our packages, not with the module itself.  Shouldn't the module be packaged into /usr/share/puppet/modules?

manifests/plugin/openscap.pp
ekohl: Agreed, but if OS-specific configuration (such as a repo) is needed IMHO that should be allowed. I'm unsure about hard failures for an unsupported OS though.

ares: From user experience perspective, they will either see "Unsupported family" or "E: Unable to locate package XY" and will raise issue that package is missing or installation does not work.

manifests/plugin/openscap.pp
domcleal: I'm also having second thoughts about whether this is the right module, given that the Puppet master is installed by puppet-puppet.  Though we don't mix in the concept of Foreman plugins in there (yet).

ekohl: I'm somewhat lost on the general design here. Maybe because I don't know openscap at all. The module sets up cronjobs, but when is it ever executed? I don't see an include and if you copy it here, then I don't think you can use it in the same foreman-installer run.

domcleal: Yeah, I'm also rather lost on what's going on.

ares: In order to configure scap clients to run scans and upload it to foreman we have to setup foreman_scap_client script. foreman-openscap plugin rely on existing foreman_scap_client puppet class which it assigns to specified hostgroup in order to configure foreman_scap_client script. Basically foreman-openscap gathers information about policies and content from wizards and apply it on host through this puppet class which sets a cron and other things so that foreman_scap_client is run.  
We wanted to make it easy for users to use foreman-openscap so they don't have to install foreman_scap_client puppet module manually. Installer could do that for them if they explicitly ask for it and they select an environment where it should be put.  
Please ask if I should elaborate some part.

ekohl: Ah, so you're trying to manage puppet environments. That's a problem that we don't tackle anywhere so requires some thought. Right now I'm not sure what the best solution to this is. Can this be split off into a separate PR so we can merge basic openscap support first?

manifests/plugin/openscap.pp
ekohl: This plugin is optional so I wouldn't mind defaulting to true.

manifests/plugin/openscap.pp
ares: Possibly could be split but what is the downside? especially when it's off by default? Since this was originally meant as installer only feature maybe I could implement it as an installer hook. But if we add foreman_scap_client as a module dependency as suggested above it could work also in use with puppetmaster.

ekohl: Because managing client environments is not something we should add lightly. The installer is not the only use case and this is not something that fits into the regular puppet deployment. That's not to say I'm against it per se, but it requires proper design IMHO.

domcleal: The downside is that at the moment it simply appears wrong - as it's probably in the wrong module and assumes where source files are, so it's something else for us to maintain and ideally replace later (probably breaking API too).

manifests/plugin/openscap.pp
ekohl: But it may not be clear that it's unsupported because of openscap. I'd at least change that in the failure message. Still, I think having it supported but unpackaged is an option because it does allow an async release.

ares:
I updated the PR. Plugin resource is now out of case which is now only used if we want to configure the repo

spec/classes/foreman_plugin_openscap_spec.rb
```diff
@@ -0,0 +1,78 @@
+require 'spec_helper'
+
+describe 'foreman::plugin::openscap' do
+  on_supported_os.each do |os, facts|
+    if facts[:osfamily] == 'RedHat'
+      context "on #{os}" do
+        let :facts do
```
ekohl: Isn't this redundant?

ares: ah sure, after all those changes, it does not make sense, thanks, fixed

ekohl:
:+1:

manifests/plugin/openscap.pp
```diff
@@ -0,0 +1,45 @@
...
+
+        yumrepo { 'isimluk-openscap':
+          enabled  => 1,
+          gpgcheck => 0,
+          baseurl  => "http://copr-be.cloud.fedoraproject.org/results/isimluk/OpenSCAP/${repo}-${major}-\$basearch/",
+          before   => [ Foreman::Plugin['openscap'] ],
+        }
+      }
+      default: {
+        fail("Unsupported osfamily ${::osfamily}")
```
domcleal: Sorry to harp on about this, but I think the default value for $configure_openscap_repo should perhaps be determined by a params file so it's only enabled if $::osfamily is 'RedHat'.

At the moment this is going to fail by default on all non-'RedHat' OSes without explicit reconfiguration.

e.g. https://github.com/theforeman/puppet-foreman/blob/master/manifests/params.pp#L33-L38

(sorry, I hadn't realised setting the default to true would result in fail() when I wrote my earlier comments)

spec/classes/foreman_plugin_openscap_spec.rb
ares: actually this is needed, oitherwise facts are empty, on_supported_os does not seem to set facts (which makes sense because of various contexts)

ares:
@domcleal could you please re-review and tell me what else must be done in order to get this merged?

manifests/plugin/openscap/params.pp
```diff
@@ -0,0 +1,4 @@
+# Default parameters for foreman::plugin::openscap
+class foreman::plugin::openscap::params {
+  $configure_openscap_repo = ($::osfamily == 'RedHat' and $::operatingsystem != 'RedHat')
```
domcleal: RHEL shouldn't be excluded

spec/classes/foreman_plugin_openscap_spec.rb
ekohl: I'd say that's a bug. @mcanevet is that intended?

mcanevet: @ares why don't you assign facts just after `on_supported_os.each do |os, facts|`?

```puppet
  on_supported_os.each do |os, facts|
    context "on #{os}" do
      let(:facts) do
        facts
      end

      if facts[:osfamily] == 'RedHat'
        ...
      end
   end
```

ares:
Something else?

domcleal:
:+1: in general, once the facts/specs question is resolved.

spec/classes/foreman_plugin_openscap_spec.rb
ares: you mean like this?
```ruby
describe 'foreman::plugin::openscap' do
  on_supported_os.each do |os, facts|
    let :facts do
      facts
    end

    if facts[:osfamily] == 'RedHat'
      context "on #{os}" do
      ...
```
This way they I'm getting errors since facts in spec seems OK while those used for catalog compilation seems to use cached value from another run, I suppose facts are cached in `describe` context in this case and can't be redeclared using let.

Also I think the question was more about whether we need 
```ruby
let (:facts) do 
  facts 
end
```

snippet even without customization, which I think we need, otherwise my facts seems empty.

ekohl: I think @mcanevet wants a context for every OS, which I think is fine.

ekohl:
:+1: I think the facts/specs question is solved now.

spec/classes/foreman_plugin_openscap_spec.rb
mcanevet: Yes, that's the idea.

manifests/plugin/openscap.pp
```diff
@@ -0,0 +1,45 @@
...
+#
+# This class installs OpenSCAP plugin
+#
+# === Parameters:
+#
+# $configure_openscap_repo::  Enable custom yum repo with packages needed for foreman_openscap,
+#                             type:boolean
+#
+class foreman::plugin::openscap (
+  $configure_openscap_repo = $foreman::plugin::openscap::params,
```
domcleal: missing `::configure_openscap_repo` on the end

domcleal:
I hit the above bug when testing, but otherwise seems fine, please squash and rebase after fixing that.

(Incidentally, the plugin blew up with a ffi RuntimeError when loading the page, but I'll go file that some place else.)

ares:
Fixed hopefully all, squashed, green.

manifests/plugin/openscap/params.pp
```diff
@@ -0,0 +1,4 @@
+# Default parameters for foreman::plugin::openscap
+class foreman::plugin::openscap::params {
+  $configure_openscap_repo = $::osfamily == 'RedHat'
```
mmoll: that's slick :dart:

mmoll:
merged as b31f38d, thanks ares :tada:

PR #274, opened 2015-01-20
update for non-english windows clients

carrolcox:
foreman 1.7.1, puppet 3.7.3
1) Default external encoding ( global )
2) filter any non-printable char from the value, if it is a String ( def build_body(certname,filename) )
    http://projects.puppetlabs.com/issues/12702#note-8

### DIFF
--- a/files/external_node_v2.rb
+++ b/files/external_node_v2.rb
@@ -14,7 +14,9 @@ $settings_file = "/etc/puppet/foreman.yaml"
 SETTINGS = YAML.load_file($settings_file)
 
 # Default external encoding
-Encoding.default_external = Encoding::UTF_8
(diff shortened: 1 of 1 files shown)


files/external_node_v2.rb
```diff
@@ -13,6 +13,9 @@
 
 SETTINGS = YAML.load_file($settings_file)
 
+# Default external encoding
+Encoding.default_external = Encoding::UTF_8
```
domcleal: add `if defined?(Encoding)` for Ruby 1.8 compatibility

mmoll:
I don't have a Windows to test, but looks good :+1:

domcleal:
Merged, thanks for the contribution @carrolcox!

PR #276, opened 2015-01-26
Openscap plugin scap client

ares:
This PR would allow to install foreman_scap_client puppet module to puppet master

Based on #273, link to original discussion https://github.com/theforeman/puppet-foreman/pull/273#discussion_r23518477

There are other possible approaches:
* move the logic to puppet-puppet module, since it's more related to puppetmaster than foreman itself
* implement as foreman-installer hook since it's mainly intended for installer support, on the other hand why not allow this to be used within puppet master as well

@ekohl what are your thoughts? Should we raise broader discussion on mailing list first?

### DIFF
--- a/manifests/plugin/openscap.pp
+++ b/manifests/plugin/openscap.pp
@@ -1,0 +1,46 @@
+# = Foreman OpenSCAP plugin
+#
+# This class installs OpenSCAP plugin
+#
+# === Parameters:
+#
+# $configure_openscap_repo::  Enable custom yum repo with packages needed for foreman_openscap,
+#                             type:boolean
+#
+class foreman::plugin::openscap (
+  $configure_openscap_repo = true,
+) {
+  validate_bool($configure_openscap_repo)
+
+  case $::osfamily {
+    'RedHat': {
+
+      if $configure_openscap_repo {
+        $repo = $::operatingsystem ? {
+          'Fedora' => 'fedora',
+          default  => 'epel',
+        }
+
+        if $::operatingsystemmajrelease == undef {
+          $versions_array = split($::operatingsystemrelease, '\.') # facter 1.6
+          $major = $versions_array[0]
+        } else {
+          $major = $::operatingsystemmajrelease # facter 1.7+
+        }
+
+        yumrepo { 'isimluk-openscap':
+          enabled  => 1,
+          gpgcheck => 0,
+          baseurl  => "http://copr-be.cloud.fedoraproject.org/results/isimluk/OpenSCAP/${repo}-${major}-\$basearch/",
+          before   => [ Foreman::Plugin['openscap'] ],
+        }
+      }
+
+      foreman::plugin {'openscap': }
+
+    }
+    default: {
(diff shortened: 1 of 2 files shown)


domcleal:
I've been thinking about this a little and I'm fairly sure that installation of modules on a master fits into puppet-puppet as a new resource.  Consider that we have puppet::server::env to add an environment, we could conceivably have puppet::server::module which would add a module to a given environment, or into the shared dir /usr/share/puppet/modules/.

I see two ways of distributing the module itself:

1. OS packages: both [Debian](https://packages.debian.org/sid/puppet-module-puppetlabs-stdlib) and [Fedora](https://apps.fedoraproject.org/packages/puppetlabs-stdlib) have packaged Puppet modules already, we could adopt their standards.
1. https://github.com/ryanycoleman/puppet_module_provider/ using `puppet module`

Of course, we could support both, no need to pick only one.

isimluk:
Creating puppet-foreman_scap_client OS package seems superior to me. puppet_module_provider would be yet another dependency in the great scheme of things and it would need to get in place somehow.

domcleal:
Well, adding a dependency is a single line in a JSON file, it isn't a big deal.

ekohl:
I'm still not entirely clear on the goal. Is it to configure the SCAP-client on the clients (puppet nodes)? For that you have a puppet module which you'd like to install on the master and then apply on the client using regular puppet runs. Am I understanding it right? If so, then this is step 1 (installing the module on the master), but you still need it in the catalog somehow.

domcleal:
I guess the foreman_openscap plugin is capable of that?  Maybe it does it.

ares:
@ekohl  @domcleal  yes, foreman_openscap plugin assign the class when you assign hostgroup in policy wizard

isimluk:
FYI: spec file for foreman_scap_client puppet module is brewing here: https://github.com/isimluk/foreman-packaging/blob/puppet-foreman_scap_client/puppet-foreman_scap_client/puppet-foreman_scap_client.spec

ohadlevy:
@ares @domcleal what is missing on this PR ?

domcleal:
In the discussion we decided it's the wrong direction.

ares:
We did? My understanding is that @domcleal suggested to package the module separately, @ekohl had questions that I answered and @isimluk started to package it separately so we are ready if this is the consensus. So @ekohl if you agree with @domcleal then I suppose we'll have to finish packaging of puppet module and change this to optionally install it.

domcleal:
Yeah, and this PR currently copies files from foreman-installer, so packaging it properly is the right (and different) direction.  I still think it should be puppet-puppet managing the contents of the env though, not puppet-foreman.

ekohl:
I agree with @domcleal that packaging and manage the package though puppet is a better way to go. I'd still be hesitant to attempt to manage puppet environments since there are so many different ways to manage it (mostly involve git, r10k and/or librarian) and users can change their environment paths that. That can subtly break assumptions.

ares:
ok, so closing this PR, we'll package it separately and if needed we'll install it from puppet-puppet module

PR #111, opened 2013-09-13
[WIP] Add a foreman subnet type

ekohl:
By adding a foreman subnet type, you can ensure a subnet exists in your
foreman instance.

PR #280, opened 2015-01-31
Add a foreman subnet type

ekohl:
By adding a foreman subnet type, you can ensure a subnet exists in your foreman instance.

TODO:
- [ ] Allow clearing of domains
- [ ] Flush all values at once

This replaces https://github.com/theforeman/puppet-foreman/pull/111 since I no longer have access to the repository I created it from.

### DIFF
--- a/lib/puppet/provider/foreman_subnet/rest.rb
+++ b/lib/puppet/provider/foreman_subnet/rest.rb
@@ -1,0 +1,263 @@
+Puppet::Type.type(:foreman_subnet).provide(:rest) do
+
+  confine :feature => :foreman_api
+
+  def subnets
+    ForemanApi::Resources::Subnet.new({
+      :base_url => resource[:base_url],
+      :oauth    => {
+        :consumer_key    => resource[:consumer_key],
+        :consumer_secret => resource[:consumer_secret]
+      }
+    }, {
+      :headers => {
+        :foreman_user => resource[:effective_user]
+      }
+    })
+  end
+
+  def subnet
+    if @subnet
+      @subnet
+    else
+      spi = subnets.index[0]
+      if spi.is_a?(Hash) && spi.has_key?('results')
+        # foreman_api 0.1.18+
+        @subnet = spi['results'].find { |s| s['name'] == resource[:name] }
+      else
+        # foreman_api < 0.1.18
+        @subnet = spi.find { |s| s['subnet']['name'] == resource[:name] }['subnet']
+      end
+    end
+  end
+
+  def id
+    subnet ? subnet['id'] : nil
+  end
+
+  def exists?
+    !!subnet
+  end
+
+  def create
+    subnets.create({
+      'subnet' => {
+        'name' => resource[:name],
+        'network' => resource[:network],
+        'mask' => resource[:netmask],
+        'gateway' => resource[:gateway],
+        'dns_primary' => resource[:dns_primary],
+        'dns_secondary' => resource[:dns_secondary],
+        'from' => resource[:from],
+        'to' => resource[:to],
+        'vlanid' => resource[:vlan],
+        'dns_id' => smartproxy_id(resource[:dns_proxy]),
+        'dhcp_id' => smartproxy_id(resource[:dhcp_proxy]),
+        'tftp_id' => smartproxy_id(resource[:tftp_proxy]),
+      }
+    })
+  end
+
+  #def flush
+  #  subnets.update({
+  #    'id' => @property_hash[:id],
+  #    'subnet' => {
+  #      'name' => @property_hash[:name],
+  #      'network' => @property_hash[:network],
+  #      'mask' => @property_hash[:netmask],
+  #      'gateway' => @property_hash[:gateway],
+  #      'dns_primary' => @property_hash[:dns_primary],
+  #      'dns_secondary' => @property_hash[:dns_secondary],
+  #      'from' => @property_hash[:from],
+  #      'to' => @property_hash[:to],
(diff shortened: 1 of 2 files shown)


ekohl:
My knowledge of types and providers is limited so I could use some help with the TODO items.

ekohl:
@domcleal I'd greatly appreciate some pointers if you have time.

lib/puppet/provider/foreman_subnet/rest.rb
```diff
@@ -0,0 +1,263 @@
+Puppet::Type.type(:foreman_subnet).provide(:rest) do
+
+  confine :feature => :foreman_api
+
+  def subnets
+    ForemanApi::Resources::Subnet.new({
```
domcleal: We ought to use apipie-bindings nowadays :)

lib/puppet/provider/foreman_subnet/rest.rb
```diff
@@ -0,0 +1,263 @@
...
+        :consumer_secret => resource[:consumer_secret]
+      }
+    }, {
+      :headers => {
+        :foreman_user => resource[:effective_user]
+      }
+    })
+  end
+
+  def subnet
```
domcleal: make it a private method, either `private :subnet` underneath or put it in a `private` block at the bottom of the class

lib/puppet/provider/foreman_subnet/rest.rb
```diff
@@ -0,0 +1,263 @@
...
+    if @subnet
+      @subnet
+    else
+      spi = subnets.index[0]
```
domcleal: nitpick that we ought to solve in the proxy provider too, but we could use a search query (name = foo) rather than pulling all results and iterating through them

lib/puppet/provider/foreman_subnet/rest.rb
```diff
@@ -0,0 +1,263 @@
...
+  #  })
+  #end
+
+  def destroy
+    subnets.destroy({'id' => id})
+    @subnet = nil
+  end
+
+  def network
+    subnet ? subnet['network'] : nil
```
domcleal: doesn't make any difference, but `subnet['network'] if subnet` would achieve the same

lib/puppet/provider/foreman_subnet/rest.rb
```diff
@@ -0,0 +1,263 @@
...
+  end
+
+  ## Helper functions
+  ##
+
+  private
+
+  def api_domains
+    @api_domains ||= ForemanApi::Resources::Domain.new({
+      :base_url => resource[:base_url],
```
domcleal: duplicates the subnet code at the top, this whole hash could be in one method

lib/puppet/provider/foreman_subnet/rest.rb
```diff
@@ -0,0 +1,263 @@
...
+    }, {
+      :headers => {
+        :foreman_user => resource[:effective_user]
+      }
+    })
+  end
+
+  def domain_name(id)
+    spi = api_domains.index[0]
+    if spi.is_a?(Hash) && spi.has_key?('results')
```
domcleal: also lots of duplication

lib/puppet/provider/foreman_subnet/rest.rb
```diff
@@ -0,0 +1,263 @@
...
+  end
+
+  def smartproxy_name(id)
+    spi = api_smartproxies.index[0]
+    if spi.is_a?(Hash) && spi.has_key?('results')
+      # foreman_api 0.1.18+
+      result = spi['results'].find { |s| s['id'] == id }
+    else
+      # foreman_api < 0.1.18
+      result = spi.find { |s| s['smartProxy']['id'] == id }['smartProxy']
```
domcleal: capitalisation here is odd, I wouldn't expect it, but this might go away with apipie-bindings :)

lib/puppet/provider/foreman_subnet/rest.rb
```diff
@@ -0,0 +1,263 @@
...
+  def smartproxy_id(name)
+    spi = api_smartproxies.index[0]
+    if spi.is_a?(Hash) && spi.has_key?('results')
+      # foreman_api 0.1.18+
+      result = spi['results'].find { |s| s['name'] == name }
+    else
+      # foreman_api < 0.1.18
+      result = spi.find { |s| s['smartProxy']['name'] == name }['smartProxy']
+    end
+    result['id']
```
domcleal: This might fail if there's no result

domcleal:
I like the look of it, the main things are to get it over to apipie-bindings which might reduce the amount of code to handle results formatting and perhaps to add some validation into the type?

I wouldn't worry about prefetch and flush, although it may be useful, I don't think it's necessary for inclusion.

I'd have thought that using array_matching => all on the list of domains would ensure that you could empty the list of domains, unless the API isn't handling the `domain_ids => []` subnet update correctly and is ignoring the value.

I'm a little unsure about using other resources in the *proxy properties, I would probably need to test that a bit.  I seem to remember they act strangely depending on what stage of the catalog compilation/application you're at, but I can't remember which way around it is (whether you get a stub object during compilation or application).

ekohl:
flushing is needed because you can get weird validation errors. If you provider a start range, you also need to provide an end range. The current implementation can only do that on creation but will fail on updates.

domcleal:
Ah yes, good point.

ekohl:
I'm going to close this now since I no longer have a specific need for it and no time to finish.

PR #281, opened 2015-01-31
Allow foreman::plugin to manage a config file

ekohl:


manifests/plugin.pp
```diff
@@ -15,4 +17,16 @@
...
+
+  if $config {
+    file { $config_file:
+      ensure  => file,
+      owner   => 'root',
+      group   => 'root',
+      mode    => '0644',
+      content => $config,
+      require => Package[$real_package],
+      notify  => Class['foreman::service'],
```
domcleal: `foreman::plugin <| |>` already notifies the service from init.pp.  (Actually, line 18 too.)

ekohl: Would you remove line 18 as well then?

domcleal: Yeah :)

ekohl: Done

ekohl:
So we need a convention about the config file. @ehelms and @komidore64 both suggested to use foreman_something.yaml instead of something.yaml as it is now implemented. I don't mind either, but do want a general convention.

Current implementation is IMHO easier and less redundant, but the full name does match the gem name also has benefits for consistency.

/cc @domcleal @mmoll any preference?

mmoll:
I don't mind either.

komidore64:
@ekohl or suggestion to use `foreman_gutterball.yaml` is because that's the full name of the [plugin](https://github.com/Katello/foreman-gutterball). down the line, if someone creates a foreman plugin titled `something`, then i would expect its config file to be called `something.yaml`.

ekohl:
@komidore64 currently that would work if a user specifies ```foreman::plugin {'something': package_name => 'something' }```. I'd strongly prefer if all foreman plugins did use ```foreman_``` as prefix though.

komidore64:
understandable

domcleal:
I don't particularly mind about the configuration file name, but should note that all of the RPMs use the %{gem_name}.yaml, so the Puppet module and packages need to match - else we'll get two configuration files.

ekohl:
So I think the consensus is leaning to the full gem name. The (technical) problem with that is that we prefix it using ```$foreman::plugin_prefix``` which could also include the SCL. Back when I introduced that variable I considered splitting it in two but decided against it. Do you think we should hardcode foreman_ here in the parameter or prefer a variable?

domcleal:
I think I prefer just hardcoding 'foreman_' into the config_file parameter default value, like we do with the package prefix.  (Yeah, package prefix is a variable of sorts, but only for OS-based resolution.)

ekohl:
How about this? I put /etc/foreman/plugins in a variable in params.pp so it's easier to override only the filename.

domcleal:
:+1: could you just add a test case for it?

ekohl:
Test added.

spec/defines/foreman_plugin_spec.rb
```diff
@@ -29,10 +30,7 @@
     } end
 
     it 'should install the correct package with notify' do
```
mmoll: "with notify"?

mmoll: I suppose the plugin packages should restart the core app now in their postinst (iirc RPMs do this, DEBs not really)?

ekohl: Right, so now the notify is gone that part of the description no longer makes sense.

domcleal: Are you fixing the desc?

ekohl: Thanks for the reminder.

domcleal:
Thanks @ekohl!

Issue #161, opened 2014-02-27
foreman::puppetmaster should restart service/http when foreman.rb changes


DavidS:
foreman::puppetmaster should restart service/http when foreman.rb changes as puppet will not reload the report processor by itself.

PR #282, opened 2015-02-02
Notify the puppet server when changing foreman.rb

ekohl:
Closes GH-161

I just noticed I pushed it to the wrong repository ...


mmoll:
this doesn't work with passenger.

ekohl:
I think we should solve that in the puppet-puppet module. @domcleal any opinion?

ekohl:
@mmoll submitted another PR to take care of the passenger part.

domcleal:
Do we need to worry about people using this class without our 'puppet' module?

ekohl:
We already have requires later in the class so I don't think we should worry.

domcleal:
Ah yes!

domcleal:
Merged, thanks @ekohl!  We'll continue to work on the puppet-puppet part.

PR #286, opened 2015-02-03
Adding option to not have module try to add the user for foreman

proofpoint-mlariz:
Default behavior is the same as current. However this option allows the option to have user managed in another module. Useful for instances that want UIDs managed centrally to ensure consistency across systems.

### DIFF
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -80,6 +80,8 @@
 #
 # $app_root::                 Name of foreman root directory
 #
+# $manage_user::              Controls whether foreman module will manage the user on the system. (default true)
+#
 # $user::                     User under which foreman will run
 #
 # $group::                    Primary group for the Foreman user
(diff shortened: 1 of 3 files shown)


manifests/init.pp
```diff
@@ -80,6 +80,8 @@
 #
 # $app_root::                 Name of foreman root directory
 #
+# $manage_user::              Controls whether foreman module will manage the user on the system. (default true)
```
ekohl: Please add a ```type:boolean``` so kafo knows the right type.

manifests/init.pp
```diff
@@ -189,6 +191,7 @@
   $db_sslmode               = 'UNSET',
   $db_pool                  = $foreman::params::db_pool,
   $app_root                 = $foreman::params::app_root,
+  $manage_user              = true,
```
ekohl: I'd prefer to centralize this in params.

ekohl:
Just some smaller technical issues, but in general this can be merged.

mmoll:
thanks @proofpoint-mlariz :tada:

PR #287, opened 2015-02-04
Leading zeros need to be quoted

raphink:
This fix is required to migration to the future parser.


ekohl:
:+1: pending green tests

mmoll:
thanks @raphink :tada:
and boo puppet-lint ;)

PR #289, opened 2015-02-10
Fixes #9286 - node.rb timeout too low and not configurable

jistr:
Make the node.rb timeout configurable. This is useful for use cases when
catalog compilation takes more time and/or the node.rb request to
Foreman is being handled by a freshly spawned Passenger worker.

### DIFF
--- a/manifests/params.pp
+++ b/manifests/params.pp
@@ -53,6 +53,8 @@ class foreman::params {
   $gpgcheck    = true
   $version     = 'present'
 
+  $puppetmaster_timeout = 60
+
   # when undef, foreman-selinux will be installed if SELinux is enabled
   # setting to false/true will override this check (e.g. set to false on 1.1)
   $selinux     = undef
--- a/manifests/puppetmaster.pp
+++ b/manifests/puppetmaster.pp
@@ -11,6 +11,7 @@ class foreman::puppetmaster (
   $puppet_user      = $foreman::params::puppet_user,
   $puppet_group     = $foreman::params::puppet_group,
   $puppet_basedir   = $foreman::params::puppet_basedir,
+  $timeout          = $foreman::params::puppetmaster_timeout,
   $ssl_ca           = $foreman::params::client_ssl_ca,
   $ssl_cert         = $foreman::params::client_ssl_cert,
   $ssl_key          = $foreman::params::client_ssl_key,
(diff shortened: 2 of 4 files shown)


jistr:
Needed by https://github.com/theforeman/puppet-puppet/pull/236 and https://github.com/theforeman/foreman-installer-staypuft/pull/134

ekohl:
I'm not quite sure about the noderb in the name, but it looks like this is only used in node.rb right now. Other than that it looks good.

manifests/params.pp
```diff
@@ -53,6 +53,8 @@
   $gpgcheck    = true
   $version     = 'present'
 
+  $noderb_timeout = 10
```
domcleal: We should probably increase this default to avoid the Passenger startup issue you describe?

domcleal:
@ekohl I agree, we could (should) use the timeout in the report processor in future.

jistr:
Ack, so what should be the name of the variable? Just `timeout`?

Re the default value -- we'll probably set it to 60 seconds in Staypuft, sounds ok as default or do you want a lower number?

domcleal:
I think just `timeout` is fine on foreman::puppetmaster as the context is clear.  We might want a prefix of some sort in puppet-puppet, but I can't think of one right now...

Regarding the value itself, we actually have a timeout of 600 seconds for starting Passenger, though this is probably a bit excessive for a client timeout.  60 seconds would be fine by me.

jistr:
@domcleal on puppet-puppet side -- `server_request_timeout`?

domcleal:
:+1:

ekohl:
Sounds good.

jistr:
Travis says all is well :) I'll be away for about 1.5 hrs now and then i'll fix up the other PR.

(I've noticed that the tests on my puppet-puppet pull request failed previously because they couldn't find the parameter in puppet-foreman, i guess it was only because that they use master of puppet-foreman and the error would disappear when we merge this PR.)

domcleal:
:+1:

manifests/params.pp
```diff
@@ -52,6 +52,7 @@
   $environment = 'production'
   $gpgcheck    = true
   $version     = 'present'
+  $timeout     = 60
```
ekohl: I'm not sure I like this generic name here. I'd name it ```puppetmaster_timeout```. In puppetmaster itself the parameter should still be just timeout.

jistr: Ok i can change that. @domcleal do you agree on this change?

domcleal: sure!

ekohl:
:+1:

domcleal:
Merged, thanks @jistr!

jistr:
Thanks for the reviews!

PR #293, opened 2015-02-19
modulesync: new lint checks, add CONTRIBUTING doc

domcleal:
Needs fixing!


ekohl:
Rebased, but I don't know if the remaining can be fixed or should be silenced.

ekohl:
Also, ```Skipping metadata validation; the metadata-json-lint gem was not found``` implies we should add that gem for additional validation.

domcleal:
@ekohl yeah, it got pulled out of puppetlabs_spec_helper's dependencies before release, much to my disappointment

domcleal:
1. I fixed one spaceship operator lint issue by only realising foreman_smartproxy resources that need to be registered on this particular Foreman server.  It'll make it a little more fragile if somebody uses foreman_proxy and doesn't have the two base URLs set, but I think it's probably OK.
2. I had to undo one change as it threw a compilation error in that conditional on Ubuntu 12.04, as `::alternatives` isn't a valid tag (the difference here being alternatives is a type, not defined type).
3. One spaceship error remains on foreman::plugin, but I don't know how we should fix it, so I think we should probably put a control comment around it for now and ignore the problem.  If we tag all our foreman::plugin instances, whenever somebody then uses foreman::plugin then they won't get the benefit of the ordering we do in init.pp.  I think that would be worse than realising virtual resources for now.

ekohl:
Regarding ```foreman::plugin```, I think you could use ```require``` but I don't think you can do the notify inside the define so I'd agree with a control comment to ignore it for now.

manifests/config/passenger/fragment.pp
```diff
@@ -7,23 +7,25 @@
...
   require foreman::config::passenger
 
-  file { "${apache::confd_dir}/05-foreman.d/${name}.conf":
-    ensure  => present,
-    content => $content,
-    owner   => 'root',
-    group   => 'root',
-    mode    => '0644',
-    notify  => Class['apache::service'],
+  if $content {
```
ekohl: I was hesitant to do this because I thought you could no longer wipe them, but I now remember that puppetlabs-apache purges unmanaged files so it should be fine, right?

domcleal: Ah, excellent point.  Yeah, it recursively purges: https://github.com/puppetlabs/puppetlabs-apache/blob/master/manifests/init.pp#L162-L163

I was wondering whether to add an `else` block that ensures the file is absent, but that's probably OK without it.

ekohl: It's most like OK, but since it's configurable I'm leaning to explicit. Maybe determine the ensure status based on ```$content```? I think that's cleaner than an else block. Also, puppet 4 will still write an empty file since ```''``` will evaluate to true.

domcleal: I think I'd prefer an else, otherwise we'd be specifying a file resource with mode, owner and content, but with ensure => absent, which feels strange.  (And Puppet would be justified in throwing a compilation or runtime error I'd say.)

I'll add a check for `''` too, good point.

ekohl: Fair enough.

domcleal: Updated.

ekohl:
:+1: once the tests pass.

domcleal:
Merged as ddfaacd..b80a3b6, thanks @ekohl!

PR #295, opened 2015-02-19
Initial puppet runs with no internet access cause failures on later runs

eimirae:
First pull request, let me know if I am doing anything wrong.

On my first puppet runs, I do not have internet access, and I discovered that the resource ordering in the apt repo installation causes later runs to fail as the theforeman deb sources never gets changed (which succeeds offline).

### DIFF
--- a/manifests/install/repos/apt.pp
+++ b/manifests/install/repos/apt.pp
@@ -2,10 +2,10 @@
 define foreman::install::repos::apt ($repo) {
   file { "/etc/apt/sources.list.d/${name}.list":
     content => "deb http://deb.theforeman.org/ ${::lsbdistcodename} ${repo}\ndeb http://deb.theforeman.org/ plugins ${repo}\n"
-  } ~>
+  } ->
   exec { "foreman-key-${name}":
-    command     => '/usr/bin/wget -q http://deb.theforeman.org/foreman.asc -O- | /usr/bin/apt-key add -',
-    refreshonly => true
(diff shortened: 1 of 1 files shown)


ekohl:
I think it looks good. @mmoll and @domcleal?

mmoll:
@eimirae I'd rather grep for the KeyID, looking good otherwise, thanks for raising this.

@ekohl @domcleal As an "out of scope for this PR" comment, apt_key from puppetlabs-apt could be used, or was/is there a reason why this was handled via exec?

lazyfrosch:
I'd prefer using puppetlabs-apt as well :)

domcleal:
@mmoll I'm unsure about how to handle key IDs, as they will probably change on a semi-regular basis.

I don't think there's a good reason any more to avoid pl-apt.  I suspect this was done probably just as a quick hack to avoid adding a dependency, but it's one we've shipped in foreman-installer for a long time now as other modules depend on it.

lazyfrosch:
@domcleal well, we can always ensure old keys away later... :)

But I think it might be safer to specify the key IDs inside params.

domcleal:
Yeah, doing it through a param would be OK too.

mmoll:
So we merge this and @lazyfrosch will do the porting to puppetlabs-apt? ;)

ekohl:
I'd be fine with that :)

lazyfrosch:
You know that feeling when you said something wrong? :runner:

Deadline for 1.8 is? :-)

mmoll:
@lazyfrosch Monday, so you have all the weekend :)

mmoll:
merged, thanks @eimirae!

PR #296, opened 2015-02-20
fixes #9449 - drop usage of lsb facts

mmoll:
where possible, so there's still one occurence in ```manifests/install/repos/apt.pp```.
Support for Debian/squeeze was dropped for Foreman 1.7, so desupport it here as well.

### DIFF
--- a/manifests/config/passenger.pp
+++ b/manifests/config/passenger.pp
@@ -75,11 +75,8 @@ class foreman::config::passenger(
   # RedHat distros come with this enabled. Newer Debian and Ubuntu distros
   # comes also with this enabled. Only old Debian and Ubuntu distros (squeeze,
   # lucid, precise) needs hand-holding.
-  case $::lsbdistcodename {
-    'squeeze','lucid','precise': {
-      ::apache::mod { 'version': }
-    }
-    default: {}
+  if ($::operatingsystemrelease == '12.04') and ($::operatingsystem == 'Ubuntu') {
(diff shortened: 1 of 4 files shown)


manifests/config/passenger.pp
```diff
@@ -75,11 +75,8 @@
   # RedHat distros come with this enabled. Newer Debian and Ubuntu distros
   # comes also with this enabled. Only old Debian and Ubuntu distros (squeeze,
   # lucid, precise) needs hand-holding.
-  case $::lsbdistcodename {
-    'squeeze','lucid','precise': {
-      ::apache::mod { 'version': }
-    }
-    default: {}
+  if $::operatingsystemrelease == '12.04' {
```
ekohl: While unlikely, maybe also check if the operatingsystem is ubuntu?

manifests/puppetmaster.pp
```diff
@@ -20,13 +20,8 @@
 ) inherits foreman::params {
 
   case $::operatingsystem {
```
ekohl: While you're at it, change it to osfamily?

mmoll: done

manifests/config/passenger.pp
mmoll: done

ekohl:
:+1:

domcleal:
Thanks @mmoll!

PR #297, opened 2015-03-02
Make CRL optional in use with kafo

ares:
When this class is used from kafo we don't have a way to disable CRL
configuration, since the parameter is String. It can't be set to false
and it's always present.

### DIFF
--- a/manifests/config/passenger.pp
+++ b/manifests/config/passenger.pp
@@ -127,9 +127,12 @@ class foreman::config::passenger(
         true  => "https://${servername}",
         false => undef,
       }
-      $ssl_crl_check = $ssl_crl ? {
-        undef   => undef,
-        default => 'chain',
+      if $ssl_crl and $ssl_crl != '' {
+        $ssl_crl_real = $ssl_crl
+        $ssl_crl_check = 'chain'
+      } else {
+        $ssl_crl_real = undef
+        $ssl_crl_check = undef
       }
 
       file { "${apache::confd_dir}/05-foreman-ssl.d":
@@ -160,7 +163,7 @@ class foreman::config::passenger(
         ssl_key                 => $ssl_key,
         ssl_chain               => $ssl_chain,
         ssl_ca                  => $ssl_ca,
-        ssl_crl                 => $ssl_crl,
+        ssl_crl                 => $ssl_crl_real,
         ssl_crl_check           => $ssl_crl_check,
         ssl_verify_client       => 'optional',
         ssl_options             => '+StdEnvVars',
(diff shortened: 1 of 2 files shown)


domcleal:
:+1:

manifests/config/passenger.pp
```diff
@@ -127,9 +127,15 @@
         true  => "https://${servername}",
         false => undef,
       }
-      $ssl_crl_check = $ssl_crl ? {
-        undef   => undef,
-        default => 'chain',
+      if $ssl_crl == '' or !$ssl_crl {
```
ekohl: nitpick: I'd reverse the order because positive is easier to read: ```if $ssl_crl and $ssl_crl != '' {```

manifests/config/passenger.pp
```diff
@@ -127,9 +127,15 @@
...
+        $ssl_crl_real = undef
+        $ssl_crl_check = undef
+      } else {
+        $ssl_crl_real = $ssl_crl
+        $ssl_crl_check = $ssl_crl ? {
```
ekohl: Since we know it's set, can't this be hardcoded to ```'chain'``` now?

ares: I think that would limit us in using other values?(http://httpd.apache.org/docs/current/mod/mod_ssl.html#sslcarevocationcheck)

domcleal: We don't allow the user to set those values though, it's either chain or nothing.

manifests/config/passenger.pp
```diff
@@ -127,9 +127,15 @@
         true  => "https://${servername}",
         false => undef,
       }
-      $ssl_crl_check = $ssl_crl ? {
-        undef   => undef,
-        default => 'chain',
+      if $ssl_crl and $ssl_crl != '' {
+        $ssl_crl_real = $ssl_crl
+        $ssl_crl_check = $ssl_crl ? {
```
ekohl: This can be set to ```'chain'``` since we already know ```$ssl_crl``` is something trueish.

manifests/config/passenger.pp
ares: yeah I meant in future, whatever, I'll hardcode `'chain'`

manifests/config/passenger.pp
ares: fixed, thanks

ekohl:
:+1: when the tests pass.

ares:
tests are green

domcleal:
Thanks @ares!

PR #298, opened 2015-03-04
Change apipie rake task to rebuild API doc indexes only

domcleal:
[Ticket #4478](http://projects.theforeman.org/issues/4478) in Foreman 1.8 changed our packages to provide prebuilt API docs
for resources, so Foreman packages provide hosts, domains etc, and plugin
packages provide things like containers, disks etc.

Packages and the installer then need to call apipie:cache:index to rebuild just
the indexes instead of needing to rebuild every resource file, resulting in a
large speed boost (especially as API docs are now multilingual).

The apipie_task parameter retains 1.7 compatibility, as all API docs need
building on this version.

---

Marked as do not merge as the Debian packaging PR is open: https://github.com/theforeman/foreman-packaging/pull/568.  I hope to have done shortly so we don't have this 1-2 minute rake task in our default 1.8 installation.

### DIFF
--- a/README.md
+++ b/README.md
@@ -49,6 +49,8 @@ previous stable release.
 
 ### Foreman 1.7 compatibility notes
 
+* set `apipie_task => 'apipie:cache'` as Foreman 1.7 packages didn't have
+  precompiled API docs
 * `foreman::compute::ec2` needs `package => 'foreman-compute'` on Foreman 1.7,
   as the package has been renamed in newer versions.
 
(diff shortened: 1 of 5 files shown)


domcleal:
Packaging change now in nightlies and I aim to have it in 1.8.0.

ekohl:
I didn't test it, but :+1: to the code and the change.

domcleal:
I've ran an end to end test now on an EL6 nightly installation, Hammer works after install.

mmoll:
merged as 25168bb, thanks @domcleal

Issue #129, opened 2013-11-27
Attempt to use the module standalone causes failure due to implicit dependency on the theforeman/puppet module


danielkza:
Hello,

I'm trying to set up The Foreman with an existing Puppet infrastructure. Part of the master bootstrapping is done through puppet itself, so I wanted to have the foreman proxy and reporting configuration done like that as well.

So I installed the theforeman/foreman module first, and tried to include the 'foreman::puppetmaster' class in the master to get things set up automatically, but I get the following error:

```
Error: Could not retrieve catalog from remote server: Error 400 on SERVER: Invalid relationship: File[/etc/puppet/node.rb] { require => Class[Puppet::Server::Install] }, because Class[Puppet::Server::Install] doesn't seem to be in the catalog
```

which seems to me like theforeman/foreman depends implicitly (but not explicitly in the Modulefile) on theforeman/puppet, but I don't actually want it to manage the whole puppet installation for me, at least for now.

ekohl:
We've improved on this and now document the (optional) dependency in the README. This is in line with the current [puppet style guide](https://docs.puppetlabs.com/guides/style_guide.html#dependencies):
> Soft dependencies should be called out in the README.md, and must not be enforced as a hard requirement in your metadata.json. A soft dependency is a dependency that is only required in a specific set of use cases.

I'm closing this now, but feel free to re-open if you think this should be addressed further.

njm506:
While the update to document the soft dependency is definitely a good move, the original issue still stands - the relationships to '::puppet::server::*' are still present in puppetmaster.pp, and these are only valid if the foreman Puppet module is in use.

There's no functional dependency between the modules, yet it still effectively forces users to define a module with that name, just to make the relationship valid.  I would strongly recommend that you remove these relationships, and instead document that the necessary relationships need to be set up externally, to work with whatever module is being used to configure the puppetmaster (which might not be the Foreman one).

ekohl:
Maybe hide them behind a parameter that defaults to true? Personally I always wanted to move this puppetmaster class to another module and often considered puppet-puppet, but maybe puppet-foreman_proxy would be a better fit. I assume you don't use our puppet-puppet so do you think that would be a good move?

We've also talked about packaging the files and maybe with the upcoming puppet 4 release that might be needed anyway. Then the whole dependencies would move to the package manager and can be solved much cleaner. I now wonder if the notification I just added in #282 should live in puppet-puppet as well, like @domcleal mentioned.

[...]

PR #300, opened 2015-03-06
Remove ordering constraints around theforeman/puppet

domcleal:
Depends on https://github.com/theforeman/puppet-puppet/issues/248

foreman::puppetmaster should be usable without our 'puppet' module.  Instead
for theforeman/puppet users, puppet::server::config is being updated to
correctly anchor the foreman::puppetmaster class.

Closes GH-129

---

I tested the two together just now, changed /etc/puppet/node.rb, re-ran the installer and it correctly refreshed the puppet master service, so the notification part of anchoring is working correctly.


ekohl:
:+1:

ekohl:
Thanks!

PR #301, opened 2015-03-06
adjust puppetlabs-postgresql version constraint

mmoll:
I tested the installer with puppetlabs-postgresql 4.1.0 on CentOS 7, Debian/wheezy and Ubuntu/trusty. Also, the changelog of 4.0.0 doesn't seem to list anything incompatible for this module.


mmoll:
I'd like to have this in 3.0.0

domcleal:
Very good catch - we actually used 4.x in Foreman 1.7 as the version req in 2.3.x permitted it.  Thanks @mmoll!

PR #302, opened 2015-03-07
Remove email separator from lookup_identiy.conf.erb

lamawithonel:
The space-separated string of email addresses generated by
*mod_lookup_identity* fails validation by Foreman during on-the-fly
registration causing this error:

> **Warning!**
>
> Validation failed: Email address is invalid, Email address is too long
> (maximum is 60 characters)

This occurs even when the resultant string is less than 60 characters.

This commit removes the concatenation separator to prevent the
concatenation behavior.  Instead, *mod_lookup_identity* will return just
the first address for the user.  Foreman will prompt the user for
confirmation so they are still free to input another address.

### DIFF
--- a/templates/lookup_identity.conf.erb
+++ b/templates/lookup_identity.conf.erb
@@ -1,6 +1,6 @@
 
 <LocationMatch ^/users/(ext)?login$>
-  LookupUserAttr email REMOTE_USER_EMAIL " "
+  LookupUserAttr email REMOTE_USER_EMAIL
   LookupUserAttr firstname REMOTE_USER_FIRSTNAME
   LookupUserAttr lastname REMOTE_USER_LASTNAME
   LookupUserGroupsIter REMOTE_USER_GROUP


mmoll:
According to some googling, this feels OK, but I have no idea of the inital reasoning nor using IPA at the moment.

domcleal:
@adelton would you mind checking this please?

adelton:
> Foreman will prompt the user for confirmation

This shouldn't be happening -- if the user has email address set and propagated via REMOTE_USER_EMAIL, it should just set that to the internal user record in the EXTERNAL auth source and should not prompt the user to do anything.

adelton:
The patch is correct if the goal is really to just use one email address even if the user has multiple addresses in IPA.

So the question for you guys is -- is it correct that in Foreman, the user can only have one email address set? Shouldn't the effort go towards the user being able to have multiple email addresses (one regular email, one routed to their phone, for example)?

Of course, if the email field in Foreman is limited to 60 characters, we will probably hit that limit with some long email addresses so how that should be addressed is another question (for another PR).

domcleal:
> So the question for you guys is -- is it correct that in Foreman, the user can only have one email address set? Shouldn't the effort go towards the user being able to have multiple email addresses (one regular email, one routed to their phone, for example)?

Correct, they can only have one e-mail address.  This change seems correct then until somebody decides to write that feature.

> Of course, if the email field in Foreman is limited to 60 characters, we will probably hit that limit with some long email addresses so how that should be addressed is another question (for another PR).

This was raised recently in Redmine, agreed.  http://projects.theforeman.org/issues/9584

adelton:
> Correct, they can only have one e-mail address. This change seems correct then until somebody decides to write that feature.

Nod. Merge it then.

adelton:
> Nod. Merge it then.

Umm, this might be confusing. So: nod, if this is expected, please merge it.

domcleal:
Merged, thanks @lamawithonel!

Thanks for the review @adelton.

PR #304, opened 2015-03-09
Fix foreman_config_entry checking of dry

domcleal:
The 'dry' method just returns :absent so always appears to be trueish, when it
should have been checking the dry parameter on the user's resource.


ekohl:
:+1:

mmoll:
finally green, thanks! :)

PR #305, opened 2015-03-10
moving away from virtual packages, declaring packages per manifest

smbambling:
moving away from virtual packages, declaring packages per manifest.  Additionally added functionality to lock a compute resource to a specific version


mmoll:
Probably it's enough to rename the class foreman-compute to foreman_compute )as I don't see any class name reusing magic here) to fix the fialing test.

smbambling:
ok renaming and squashing

smbambling:
@mmoll: you were correct looks like test passed now.

manifests/init.pp
```diff
@@ -258,7 +258,6 @@
   # Anchor these separately so as not to break
   # the notify between main classes
   Class['foreman::install'] ~>
-  class { '::foreman::compute': } ~>
```
domcleal: I don't think this block does anything right now, but I think we ought to keep the idea going so that any installation of a foreman-vmware package etc notifies the foreman::service class.

Maybe we could add `tag => 'foreman-compute'` to all of those package resources inside the CR classes and then do:

<pre>
Class['foreman::install'] ~>
Package <| tag == 'foreman-compute' |> ~>
Class['foreman::service']
</pre>

smbambling:
Thats a pretty neat way of doing it.  The other option would just be to add the notify parameter to each of the package manifest.  

package { 'foreman-vmware':
    ensure => $version,
    notify => Service['foreman'],
}

What direction would you like to take this ?

smbambling:
I take that back your way is cleaner in that there is less inheritance to worry about dealing with

domcleal:
Yeah, using init.pp for relationships helps keep it a bit more abstract in case there are more resources in foreman::service that need refreshing etc.

smbambling:
I pushed before squashing and a T-CI build kicked off then I quickly squashed and force pushed...I think that might have caused the failed build.  How would I go about restarting the test, do I need to cancel the PR and resubmit or submit a NOOP commit ?

domcleal:
Hm, it does seem to have tested the wrong commit and the restart button doesn't appear to work.  Can you do a "git commit --amend", and then just force push to trigger it?

smbambling:
I 'think' its running though no progress is listed currently

manifests/init.pp
```diff
@@ -258,7 +258,7 @@
...
+  Package <| tag == 'foreman-compute' |>
```
domcleal: missing a `~>` on the end of this line

ekohl:
I like this.

manifests/compute/foreman_compute.pp
```diff
@@ -0,0 +1,13 @@
+# Provides a common class to install foreman-compute
+# used by multiple compute resources
+#
+# === Parameters:
+#
+# $version::  Package version to install, defaults to installed
+#
+class foreman::compute::foreman_compute ( $version = 'installed' ) {
```
domcleal: Would you mind marking this class as private, so people have to go through the main classes?

Just call private() and increase the stdlib dependency to >= 4.4.0 in metadata.json.

smbambling: That limits the ability to pin a specific version of the foreman-compute module.  If it can be referenced in a manifest you can pin it for rackspace as an example 

````
class { '::foreman::compute::foreman_compute':
  version => '1.7.3',
}

domcleal: Ah yes, fair point, as the other classes don't/can't pass that parameter through.

ekohl:
:+1: when the tests are green.

smbambling:
:100:

domcleal:
Thanks @smbambling!

PR #306, opened 2015-03-11
modulesync: enable future parser tests

domcleal:


mmoll:
rebased.

mmoll:
Switched the discovery tests to rspec-puppet-facts. Not sure about the array vs. string thing,  but I think thats more correct like that. No idea about the IPA fails.

ekohl:
@mmoll I think the ```$::default_ipa_server``` and ```$::default_ipa_realm``` values are not ```''```, but most likely undef.

mmoll:
Alright, let's see what Travis is saying, but that should turn :green_heart:

mmoll:
One 2h timeout, the rest green... as with puppet-puppet, I expect still problem with future_parser as not everything is covered by the tests, but we could just go with it for now and see what's bubbling up out there ;)

manifests/config.pp
```diff
@@ -49,7 +49,7 @@
     class { '::foreman::config::passenger': } -> anchor { 'foreman::config_end': }
 
     if $foreman::ipa_authentication {
-      if $::default_ipa_server == '' or $::default_ipa_realm == '' {
+      if $::default_ipa_server == undef or $::default_ipa_server == '' or $::default_ipa_realm == undef or $::default_ipa_realm == '' {
```
domcleal: feels clunky, but OK.  My only idea is: `if !$::default_ipa_server or empty($::default_ipa_server) ...`

domcleal:
:+1:

manifests/config.pp
ekohl: Your idea is easier to read.

ekohl:
:+1:

manifests/config.pp
domcleal: Amended.

domcleal:
Thanks all!

Issue #308, opened 2015-03-17
mysql class does not exist


treydock:
I'm using puppetlabs/mysql-3.0.0 and just updated to theforeman/foreman-3.0.0 and am receiving this error when I attempt to run Puppet on the Foreman host:

```
Error: Could not retrieve catalog from remote server: Error 400 on SERVER: Could not find class ::mysql for dc-foreman-web.tamu.edu on node dc-foreman-web.tamu.edu
Warning: Not using cache on failed catalog
Error: Could not retrieve catalog; skipping run
```

puppetlabs/mysql 2.3.1 was the last release to include the mysql class - https://github.com/puppetlabs/puppetlabs-mysql/blob/2.3.1/manifests/init.pp. Notice the deprecation warnings.

```
include ::mysql, ::mysql::server, ::mysql::server::account_security
```

Should probably be

```
include ::mysql::server, ::mysql::server::account_security
```

ekohl:
I think you're right and looking at foreman-installer 1.8-stable I think this should also be a problem for RCs. @domcleal I think we should have a bats test that verifies MySQL installation works as well.

treydock:
I'm working on a PR to fix this, with some extra unit tests added since right now the unit tests for this module only cover postgresql.

ekohl:
Thanks for both reporting and fixing!

PR #310, opened 2015-03-17
Do not include mysql class

treydock:
Fix support for mysql by removing inclusion of mysql class which was removed by puppetlabs/mysql in version 3.0.0

Resolves issue #308


domcleal:
:+1: good catch!

ekohl:
Thanks!

PR #311, opened 2015-03-24
Complex values docs for foreman_config_entry type

dLobatog:
`trusted_puppetmaster_hosts` and other arrays need non-obvious strings to work, such as `'[\"xxxx.com\", \"yyyy.net\"]'`

### DIFF
--- a/lib/puppet/type/foreman_config_entry.rb
+++ b/lib/puppet/type/foreman_config_entry.rb
@@ -7,7 +7,7 @@ Puppet::Type.newtype(:foreman_config_entry) do
   end
 
   newproperty(:value) do
-    desc 'The value of the parameter.'
+    desc "The value of the parameter. Hashes and Arrays are expected to be JSON encoded, escaping the quotes. '[\"localhost\"]' would represent an Array containing the string 'localhost'"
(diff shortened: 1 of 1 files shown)


domcleal:
Perhaps better to have it run .to_json when munging?

eLobato:
@domcleal Here? https://github.com/theforeman/puppet-foreman/blob/master/lib/puppet/provider/foreman_config_entry/cli.rb#L77 

(not sure what step is munging..?)

domcleal:
Yeah, it could be there actually, as long as you're able to compare the current value accurately too - i.e. what will the `value` method of the provider return for an array?  Will it equal what the user puts in?  I'd imagine it would output the JSON equivalent again, so you'd have to convert it in both the getter to compare, and the setter.

I was thinking it could be two lines below your patch, which is in the type and handles converting the user input into a standard format that the provider can use.

mmoll:
what's the status here?

mmoll:
ping?

mmoll:
anybody? :)

mmoll:
closing, as nobody reacted.

PR #312, opened 2015-03-31
fixes #9928 - set HOME for foreman-config on all Puppet versions

domcleal:
http://projects.theforeman.org/issues/9928

I haven't figured out why our nightly tests don't hit it, but our 1.8 tests did and I can reproduce it on both a nightly and 1.8 installation on 14.04.  You should see a `foreman-installer -v` tries to re-migrate/seed, or `puppet resource foreman_config_entry --modulepath /usr/share/foreman-installer/modules` simply fails with an error about collect.

The HOME var, like on the < 3.4 code path needed setting for rb-readline as foreman-config loads it and fails quietly otherwise.


ekohl:
:+1:

mmoll:
travis seems to have some funny problems, but as all the other runs are green and the change is looking fine... merged. thanks @domcleal!

PR #313, opened 2015-03-31
fix concat paths in tests

mmoll:
after theforeman/puppet-concat_native#11


ekohl:
:+1:

mmoll:
For some reason the tests with Ruby 1.8.7 and Puppet 3.5 don't finish in two hours time anymore...

domcleal:
Thanks @mmoll!

Yeah, that test is failing almost every time now, we'll have to think about what to do...

PR #314, opened 2015-04-04
fix configure_openscap_repo variable value

cristi1979:
Because the main class doesn't inherit the params class, the variable configure_openscap_repo is not set correctly.


ekohl:
:+1: test failure is unrelated.

mmoll:
thanks @cristi1979 :tada:

PR #315, opened 2015-04-09
use foreman-rake instead of foreman-config

mmoll:
foreman-config is deprecated and also not working for me on Debian/jessie, so change this to foreman-rake, which should be 100% compatible.


domcleal:
Thanks @mmoll!  This is available from 1.7 like the db_migrate changes we made, so it's fine to be merged into 3.0.

PR #316, opened 2015-04-10
use foreman-rake instead of foreman-config (also in 3.0)

mmoll:
(cherry picked from commit da98ce43634af1e4440c726804ba3b5357e97099)

I'd like to have this also in 3.0 and a 3.0.2 tagged and released, so 1.8 packages for Debian/jessie are possible.


domcleal:
Thanks @mmoll!

PR #319, opened 2015-04-30
Write websockets_encrypt as on/off instead of a bool

ekohl:


mmoll:
:+1: no idea about the Puppet 2.7 + Ruby 1.8 test error...

ekohl:
Added 2 commits that should update the tests a bit. Maybe they help.

ekohl:
Test failures on puppet 2.7 are now due to the unless keyword in puppetlabs-apt. I think we should pin that as well.

mmoll:
merged, thanks @ekohl!

PR #320, opened 2015-05-11
Fixes #10466 - nice error on proxy registration error

lzap:
Opinions on warnings instead of errors? Since registration step is not fatal
and it often confuses users I'd vote for this. Otherwise I can just raise an
error with `raise "bla bla"`

### DIFF
--- a/lib/puppet/provider/foreman_smartproxy/rest_v2.rb
+++ b/lib/puppet/provider/foreman_smartproxy/rest_v2.rb
@@ -2,6 +2,11 @@ Puppet::Type.type(:foreman_smartproxy).provide(:rest_v2) do
 
   confine :feature => :apipie_bindings
 
+  def raise_error(e)
+    body = JSON.parse(e.response)["error"]["full_messages"].join(" ") rescue 'N/A'
+    fail "Proxy #{resource[:name]} cannot be registered (#{e.message}): #{body}"
+  end
+
   # when both rest and rest_v2 providers are installed, use this one
   def self.specificity
     super + 1
@@ -30,6 +35,8 @@ Puppet::Type.type(:foreman_smartproxy).provide(:rest_v2) do
     else
       @proxy = api.call(:index, :search => "name=#{resource[:name]}")['results'][0]
     end
+  rescue Exception => e
+    raise_error e
   end
 
   def id
@@ -47,11 +54,15 @@ Puppet::Type.type(:foreman_smartproxy).provide(:rest_v2) do
         :url  => resource[:url]
       }
     })
+  rescue Exception => e
+    raise_error e
   end
 
   def destroy
     api.call(:destroy, :id => id)
     @proxy = nil
+  rescue Exception => e
+    raise_error e
   end
 
   def url
@@ -60,10 +71,14 @@ Puppet::Type.type(:foreman_smartproxy).provide(:rest_v2) do
 
   def url=(value)
     api.call(:update, { :id => id, :smart_proxy => { :url => value } })
+  rescue Exception => e
+    raise_error e
   end
 
   def refresh_features!
     api.call(:refresh, :id => id)
+  rescue Exception => e
+    raise_error e
   end
 
 end


lzap:
To reproduce just send invalid name/url or simply rename existing proxy name to something different and re-run with the same URL.

domcleal:
This should still raise an error, as a resource that doesn't actually fail when it meets an error is really quite useless...

lib/puppet/provider/foreman_smartproxy/rest_v2.rb
```diff
@@ -2,6 +2,12 @@
 
   confine :feature => :apipie_bindings
 
+  def print_warning(e)
+    body = JSON.parse(e.response)["error"]["full_messages"].join(" ") rescue 'N/A'
+    warning("Error while registering proxy (#{e.message}): #{body}, " +
+      "registration can be skipped by unsetting --register-in-foreman option.")
```
domcleal: This message is rather tied to the module.. where the parameter is `register_in_foreman`, or if you meant the foreman-installer option, then it's `--foreman-register-in-foreman`.

lzap:
Yeah sure. Fixed.

lib/puppet/provider/foreman_smartproxy/rest_v2.rb
```diff
@@ -2,6 +2,12 @@
 
   confine :feature => :apipie_bindings
 
+  def raise_error(e)
+    body = JSON.parse(e.response)["error"]["full_messages"].join(" ") rescue 'N/A'
+    raise("Error while registering proxy (#{e.message}): #{body}, " +
```
domcleal: Better to use the builtin fail() in a provider

lib/puppet/provider/foreman_smartproxy/rest_v2.rb
```diff
@@ -2,6 +2,12 @@
...
+      "registration can be skipped by unsetting --register-in-foreman option.")
```
domcleal: see my last comment about the name of the parameter

domcleal: Could we emphasise finding the reason before disabling it too, so people don't follow this advice and get a partially functioning installation?  IIRC we have debug output from apipie-bindings nowadays.

lib/puppet/provider/foreman_smartproxy/rest_v2.rb
lzap: Fixed.

lib/puppet/provider/foreman_smartproxy/rest_v2.rb
lzap: Re: name of parameter - I don't see any related comment on that.

Can you suggest wording you expect? I am really bad in this... I don't follow what debug output from bindings you mean.

lib/puppet/provider/foreman_smartproxy/rest_v2.rb
```diff
@@ -2,6 +2,12 @@
 
   confine :feature => :apipie_bindings
 
+  def raise_error(e)
+    body = JSON.parse(e.response)["error"]["full_messages"].join(" ") rescue 'N/A'
+    fail("Error while registering proxy (#{e.message}): #{body}, " +
+      "registration can be skipped by unsetting --register-in-foreman option.")
```
ekohl: This is only relevant for the installer. Can we come up with something that also works in all cases? I'd prefer if all users registered the proxy so I wonder if we should even encourage users to disable registration.

ekohl:
I think the more detailed error can be very helpful so yay for the idea.

lzap:
Ok I've simplified the error: `fail "Proxy #{resource[:name]} cannot be registered (#{e.message}): #{body}"`

ekohl:
:+1: and I think the test is unrelated (and was already broken before).

domcleal:
Looks better, thanks.  There are a couple of other methods at the bottom which should perhaps be guarded too - url= and refresh_features.

lzap:
Oh thanks for help. We should be all set.

mmoll:
very nice, thanks @lzap!

PR #321, opened 2015-05-19
Use puppetlabs-concat instead of theforeman-concat_native

mcanevet:
This works with either puppetlabs-concat 1.2.x or 2.x so can be merge without waiting for dependencies to be compatible with puppetlabs-concat 2.x IMHO.

### DIFF
--- a/.fixtures.yml
+++ b/.fixtures.yml
@@ -8,7 +8,6 @@ fixtures:
     concat:
       repo:           'git://github.com/puppetlabs/puppetlabs-concat'
       branch:         '1.2.x'
-    concat_native:    'git://github.com/theforeman/puppet-concat'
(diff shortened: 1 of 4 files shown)


ekohl:
Same comments as https://github.com/theforeman/puppet-puppet/pull/263

:+1: code wise, but I think this is a major version bump.

mmoll:
what @ekohl says :)

mcanevet:
@ekohl could you restart Travis job as it exceeded the timeout?

ekohl:
@mcanevet That's a problem we're having for some time on this module. Not sure how to solve it.

mmoll:
works for me :+1:

domcleal:
Nice one, thanks @mcanevet!  Merged as de8e031.

PR #324, opened 2015-05-28
Allow changing email settings for foreman through params

ege1:
Added template, config, and params for managing content of the email.yaml file which configures email settings for theforeman.

### DIFF
--- a/manifests/config.pp
+++ b/manifests/config.pp
@@ -19,6 +19,16 @@ class foreman::config {
     content => template('foreman/database.yml.erb'),
   }
 
+  if $foreman::email_delivery_method and !empty($foreman::email_delivery_method) {
+    file { "/etc/foreman/${foreman::email_conf}":
+      ensure  => file,
+      owner   => 'root',
+      group   => $foreman::group,
+      mode    => '0640',
+      content => template("foreman/${foreman::email_source}"),
+    }
+  }
+
   file { $foreman::init_config:
     ensure  => file,
     content => template("foreman/${foreman::init_config_tmpl}.erb"),
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -2,245 +2,273 @@
 #
 # === Parameters:
 #
-# $foreman_url::              URL on which foreman is going to run
+# $foreman_url::               	URL on which foreman is going to run
 #
-# $unattended::               Should foreman manage host provisioning as well
-#                             type:boolean
+# $unattended::                	Should foreman manage host provisioning as well
+#                              	type:boolean
 #
-# $authentication::           Enable user authentication. Initial credentials are set using admin_username
-#                             and admin_password.
-#                             type:boolean
+# $authentication::            	Enable user authentication. Initial credentials are set using admin_username
+#                              	and admin_password.
+#                              	type:boolean
 #
-# $passenger::                Configure foreman via apache and passenger
-#                             type:boolean
+# $passenger::                 	Configure foreman via apache and passenger
+#                              	type:boolean
 #
-# $passenger_ruby::           Ruby interpreter used to run Foreman under Passenger
+# $passenger_ruby::            	Ruby interpreter used to run Foreman under Passenger
 #
-# $passenger_ruby_package::   Package to install to provide Passenger libraries for the active Ruby
-#                             interpreter
+# $passenger_ruby_package::    	Package to install to provide Passenger libraries for the active Ruby
+#                              	interpreter
 #
-# $use_vhost::                Enclose apache configuration in <VirtualHost>...</VirtualHost>
-#                             type:boolean
+# $use_vhost::                 	Enclose apache configuration in <VirtualHost>...</VirtualHost>
+#                              	type:boolean
 #
-# $servername::               Server name of the VirtualHost in the webserver
+# $servername::                	Server name of the VirtualHost in the webserver
 #
-# $ssl::                      Enable and set require_ssl in Foreman settings (note: requires passenger, SSL does not apply to kickstarts)
-#                             type:boolean
+# $ssl::                       	Enable and set require_ssl in Foreman settings (note: requires passenger, SSL does not apply to kickstarts)
+#                              	type:boolean
 #
-# $custom_repo::              No need to change anything here by default
-#                             if set to true, no repo will be added by this module, letting you to
-#                             set it to some custom location.
-#                             type:boolean
+# $custom_repo::               	No need to change anything here by default
+#                              	if set to true, no repo will be added by this module, letting you to
+#                              	set it to some custom location.
+#                              	type:boolean
 #
-# $repo::                     This can be stable, nightly or a specific version i.e. 1.7
+# $repo::                      	This can be stable, nightly or a specific version i.e. 1.7
 #
-# $configure_epel_repo::      If disabled the EPEL repo will not be configured on RedHat family systems.
-#                             type:boolean
+# $configure_epel_repo::       	If disabled the EPEL repo will not be configured on RedHat family systems.
+#                              	type:boolean
 #
-# $configure_scl_repo::       If disabled the SCL repo will not be configured on Red Hat clone systems.
-#                             (Currently only installs repos for CentOS and Scientific)
-#                             type:boolean
+# $configure_scl_repo::        	If disabled the SCL repo will not be configured on Red Hat clone systems.
+#                              	(Currently only installs repos for CentOS and Scientific)
+#                              	type:boolean
 #
-# $configure_brightbox_repo:: Configure the Brightbox PPA for Ubuntu, providing updated Ruby and
-#                             Passenger packages
-#                             type:boolean
+# $configure_brightbox_repo::  	Configure the Brightbox PPA for Ubuntu, providing updated Ruby and
+#                              	Passenger packages
+#                              	type:boolean
 #
-# $selinux::                  when undef, foreman-selinux will be installed if SELinux is enabled
-#                             setting to false/true will override this check (e.g. set to false on 1.1)
-#                             type:boolean
+# $selinux::                   	when undef, foreman-selinux will be installed if SELinux is enabled
+#                              	setting to false/true will override this check (e.g. set to false on 1.1)
+#                              	type:boolean
 #
-# $gpgcheck::                 turn on/off gpg check in repo files (effective only on RedHat family systems)
-#                             type:boolean
+# $gpgcheck::                  	turn on/off gpg check in repo files (effective only on RedHat family systems)
+#                              	type:boolean
 #
-# $version::                  foreman package version, it's passed to ensure parameter of package resource
-#                             can be set to specific version number, 'latest', 'present' etc.
+# $version::                   	foreman package version, it's passed to ensure parameter of package resource
+#                              	can be set to specific version number, 'latest', 'present' etc.
 #
-# $db_manage::                if enabled, will install and configure the database server on this host
-#                             type:boolean
+# $db_manage::                 	if enabled, will install and configure the database server on this host
+#                              	type:boolean
 #
-# $db_type::                  Database 'production' type (valid types: mysql/postgresql/sqlite)
+# $db_type::                   	Database 'production' type (valid types: mysql/postgresql/sqlite)
 #
-# $db_adapter::               Database 'production' adapter
+# $db_adapter::                	Database 'production' adapter
 #
-# $db_host::                  Database 'production' host
+# $db_host::                   	Database 'production' host
 #
-# $db_port::                  Database 'production' port
-#                             type:integer
+# $db_port::                   	Database 'production' port
+#                              	type:integer
 #
-# $db_database::              Database 'production' database (e.g. foreman)
+# $db_database::               	Database 'production' database (e.g. foreman)
 #
-# $db_username::              Database 'production' user (e.g. foreman)
+# $db_username::               	Database 'production' user (e.g. foreman)
 #
-# $db_password::              Database 'production' password (default is random)
+# $db_password::               	Database 'production' password (default is random)
 #
-# $db_sslmode::               Database 'production' ssl mode
+# $db_sslmode::                	Database 'production' ssl mode
 #
-# $db_pool::                  Database 'production' size of connection pool
-#                             type:integer
+# $db_pool::                   	Database 'production' size of connection pool
+#                              	type:integer
 #
-# $apipie_task::              Rake task to generate API documentation.
-#                             Use 'apipie:cache' on 1.7 or older, 'apipie:cache:index' on 1.8 or newer.
+# $apipie_task::               	Rake task to generate API documentation.
+#                              	Use 'apipie:cache' on 1.7 or older, 'apipie:cache:index' on 1.8 or newer.
 #
-# $app_root::                 Name of foreman root directory
+# $app_root::                  	Name of foreman root directory
 #
-# $manage_user::              Controls whether foreman module will manage the user on the system. (default true)
-#                             type:boolean
+# $manage_user::               	Controls whether foreman module will manage the user on the system. (default true)
+#                              	type:boolean
 #
-# $user::                     User under which foreman will run
+# $user::                      	User under which foreman will run
 #
-# $group::                    Primary group for the Foreman user
+# $group::                     	Primary group for the Foreman user
 #
-# $user_groups::              Additional groups for the Foreman user
-#                             type:array
+# $user_groups::               	Additional groups for the Foreman user
+#                              	type:array
 #
-# $environment::              Rails environment of foreman
+# $environment::               	Rails environment of foreman
 #
-# $puppet_home::              Puppet home directory
+# $puppet_home::               	Puppet home directory
 #
-# $locations_enabled::        Enable locations?
-#                             type:boolean
+# $locations_enabled::         	Enable locations?
+#                              	type:boolean
 #
-# $organizations_enabled::    Enable organizations?
-#                             type:boolean
+# $organizations_enabled::     	Enable organizations?
+#                              	type:boolean
 #
-# $passenger_interface::      Defines which network interface passenger should listen on, undef means all interfaces
+# $passenger_interface::       	Defines which network interface passenger should listen on, undef means all interfaces
 #
-# $passenger_prestart::       Pre-start the first passenger worker instance process during httpd start.
-#                             type:boolean
+# $passenger_prestart::        	Pre-start the first passenger worker instance process during httpd start.
+#                              	type:boolean
 #
-# $passenger_min_instances::  Minimum passenger worker instances to keep when application is idle.
+# $passenger_min_instances::   	Minimum passenger worker instances to keep when application is idle.
 #
-# $passenger_start_timeout::  Amount of seconds to wait for Ruby application boot.
+# $passenger_start_timeout::   	Amount of seconds to wait for Ruby application boot.
 #
-# $server_ssl_ca::            Defines Apache mod_ssl SSLCACertificateFile setting in Foreman vhost conf file.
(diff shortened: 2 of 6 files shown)


manifests/config.pp
```diff
@@ -19,6 +19,13 @@
     content => template('foreman/database.yml.erb'),
   }
 
+  file { "/etc/foreman/$foreman::email_conf":
```
ekohl: You should use braces around variables: ```"/etc/foreman/${foreman::email_conf}"

manifests/config.pp
```diff
@@ -19,6 +19,13 @@
...
+    owner   => 'root',
+    group   => $foreman::group,
+    mode    => '0640',
+    content => template("foreman/$foreman::email_source"),
```
ekohl: Also needs braces

ekohl:
Since this makes it explicit, I wonder if this will overwrite any existing user settings for upgrading users.

ege1:
Fixed the missing braces.

You are right, it will overwrite an already existing email.yaml file and overwrite it with the default setting which are there after installing theforeman...

I don't know how to avoid that.

ege1:
Ah, maybe it is better to only touch the email.yaml file if one of this config parameters got touched. I could add an 
if ($foreman::email_smtp_address != 'smtp.example.com') {
  config changes
}

Because this setting will surely be changed from default if someone is gonna use the email functionality....

ekohl:
It's common to ship with empty defaults then.

ege1:
OK, changed the defaults and only kept values that make any sense. Now the file will only be created/replaced if at least the delivery method is set. Also enhanced the template for all possible options found here: http://projects.theforeman.org/projects/foreman/wiki/Email_configuration

ege1:
Is it fine now?

manifests/init.pp
```diff
@@ -168,72 +168,91 @@
 # $websockets_ssl_key::       SSL key file to use when encrypting websocket connections
 # $websockets_ssl_cert::      SSL certificate file to use when encrypting websocket connections
 #
+# $email_conf::               email configuration file, defaults to /etc/foreman/email.yaml
+# $email_source::             template to use for email configuration file
+# $email_delivery_method::    can be :sendmail or :smtp regarding to foreman docu
+# $email_smtp_address::       if delivery_method is smtp, this should contain an valid smtp host
+# $email_smtp_port::          smtp port, defaults to 25
+# $email_smtp_domain::        email domain
+# $email_smtp_authentication::authentication settings, can be :none or :login, defaults to :none
```
ekohl: I don't know if kafo can parse this or if it needs whitespace in between.

templates/email.yaml.erb
```diff
@@ -0,0 +1,15 @@
+# Outgoing email settings
+
+production:
+  delivery_method: <%= @email_delivery_method %>
```
ekohl: Not sure what erb does with undef here (and foreman if the value is empty/invalid). Maybe move the if above ```production:```?

ege1:
OK, added the whitespace.

The template only gets used
  if $foreman::email_delivery_method 
is set. So this shouldn't be an issue.

If you like I can add a condition in the erb as well, but as it doesn't get called, I hope that shouldn't be necessary.

mmoll:
Parameters and documentation headers seem to differ. BTW, testing directly with foreman-installer (which is using kafo) is easy, at least when running nightly packages (so the diffs against master apply without hassle):
```
mmoll@sledge:/tmp$ wget https://patch-diff.githubusercontent.com/raw/theforeman/puppet-foreman/pull/324.diff
[...]
mmoll@sledge:/tmp$ cd /usr/share/foreman-installer/modules/foreman
mmoll@sledge:/usr/share/foreman-installer/modules/foreman$ sudo patch -p1 </tmp/324.diff
patching file manifests/config.pp
patching file manifests/init.pp
patching file manifests/params.pp
patching file templates/email.yaml.erb
mmoll@sledge:/usr/share/foreman-installer/modules/foreman$ sudo foreman-installer
[FATAL 2015-05-31 00:19:17 fatal] Unable to continue because of: undocumented parameters in foreman: email_smtp_password, email_smtp_user_name
```

ege1:
Hi mmoll,

Thanks for testing and sorry, I wasn't careful enough. 

Right now I don't have a machine to run such tests on. I'm just about to set-up foreman and a KVM host so I can test such changes in a VM.

Where do I get nightly from foreman-installer?

mmoll:
No worries, that's what code reviews are for - maybe we should think about how to test this automatically per PR... @ares is there a quick way with kafo(_parsers) standalone to check if the module parameters and the doc headers match?

When you want to run the nightly packages (which includes all components built from their master/develop versions), just use the Quickstart Guide from the website, but use ```nightly``` instead of ```1.8``` when it comes to the version of the repositories.

ege1:
Ahhh, so it is also directly in the repositories. I thought to find a download link or the like for the installer itself.

mmoll:
Now -installer is OK with it, however, there's still a variable mismatch between ```init.pp``` and ```params.pp```:
```$email_user_name``` vs. ```$email_smtp_user_name```
```$email_password``` vs. ```$email_smtp_password```

**Comment by ekohl** [comment]
@mmoll I recall @domcleal did work on a puppet-lint check for this, but not sure if it ever got close to production ready.

**Comment by ege1** [comment]
Good Morning,

Thanks for having a good eye on that mmoll. I changed the variables now so they all match. All smtp relevant variables should now be email_smtp_

File: manifests/config.pp [hunk 5]
```diff
@@ -19,6 +19,15 @@
     content => template('foreman/database.yml.erb'),
   }
 
+  if $foreman::email_delivery_method {
```
ekohl: One problem may be that if a user changes the method to be unmanaged, you have a left over file. Also puppet 4 is a bit different about empty strings so maybe use ```if $foreman::email_delivery_method && !empty($foreman::email_delivery_method)```?

ege1:
OK, the puppet 3.8 docu states:
Empty strings are false; all other strings are true.

is !empty($foreman::email_delivery_method) valid puppet syntax? If so I can change that if you think it is better. Or should I use if $foreman::email_delivery_method != ""

That should also cover both. If not set and if empty, doesn't it?

ekohl:
https://docs.puppetlabs.com/puppet/4.1/reference/experiments_future.html#check-your-comparisons states:
> The rules for converting non-boolean values to boolean have changed. In Puppet 4, undef converts to false and all other non-boolean values convert to true; in Puppet 3.x, empty strings were also false.

So if you use ```empty()``` or ```!= ""``` doesn't matter to me. Maybe the latter is easier to read. It is important to check for both undef and empty strings.

ege1:
OK, I thought != "" would also check if undef. So if it is undef, the check != "" would be true?

Changed it like you suggested. I wasn't aware that !empty(variable) is valid puppet syntax. What else from ruby syntax is valid like that? Or is this stdlib syntax?

ekohl:
> OK, I thought != "" would also check if undef. So if it is undef, the check != "" would be true?

Yes, that would be true.

> Changed it like you suggested. I wasn't aware that !empty(variable) is valid puppet syntax. What else from ruby syntax is valid like that? Or is this stdlib syntax?

I think ```empty()``` is a regular puppet function call (that happens to come from stdlib) and ```!``` just negates it.

templates/email.yaml.erb
mmoll: Moving this also into the if block sounds like the right thing to do.

manifests/config.pp
```diff
@@ -19,6 +19,15 @@
     content => template('foreman/database.yml.erb'),
   }
 
+  if $foreman::email_delivery_method && !empty($foreman::email_delivery_method) {
```
mmoll: should probably be ```and``` instead of ```&&```

ege1:
Haha, I blindly followed your suggestion :)

ekohl:
I shouldn't combine programming languages :)

mmoll:
Using the according plugins for vim/emacs or a IDE with Puppet support (like RubyMine or Gepetto) helps a lot, I highly recommend that. :)

ege1:
I use puppet-syntax for vim, but I can't see a difference in syntax highlighting if I use "&&" or "and" though :(

mmoll:
I use pathogen/syntastic in vim to get everything checked with puppet-lint (and probably also parsed with puppet paster validate), that's a combo I can can recommend. If you need help, setting it up, drop me a mail or contact me via IRC.

mmoll:
Finally tested this more extensively, and it already is usable for me in foreman-installer, but I have some notes:

This needs a ```validate_re``` on ```email_delivery_method``` so that only ```sendmail``` and ```smtp``` can be given as parameters, also I'd recommend to add the colon before it in the template itself. At least to me it's counterintuitive to have it in the string of the class parameter.

The comment from @ekohl on the template also needs to get adressed, and maybe it would make even sense to manage the ```email.yaml``` file in a way it's getting set to ```absent``` when ```email_delivery_method``` is not set.

Also, spec tests would be nice, but we can take care of that, if this is getting too much now. :panda_face:

ege1:
Can you help me with that? Where do I put this validation? init, params or config? Do you know if there are more options than sendmail and smtp?

Regarding the colon, that's what I had before, but then I was afraid someone would add the colon because it is in the documents. So I will change that again, that is more intuitive.

My problem with the email.yaml file is, that it already gets delivered through the packages (rmp) at least at CentOS7. So I would remove an previously installed file, or even worse a file the user managed on his own before. As you for sure do have more experience in that, I would follow your suggestion, though. 

Which comment by @ekohl are your referring to? The one if the possibly empty variable? This should be caught by the config already

If you tell me how to do these spec tests, I would love to do them also...

domcleal:
> Where do I put this validation? init, params or config?

Usually in init.pp, towards the top.  When validating parameters that default to undef, you'll want to wrap them in a conditional, e.g. `if $normally_undef { validate_string($normally_undef) }`.

> Do you know if there are more options than sendmail and smtp?

That's pretty much it.  `test` and `file` are valid, but not really useful here.

I agree about removing the colon from the parameter value.

> My problem with the email.yaml file is, that it already gets delivered through the packages (rmp) at least at CentOS7. So I would remove an previously installed file, or even worse a file the user managed on his own before.

Yeah, tricky.  We could deploy a commented file rather than deleting it, so at least it's a little clearer what's happening - and also the package manager won't then just put it back again on upgrade.  Puppet also keeps a backup of files it overwrites.

> If you tell me how to do these spec tests, I would love to do them also...

That'd be great, thanks!  Start with spec/classes/foreman_config_spec.rb which covers this class.  If you're not familiar with these types of tests, perhaps have a quick read of http://rspec-puppet.com - they test the resulting Puppet catalog.

There are two sections (one RH, one Debian) which test the default configurations.  I'd add something like this to check that email.yaml isn't deployed by default:

    it { should_not contain_file('/etc/foreman/email.yaml') }

(Or if you're making it absent, it `should contain_file('...').with_ensure('absent')`.)

Then you can make a new `describe` section to describe when say, SMTP is configured.

<pre>
describe 'with email configured for SMTP' do
  let :pre_condition do
    "class {'foreman':
       email_delivery_method => 'smtp',
     }"
  end

  it 'should contain email.yaml' do
    should contain_file('/etc/foreman/email.yaml').
      with_content('delivery_method: :smtp$').
      with_content('something: else$').
      with_ensure('present')
  end
end
</pre>

mmoll:
> Which comment by @ekohl are your referring to? The one if the possibly empty variable?

There was also one in templates/email.yaml.erb about moving the if-clause up (above the ```production:``` line. This is also a bit related to the question what to do when the mail settings are "unmanaged".

ege1:
OK, that should be handled by the config.pp already. The template only gets used if the variable is set. Would you like to have an additional test in the template also?

ege1:
> > My problem with the email.yaml file is, that it already gets delivered
> > through the packages (rmp) at least at CentOS7. So I would remove an
> > previously installed file, or even worse a file the user managed on his
> > own before.
> Yeah, tricky.  We could deploy a commented file rather than deleting it, so
> at least it's a little clearer what's happening - and also the package
> manager won't then just put it back again on upgrade.  Puppet also keeps a
> backup of files it overwrites.

The original File which get's delivered through the package isn't even 
commented and has example.com defaults. So we can't do it worse with these 
files. My concern is if a user edited it. As long as he doesn't touch the 
parameters for this option, we shouldn't touch an existing file, should we? 

> > If you tell me how to do these spec tests, I would love to do them also...
> 
> That'd be great, thanks!  Start with spec/classes/foreman_config_spec.rb
> which covers this class.  If you're not familiar with these types of tests,
> perhaps have a quick read of http://rspec-puppet.com - they test the
> resulting Puppet catalog.

What is this spec for? Is this used by the foreman-installer or does it have 
to do with puppet itself?

I will have a closer look into that as soon as I have some spare time. I 
didn't get it completely right now.

ege1:
Hi @domcleal,

I added some spec clauses for the email.yaml settings. Can you have a look at them?

If I got it right these tests are run on the compiled catalog. Does it run automatically on every puppet run or do you have to run specific tests for that?

How can I verify my tests are valid syntax and everything works as expected?

ekohl:
The tests are run when a developer manually runs them or Travis CI does it for us in pull requests. It does not happen on every catalog compilation because that would be too heavy.

For running and writing tests, see https://github.com/theforeman/puppet-foreman/blob/master/CONTRIBUTING.md#testing

Also, it appears there's a merge conflict somewhere. Please rebase your patches so it can be merged.

manifests/init.pp
```diff
@@ -248,6 +267,9 @@
     fail("${::hostname}: External authentication via IPA can only be enabled when passenger is used.")
   }
   validate_bool($websockets_encrypt)
+  if $email_delivery_method {
+    validate_re($email_delivery_method, ['^sendmail$', '^smtp$'], "email_delivery_method can be either sendmail or smtp, not $email_delivery_method")
```
ekohl: The variable must be enclosed in {} in a string.

ege1:
The rebase is done through git rebase master? Or is there something more specific?

mmoll:
Exactly, but you first have to update your master branch to the currect theforeman master branch to work, check the GitHub help on how to synchronize your fork with this upstream repo.

ege1:
OK, the synchronization worked, but the rebase is showing me a lot of conflicts that I've to solve. Can you give me a short hint how to solve them? I'm not use to using git though. For me the patch that can't be merged looks quit good. 

What do I have to do to resolve the conflicts?

mmoll:
That's a bit out of scope here, but there should be plenty of documentation on the interwebs. Personally I use git with meld (http://meldmerge.org/) to resolve such conflicts. Having conflicts that must be resolved manually just means git is unable to auto-merge because it does not know which side to choose by itself.

mmoll:
tests are failing

ege1:
Ah, that's the specs I added. Could you help me with these? I mean the file should be set to "present" and I thought it tests the content as "includes"...

manifests/init.pp
```diff
@@ -2,238 +2,263 @@
...
+# $websockets_ssl_key::        	SSL key file to use when encrypting websocket connections
+# $websockets_ssl_cert::       	SSL certificate file to use when encrypting websocket connections
+#
+# $logging_level::              Logging level of the Foreman application (valid values: debug, info, warn, error, fatal)
+#
+# $loggers::                    Enable or disable specific loggers, e.g. {"sql" => true}
+#                               type:hash
+# $email_conf::                	email configuration file, defaults to /etc/foreman/email.yaml
+# $email_source::              	template to use for email configuration file
+# $email_delivery_method::     	can be sendmail or smtp regarding to foreman docu
```
elyezer: s/docu/documentation?

mmoll:
It seems the rebase got somehow wrong, I see commits twice and also commits from Dominic, however, I extracted the diff between the branches, applied mmoll/puppet-foreman@cc88af8c to fancy my OCD and getting green tests now with mmoll/puppet-foreman@a74b3344, mainly the problems where about setting the ```end```s of the ```describe``` blocks and the default of ```with_content``` to math the exact content of the whole file.

ege1:
Thanks @mmoll for your effort. So the end blocks were wrong? The with_content was a problem I saw and tried to fix, but with no luck.

mmoll:
The first describe block spanned a bit too much to the bottom. Anyway, feel free to take my commits into your branch. As I wasn't able to get the rebase going to squash your commits together, it might make sense for you to open up another branch and import the diff (you can use mmoll/puppet-foreman@35f3a78a if that diff looks correct to you.

On another note, it's not necessary to have the tests in the Debian section again, as there is no OS/distro specfic logic applied regarding that template. Also, I'm a bit unsure if I would like to have ```$email_conf``` and ```$email_source``` as variables, but leave that open for discussion.

mmoll:
@ege1 as Foreman 1.9 is coming closer, I'd like to get this in, could you tidy this PR up or open another one with a fresh commit?

mmoll:
You'll need to use ```scope::lookupvar``` in the template to pass the tests with strict variables, like that:
```diff
diff --git a/templates/email.yaml.erb b/templates/email.yaml.erb
index ea93c2f..484c874 100644
--- a/templates/email.yaml.erb
+++ b/templates/email.yaml.erb
@@ -1,15 +1,15 @@
 # Outgoing email settings
 
 production:
-  delivery_method: :<%= @email_delivery_method %>
-<%- if @email_delivery_method == 'smtp' -%>
+  delivery_method: :<%= scope.lookupvar("::foreman::email_delivery_method") %>
+<%- if scope.lookupvar("::foreman::email_delivery_method") == 'smtp' -%>
   smtp_settings:
-    address: <%= @email_smtp_address %>
-    port: <%= @email_smtp_port %>
-    domain: <%= @email_smtp_domain %>
-    authentication: :<%= @email_smtp_authentication %>
-  <%- if @email_smtp_authentication == 'login' -%>
-    user_name: <%= @email_smtp_user_name %>
-    password: <%= @email_smtp_password %>
+    address: <%= scope.lookupvar("::foreman::email_smtp_address") %>
+    port: <%= scope.lookupvar("::foreman::email_smtp_port") %>
+    domain: <%= scope.lookupvar("::foreman::email_smtp_domain") %>
+    authentication: :<%= scope.lookupvar("::foreman::email_smtp_authentication") %>
+  <%- if scope.lookupvar("::foreman::email_smtp_authentication") == 'login' -%>
+    user_name: <%= scope.lookupvar("::foreman::email_smtp_user_name") %>
+    password: <%= scope.lookupvar("::foreman::email_smtp_password") %>
   <%- end -%>
 <%- end -%>
```
otherwise it's looking good, IMHO

ege1:
OK, thanks. That's new to me. I thought the scope is exactly limited to the module I'm in when using @veriable_name

mmoll:
@ege1 future parser is changing the game here.
@domcleal @ekohl mind to have a look?

ekohl:
@mmoll I think that's only the max execution time in travis that we're passing.

mmoll:
@ekohl tests were failing with future_parser because the variables didn't arrive in the template.

domcleal:
@mmoll are you looking at https://travis-ci.org/theforeman/puppet-foreman/builds/69238130 or another?

mmoll:
https://travis-ci.org/theforeman/puppet-foreman/builds/68914845

domcleal:
@mmoll that's the old commit with `@` variable access, the new commit with scope.lookupvar seems fine.

mmoll:
yes, that's what I was talking about, it's fine now :)

domcleal:
Oh, sorry, I had the wrong end of the stick!  You meant a general look, rather than a look at some specific failure.  I'll do so in a bit.

mmoll:
ah, yes... Please review, as I think it's good to go now. :)

domcleal:
Merged, thanks very much @ege1!

ege1:
@domcleal, @mmoll, thanks for your support and patience ;-)

PR #326, opened 2015-05-31
refactor yumrepo data to a different class

shlomizadok:


manifests/plugin/openscap.pp
```diff
@@ -13,32 +13,7 @@
...
-          gpgcheck => 0,
-          baseurl  => "http://copr-be.cloud.fedoraproject.org/results/isimluk/OpenSCAP/${repo}-${major}-\$basearch/",
-          before   => [ Foreman::Plugin['openscap'] ],
-        }
-      }
-      default: {
-        fail("Unsupported osfamily ${::osfamily}")
-      }
-    }
+    contain foreman::plugin::openscap::repo
```
ekohl: IIRC contain is puppet 3.4+, but we still support puppet 2.7. You can use a class relation though.

ekohl:
You can force push to a PR with an amended commit, or just rewrite ti and squash it after you're done.

domcleal:
I recommended not doing this on the linked foreman-infra PR, as @shlomizadok was trying to use this one class in isolation, which is an abuse of this module I think.

PR #328, opened 2015-06-05
fixes #5838 - add logging_level, loggers parameters

domcleal:
1. I took a shortcut with the `loggers` parameter in making it a one-level deep hash so you can only specify whether it's enabled or not, no other fancy options (overriding what's in logging.yaml).  This makes it usable from kafo (e.g. `foreman-installer --foreman-logging-level=debug --foreman-loggers=ldap:true`)
2. I didn't think templating logging.yaml was necessary, the main switches are in settings.yaml (see http://projects.theforeman.org/projects/foreman/wiki/Troubleshooting#How-do-I-enable-debugging).
3. The behaviour of `loggers` depends on ~~https://github.com/theforeman/foreman/pull/2437~~ which makes Foreman merge settings into sane defaults rather than us having to list all loggers here.  Best to wait for that PR before merging this.

### DIFF
--- a/templates/settings.yaml.erb
+++ b/templates/settings.yaml.erb
@@ -23,3 +23,15 @@
 :websockets_encrypt: <%= scope.lookupvar("foreman::websockets_encrypt") ? 'on' : 'off' %>
 :websockets_ssl_key: <%= scope.lookupvar("foreman::websockets_ssl_key") %>
 :websockets_ssl_cert: <%= scope.lookupvar("foreman::websockets_ssl_cert") %>
+
+# Log settings for the current environment can be adjusted by adding them
+# here. For example, if you want to increase the log level.
+:logging:
+  :level: <%= scope.lookupvar("foreman::logging_level") %>
+
+# Individual logging types can be toggled on/off here
+:loggers:
+<% scope.lookupvar("foreman::loggers").each do |logger,enabled| -%>
+  :<%= logger %>:
+    :enabled: <%= enabled %>
+<% end -%>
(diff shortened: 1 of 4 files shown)


templates/settings.yaml.erb
```diff
@@ -3,11 +3,7 @@
 ---
 <%= @header %>
 
-#your default puppet server - can be overridden in the host level
```
ekohl: Maybe a nitpick, but maybe this should be a separate commit? I'm in favour of removing it BTW.

domcleal: Good point, fixed.

domcleal:
I was also thinking about adding some functionality to foreman::plugin to configure loggers inside the plugin settings file (see [plugin logging](http://projects.theforeman.org/projects/foreman/wiki/How_to_Create_a_Plugin#Logging)), e.g. by changing the config file to be a concat and then if $loggers is passed in, we add a new fragment.

We could then add a $loggers parameter to any foreman::plugin::* class that we know supports other loggers - how does that sound?

domcleal:
Hrm, scrap that last idea.  Since the file contents are provided entirely externally in foreman::plugin::config we can't easily just add a loggers section to it since it's in YAML, without assuming what the content of $config looks like.

mmoll:
:+1:

domcleal:
> The behaviour of loggers depends on theforeman/foreman#2437

Merged.

mmoll:
merged, thanks @domcleal!

PR #329, opened 2015-06-11
Expose client_ssl_* class params

kmcfate:
You can specify a different server certificate to use for the passenger server, however if it uses a different CA (in the case of using a third party cert) you cannot change the CA cert the client uses to verify it against. This exposes the client_ssl params as class params.


ekohl:
I don't think this will work since ```foreman::puppetmaster``` directly inherits from ```foreman::params```.

ekohl:
Since ```foreman::puppetmaster``` is usually included through ```puppet::server::config``` (which sets ssl_* explicitly, I think you may want to have a look at that. You can override the ```puppet::server_foreman_ssl_*``` parameters specifically for that purpose.

ekohl:
Since ```foreman::puppetmaster``` is usually included through ```puppet::server::config``` (which sets ssl_* explicitly, I think you may want to have a look at that. You can override the ```puppet::server_foreman_ssl_*``` parameters specifically for that purpose.

PR #332, opened 2015-06-28
Add puppet 4 to the travis matrix

ekohl:


spec/spec_helper.rb
```diff
@@ -2,7 +2,6 @@
 #   https://github.com/theforeman/foreman-installer-modulesync
 
 require 'puppetlabs_spec_helper/module_spec_helper'
-require 'lib/module_spec_helper'
```
domcleal: This doesn't look right.. causing broken tests?

ekohl: I was looking at our modules and I wonder if we should include that helper in modulesync.

domcleal: Yeah, I was thinking the same!

ekohl:
It now fails in the report processor, but I don't know if that's test related or just a general incompatibility with puppet 4.

domcleal:
General incompatibility, but easily fixed it seems.  It looks like it unlocks a whole bunch of other failures, which we should see from Travis soon.

ekohl:
Much of the failures are now due to a missing ```$::rubyversion``` fact in our tests.

ekohl:
Rebased, now with more rspec-puppet-facts usage. It now fails on the foreman_report_processor_spec but haven't had time to look into it.

domcleal:
> It now fails on the foreman_report_processor_spec but haven't had time to look into it.

I think our deserialisation of old reports into Puppet 4 is failing, probably due to internal class changes.  We might have to recreate serialised reports showing the different error conditions and separate them by version (2/3 vs 4).

There's a failure remaining in foreman_config_spec where apache::vhost's failing with a mix of non-IP and IP vhosts.  This used to work before r-p-f because $::interfaces was set to '' and https://github.com/theforeman/puppet-foreman/blob/4.0.0/manifests/config/passenger.pp#L89-L90 would ignore it.  Not quite sure how this is meant to work, tbh.

ekohl:
> It now fails on the foreman_report_processor_spec but haven't had time to look into it.
> 
> I think our deserialisation of old reports into Puppet 4 is failing, probably due to internal class changes. We might have to recreate serialised reports showing the different error conditions and separate them by version (2/3 vs 4).

Makes sense. A shortcut would be to throw an error if we detect Puppet 4 to say it's not yet supported. Benefit would be that we test most of the code with Puppet 4 already.

> There's a failure remaining in foreman_config_spec where apache::vhost's failing with a mix of non-IP and IP vhosts. This used to work before r-p-f because $::interfaces was set to '' and https://github.com/theforeman/puppet-foreman/blob/4.0.0/manifests/config/passenger.pp#L89-L90 would ignore it. Not quite sure how this is meant to work, tbh.

I noticed that too and will have a look at it. I think it's time to revisit https://github.com/theforeman/puppet-foreman/pull/255

mmoll:
As this already needs a serious rebase (and I failed to get the tests right), I propose to...
* open a new PR to get the compatible parts of this PR into master
* work on #255 to get it into master
* continue here

ekohl:
> As this already needs a serious rebase (and I failed to get the tests right), I propose to...
> 
>     open a new PR to get the compatible parts of this PR into master

I'll do some cherry picking.

>     work on #255 to get it into master

That may not be needed, but probably not a bad thing in general.

mmoll:
what's the status here?

mmoll:
rebased, using #368 and the diff I posted there.

mmoll:
added `foreman_url` also in spec/defines/foreman_config_passenger_fragment_spec.rb, now tests should become green

domcleal:
Looks fine to me I think.  My only question is - what happened to the earlier test failures (report serialisation and vhost related)?  Something changed externally to fix it?

mmoll:
I also asked that myself, the only thing I can imagine is that e.g. the interfaces/ip facts are now populated correctly via updated rspec-puppet-facts, but unsure if that's the real explanation.

domcleal:
Rebased it on top of the rspec-puppet-facts merge to master.

.sync.yml
```diff
@@ -16,6 +16,7 @@ Rakefile:
     - manifests/plugin/*.pp
 spec/spec_helper.rb:
   requires:
+    - lib/module_spec_helper
```
ekohl: I think this can be dropped by now since we merged those into ```spec_helper.rb```.

domcleal: Thanks, removed that commit.

mmoll:
re-added 465f736

ekohl:
Amended https://github.com/theforeman/puppet-foreman/commit/58a2802af4d6fd3dcc6da4aeeea0f0dfbcc1768f with spec tests for the passenger fragment defines.

mmoll:
this is looking good :+1:

domcleal:
Thanks @ekohl and @mmoll, merged!

PR #333, opened 2015-07-09
The --foreman-configure-ipa-repo option is no longer required

veruu:
The --foreman-configure-ipa-repo option is no longer required since the packages are included in RHEL/CentOS versions 6 and 7 now.

This feature was needed in older RHEL/CentOS versions, in which some packages weren't available. Now there is no need for it. We wanted to remove it from the documentation first and then from the code, after the comment from theforeman.org and discussion with @ares we decided to change it to deprecated and notify the user if he wants to use it.

### DIFF
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -153,8 +153,8 @@
 #
 # $pam_service::               	PAM service used for host-based access control in IPA
 #
-# $configure_ipa_repo::        	Enable custom yum repo with packages needed for external authentication via IPA,
-#                              	this may be needed on RHEL 6.5 and older.
+# $configure_ipa_repo::        	DEPRECATED: Enable custom yum repo with packages needed for external authentication via IPA,
+#                              	this was needed on RHEL 6.5 and older, no longer required
 #                              	type:boolean
 #
 # $ipa_manage_sssd::           	If ipa_authentication is true, should the installer manage SSSD? You can disable it
--- a/manifests/install.pp
+++ b/manifests/install.pp
@@ -44,6 +44,7 @@ class foreman::install {
         ensure  => installed,
         require => $repo,
       }
+      notice ('Using configure_ipa_repo is deprecated and no longer required.')
     }
   }
 
--- a/manifests/install/repos/extra.pp
+++ b/manifests/install/repos/extra.pp
@@ -36,6 +36,7 @@ class foreman::install::repos::extra(
       baseurl  => "http://copr-be.cloud.fedoraproject.org/results/adelton/identity_demo/epel-${osreleasemajor}-\$basearch/",
       before   => Package['mod_authnz_pam', 'mod_lookup_identity', 'mod_intercept_form_submit', 'sssd-dbus'],
     }
+    notice ('Using configure_ipa_repo is deprecated and no longer needed. If you have already used it, disable and delete the repo.')
   }
 
   if $configure_brightbox_repo {


manifests/init.pp
```diff
@@ -153,8 +153,8 @@
 #
 # $pam_service::               	PAM service used for host-based access control in IPA
 #
-# $configure_ipa_repo::        	Enable custom yum repo with packages needed for external authentication via IPA,
-#                              	this may be needed on RHEL 6.5 and older.
+# $configure_ipa_repo::        	DEPRECATED: Enable custom yum repo with packages needed for external authentication via IPA,
+#                               this was needed on RHEL 6.5 and older, no longer required
```
ares: this must be aligned correctly

manifests/install/repos/extra.pp
```diff
@@ -30,12 +30,7 @@
   }
 
   if $ipa_authentication and $configure_ipa_repo {
-    yumrepo { 'adelton-identity':
-      enabled  => 1,
-      gpgcheck => 0,
-      baseurl  => "http://copr-be.cloud.fedoraproject.org/results/adelton/identity_demo/epel-${osreleasemajor}-\$basearch/",
-      before   => Package['mod_authnz_pam', 'mod_lookup_identity', 'mod_intercept_form_submit', 'sssd-dbus'],
-    }
+    notify {"The --foreman-configure-ipa-repo option is deprecated and no longer needed. If you have used it before, disable and delete the repo.":}
```
ares: you'll have to use single quotes here (that's the reason why automated check fails), double qoute string is used when there's some $variable that should be interpolated

manifests/install.pp
```diff
@@ -39,12 +39,6 @@
       ensure  => $foreman::version,
       require => $repo,
     }
-    if $foreman::ipa_authentication and $foreman::configure_ipa_repo and $foreman::osreleasemajor == '6' {
-      package { 'mod_lookup_identity-selinux':
```
domcleal: Why remove this resource?

manifests/install/repos/extra.pp
```diff
@@ -30,12 +30,7 @@
   }
 
   if $ipa_authentication and $configure_ipa_repo {
-    yumrepo { 'adelton-identity':
```
domcleal: I don't think you should remove the resource, else that isn't really a deprecation, it's an incompatible removal.  It should probably remain until the next major module release.

manifests/install/repos/extra.pp
```diff
@@ -30,12 +30,7 @@
...
-      enabled  => 1,
-      gpgcheck => 0,
-      baseurl  => "http://copr-be.cloud.fedoraproject.org/results/adelton/identity_demo/epel-${osreleasemajor}-\$basearch/",
-      before   => Package['mod_authnz_pam', 'mod_lookup_identity', 'mod_intercept_form_submit', 'sssd-dbus'],
-    }
+    notify {'The --foreman-configure-ipa-repo option is deprecated and no longer needed. If you have used it before, disable and delete the repo.':}
```
domcleal: Use a notice(), not a notify so the warning is printed during compilation.

manifests/install.pp
ares: this is not needed anymore, if we remove the repo (below)

manifests/install/repos/extra.pp
ares: I thought the same but the thing is that the packages provided by this custom repo are now part of official repositories, the question is if we want to support installation on older RHEL/CentOS releases

adelton: > it's an incompatible removal

Our hope is that it's not incompatible because whatever the repo was providing is in base OSes by now.

manifests/install/repos/extra.pp
ares: I wanted it to be in reports coming from agents, we don't display master log anywhere AFAIK (in installer it does not matter that much), the message should be changed a bit so it does not refer to installer argument but rather class parameter right?

adelton: > but rather class parameter

Do we have many users that knowingly use these classes and class parameters? My guess would be that the vast majority would hit this code via `foreman-install` command.

OTOH, if the installer has some mechanism to mark the options as obsolete and notify the user even before running puppet, that might be the best course of action.

manifests/install.pp
adelton: And more specifically, it was only provided in the custom repo when the boolean was not yet available in the core selinux-policy-targeted.

manifests/install/repos/extra.pp
domcleal: > I wanted it to be in reports coming from agents, we don't display master log anywhere AFAIK (in installer it does not matter that much)

That'd be unusual, and would continually result in an active resource.  Using the parser functions for deprecations is the form.

> the message should be changed a bit so it does not refer to installer argument but rather class parameter right?

Indeed.

manifests/install/repos/extra.pp
ekohl: I agree with @domcleal: a deprecation means it still works but issues a warning. Removing the functionality means breaking the API and thus a major version bump. Now I don't care if we do break the API in this way, but it is breaking the API.

mmoll:
I don't use IPA, but that's looking OK. I guess that parameter and logic can be removed at the next major version bump?

ekohl:
I agree. If you can squash it into a single commit then we can merge it.

veruu:
Commits squashed (9dc2cdb1b1301d69c6a6187146f09abf6d14ce1f).

ekohl:
:+1:

mmoll:
Tested locally and tests were green, merging now :bomb:
Ďakujeme @veruu!

PR #334, opened 2015-07-13
fixes #11069 - add brightbox/passenger-legacy PPA for Passenger

domcleal:
Untested, will try to check it's all OK later.

http://projects.theforeman.org/issues/11069


mmoll:
tested, makes it work again. :+1:

ekohl:
Thanks!

sodabrew:
These passenger packages are terribly old and not actively maintained (that's why @johnl moved them to a separate repo). I hope they aren't still in use in Foreman 1.9 and up!?

domcleal:
@sodabrew they are if you're still using 12.04.  I'm not sure we have an alternative at the moment?

PR #337, opened 2015-07-20

igalic:
appear to have different levels of acceptance for such a certificate


manifests/init.pp
```diff
@@ -24,6 +24,8 @@
 #
 # $servername::                	Server name of the VirtualHost in the webserver
 #
+# $serveraliases::             	Server aliases of the VirtualHost in the webserver
```
ekohl: I think it could use ```type:array``` and a ```validate_array``` in the body.

ekohl:
Looking good.

mmoll:
tests fail :panda_face:

igalic:
:( worked on my machine :(

mmoll:
strange. when moving the validate_array to init.pp, it works

mmoll:
also, please rebase :)

ekohl:
> strange. when moving the validate_array to init.pp, it works

That's because in our tests we have no pre condition on ```foreman``` so the value is ```undef```. You need to set it to something even in our tests.

mmoll:
see #401

PR #338, opened 2015-07-20

igalic:
appear to have different levels of acceptance for such a certificate


manifests/init.pp
```diff
@@ -24,6 +24,8 @@
 #
 # $servername::               Server name of the VirtualHost in the webserver
 #
+# $serveraliases::             	Server aliases of the VirtualHost in the webserver
```
ekohl: Like in #337 ```type:array``` and ```validate_array```.

mmoll:
see #401

PR #339, opened 2015-07-23
Set explicit permissions

ekohl:
This matches the puppet default and it reduces differences if you set a
global default in site.pp.


mmoll:
:+1:

PR #340, opened 2015-07-24
Remove verify_*, enable param docs linting

domcleal:


ekohl:
:+1: on green tests.

domcleal:
I added cli.pp to the test list too, it's another top-level entry point.

mmoll:
thanks @domcleal!

PR #341, opened 2015-07-25
Use absolute variables

ekohl:
This shaves the tests from 27 minutes to 24 minutes on my local machine.


mmoll:
The travis runtime decreased a little with newer Rubies and fully qualified variables are always good, merging.

mmoll:
merged as 7649d8d

PR #342, opened 2015-07-25
Do not include foreman as a pre condition

ekohl:
This should save time on running tests


mmoll:
While the two long runners are still not under two hours, this is still bringing a noticeable difference.
bedankt @ekohl!

PR #343, opened 2015-07-29
Update install.pp

PierreR:
If passenger is set to `false`, we shouldn't install the `passenger ruby packages`.


domcleal:
:+1:

ekohl:
Thanks!

PR #344, opened 2015-07-30
Cosmetic update in service.pp

PierreR:
Don't use a `case` instead of an `if`.

If there is a good reason for the `case` could we at least replace `default, false: {` with `default: {` ?

### DIFF
--- a/manifests/service.pp
+++ b/manifests/service.pp
@@ -3,22 +3,19 @@ class foreman::service(
   $passenger = $::foreman::passenger,
   $app_root  = $::foreman::app_root,
 ) {
-  case $passenger {
-    true: {
-      exec {'restart_foreman':
-        command     => "/bin/touch ${app_root}/tmp/restart.txt",
-        refreshonly => true,
-        cwd         => $app_root,
-        path        => '/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin',
-      }
+  if $passenger {
+    exec {'restart_foreman':
+      command     => "/bin/touch ${app_root}/tmp/restart.txt",
+      refreshonly => true,
+      cwd         => $app_root,
+      path        => '/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin',
+    }
 
-      $service_ensure = 'stopped'
-      $service_enabled = false
-    }
-    default, false: {
-      $service_ensure  = 'running'
-      $service_enabled = true
-    }
+    $service_ensure = 'stopped'
+    $service_enabled = false
+  } else {
+    $service_ensure  = 'running'
+    $service_enabled = true
   }
 
   service {'foreman':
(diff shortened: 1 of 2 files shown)


manifests/service.pp
```diff
@@ -3,24 +3,22 @@
...
+      refreshonly => true,
+      cwd         => $foreman::app_root,
+      path        => '/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin',
     }
-    default, false: {
+
+    $service_ensure = 'stopped'
+    $service_enabled = false
+  }
+  else {
```
domcleal: `} else {` should be on the same line

manifests/service.pp
```diff
@@ -3,24 +3,22 @@
...
       $service_ensure  = 'running'
```
domcleal: indentation needs decreasing by two spaces

domcleal:
We should probably add a validate_bool to init.pp for $passenger to ensure it is always true/false.

manifests/service.pp
```diff
@@ -3,24 +3,22 @@
...
       $service_enabled = true
-    }
   }
-
+
```
ekohl: Trailing whitespace here?

ekohl:
Good idea, but I agree we should add ```validate_bool``` and fix some cosmetic nitpicks. Our policy is fast forward only and single (logical) commits so please squash your changes into this commit and rebase it. We will cherry pick and squash if needed, but then github doesn't detect it as merged and I like that purple Merged thingie :)

PierreR:
Ok I can do a proper PR. I have to admit that in doubt this was done sloppily with the github edit button ;-)

ekohl:
No problem. Sometimes you just need something to get started :)

PierreR:
Tests pass except for puppet 2.7 ... Is this a problem ?

mmoll:
The current stdlib master is not 2.7 compatible anymore since a few days, so that's expected.

manifests/service.pp
```diff
@@ -3,22 +3,19 @@
...
-
-      $service_ensure = 'stopped'
-      $service_enabled = false
-    }
-    default, false: {
-      $service_ensure  = 'running'
-      $service_enabled = true
+  if $passenger {
+    exec {'restart_foreman':
+      command     => "/bin/touch ${app_root}/tmp/restart.txt",
```
ekohl: Unrelated to your change, but I thought we removed ````/bin/```` here because there is no a path.

mmoll: That was in puppet-puppet for the puppet master restart, but can be applied here as well.

ekohl:
:+1: with or without the additional fix.

PierreR:
I think it is cleaner to separate concern.
I can do another PR that remove `/bin/` from the path as soon as this PR is merged if it helps.

mmoll:
The path/bin change is not relevant now anyway, as the module is Linux only and there's still a long way to go for the next nearest target (FreeBSD).

mmoll:
merci @PierreR!

PR #348, opened 2015-08-21
refs #4841 - change EL OSes to use 'tfm' SCL prefix

domcleal:
Parameters are supplied on the main foreman class to revert behaviour
for Foreman 1.9 and prior, plus on three plugin classes with
non-standard package names.

---

Merge after https://github.com/theforeman/foreman-packaging/pull/249

### DIFF
--- a/manifests/plugin/ovirt_provision/params.pp
+++ b/manifests/plugin/ovirt_provision/params.pp
@@ -1,0 +1,31 @@
+# Data for the ovirt_provision plugin
+class foreman::plugin::ovirt_provision::params {
+  case $::osfamily {
+    'RedHat': {
+      case $::operatingsystem {
+        'fedora': {
+          $package = 'rubygem-ovirt_provision_plugin'
+        }
+        default: {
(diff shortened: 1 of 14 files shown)


manifests/plugin/ovirt_provision/params.pp
```diff
@@ -0,0 +1,31 @@
...
+        }
+      }
+    }
+    'Debian': {
+      $package = 'ruby-ovirt-provision-plugin'
+    }
+    'Linux': {
+      case $::operatingsystem {
+        'Amazon': {
+          $package = 'tfm-ovirt_provision_plugin'
```
ekohl: Shouldn't this match the EL package name?

domcleal: Good catch, seems the original was wrong too.

ekohl:
:+1:

mmoll:
looking good, @domcleal merge whenever the rest is ready :)

domcleal:
Thanks for the reviews, merged now.

PR #350, opened 2015-08-25
Allow newer apt module

edestecd:
I need a newer version of apt for other module dependencies.
So far this does not break anything in foreman, for me.

### DIFF
--- a/metadata.json
+++ b/metadata.json
@@ -26,7 +26,7 @@
     },
     {
       "name": "puppetlabs/apt",
-      "version_requirement": "< 2.0.0"
+      "version_requirement": "< 3.0.0"
     },
     {
       "name": "puppetlabs/concat",


mmoll:
Unfortunately this story is a bit more complicated...

What would really be needed here is version range which includes 1.x and >= 2.1, but not 2.0.x, as 2.0 broke compatibilty in some ways, while 2.1 restored it. AFAICT, this is not possible with Puppet's capabilities in metadata.json.

So >= 2.1 would be the goal, but as this module is used in foreman-installer, all modules used by it and their dependencies would need to switch over to be ok WIth 2.1.x. If you want to take care of this, it would be really appreciated! :doughnut:

domcleal:
Or just say it's compatible with 1.x and 2.x, if 2.1 restored compatibility.  Seems more forgiving than trying to blacklist a particular version that we know isn't working.

mmoll:
then `< 3.0.0` would be OK, I guess

ekohl:
I wonder if we should document that 2.0 is broken. Release notes?

edestecd:
```< 3.0.0``` with doc it is then.

As far as the foreman-installer goes (1.9-stable branch) all other modules are good with the newest version of apt already.
The only change I had to make was to bump postgresql in the Puppetfile.

```ruby
mod 'puppetlabs/postgresql',    '< 4.5.0'
```

Shall I make a pull request to the foreman-installer doing that?

mmoll:
Nope, 1.9 will keep the version constraints the whole 1.9 lifecycle and it's open ended in the develop branch which will be 1.10.

Is apt 2.1.x also good with all our modules in their respective metadata.json?

mmoll:
why >= 1.8, BTW?

edestecd:
Yes, apt 2.1.x is good in the metadata.json of all the modules.
The only other module that has a metadata.json dependency is postgresql and starting wit 4.4.0, its ok
with newer apt.

edestecd:
>= 1.8, because I copied the dependency from the postgresql module metadata.json

mmoll:
I'm :+1:, once `>= 1.8.0` part is deleted :)

edestecd:
Want some doc about 2.0.x apt being bad news? If so then where?

mmoll:
As this is a topic of puppetlabs-apt and concerning all modules using apt, I don't think we really need to document this (just my two cents).

edestecd:
OK, removed the ```>= 1.8.0```

mmoll:
merged, thanks @edestecd!

PR #352, opened 2015-08-31
Add plugin_prefix option on main class

mmerfort:
This allows overriding the default which is set in params.pp and used in
plugin.pp.

* Add parameter to main class
* Add documentation

### DIFF
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -19,6 +19,8 @@
 # $passenger_ruby_package::    	Package to install to provide Passenger libraries for the active Ruby
 #                              	interpreter
 #
+# $plugin_prefix::              String which is prepended to the package name
+#
 # $use_vhost::                 	Enclose apache configuration in <VirtualHost>...</VirtualHost>
 #                              	type:boolean
 #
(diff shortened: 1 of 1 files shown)


ekohl:
Looks sane, but I wonder what your use case is. Do you rebuild all packages with a custom prefix?

mmerfort:
The defined type foreman::plugin references foreman::plugin_prefix [here](https://github.com/theforeman/puppet-foreman/blob/master/manifests/plugin.pp#L3).

However this parameter does currently not exist.

ekohl:
It should, since ```foreman``` inherits from ```foreman::params```.

mmerfort:
That's true, however for Foreman Version 1.9 and 1.8 the README suggests setting plugin_prefix          to “ruby193-rubygem-foreman_” which is not easily possible without this change.

ekohl:
That's a very good reason I should have come up with myself :)

mmoll:
:+1: after fixing indentation in the header

mmerfort:
I don't see the wrong indentation in my sources: https://github.com/SC-Networks/puppet-foreman/blob/plugin_prefix/manifests/init.pp

Any idea why they are shown this way in the commit?

mmoll:
Are you, by any chance, using hard tabs instead of spaces?

manifests/init.pp
```diff
@@ -19,6 +19,8 @@
 # $passenger_ruby_package::    	Package to install to provide Passenger libraries for the active Ruby
 #                              	interpreter
 #
+# $plugin_prefix::              String which is prepended to the package name
```
domcleal: *plugin* package names

mmoll:
As it turned out, it was a mix of tabs and spaces, I converted all tabs to spaces and also used "plugin package names". Danke @mmerfort!

mmerfort:
Thank you very much.

PR #355, opened 2015-09-03
The --foreman-configure-ipa-repo option is no longer required

veruu:
The --foreman-configure-ipa-repo option is no longer required since the
packages are included in RHEL/CentOS versions 6 and 7 now.

This feature was needed in older RHEL/CentOS versions, in which some
packages weren't available. Now there is no need for it.

After including a deprecation warning (pull request #333), we believe
that with new release the option can be removed completely.

### DIFF
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -155,10 +155,6 @@
 #
 # $pam_service::                PAM service used for host-based access control in IPA
 #
-# $configure_ipa_repo::         DEPRECATED: Enable custom yum repo with packages needed for external authentication via IPA,
-#                               this was needed on RHEL 6.5 and older, no longer required
-#                               type:boolean
-#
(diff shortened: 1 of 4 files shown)


mmoll:
:+1:

mmoll:
@domcleal @ekohl ping?
Per semver we'd need to bump the major version after that, but I'm unsure if that paramater is really in wide use.

ekohl:
:+1: if we bump the major version.

domcleal:
Just merge this whenever we come to do a major release I guess, there's no reason to rush into it now that I can think of.

ekohl:
Looks like with https://github.com/theforeman/puppet-foreman/pull/364 will do a major release.

ekohl:
Thanks!

Issue #331, opened 2015-06-25
should the module update plugins before running rake tasks?


cristifalcas:
In init.pp there is this dependency:

      Class['foreman::database']~>
      Foreman::Plugin <| |>

Shouldn't the package first be installed and after that run db::migrate? I think that at least openscap installs files for db::migrate

domcleal:
I don't remember if there's a specific reason it's that way around.  It might have been our unfamiliarity with how plugins would work in practice at the time, or it might be covering up some bugs in how plugins work (i.e. whether you can take a blank DB, activate plugins and db:migrate from scratch without the plugin breaking it).

That said, it works at the moment because all packages for plugins that contain DB migrations or seeds also call them from their post install sections, so installing the package after should perform any necessary steps.  (This is to ensure plugin upgrades also migrate DBs etc.)

mmoll:
what's the status here?

cristifalcas:
i will close it

PR #357, opened 2015-09-07
add a katello plugin

cristifalcas:
This PR is adding the possibility to install and configure the katello plugin in foreman.

It will most probably fail if the foreman database is already populated. Currently it was only tested with upgrades.

Based on the discussion from issues/331 and mostly from the katello installation module, the order on which plugins and db scripts are executed was changed.


ekohl:
Before we get into the code: would this make puppet-katello redundant? And do you mean https://github.com/theforeman/puppet-foreman/issues/331 as issue 331?

cristifalcas:
This doesn't manage qpid-config and guterball installation

ekohl:
So is the intention to ```include ::foreman::plugin::katello``` in puppet-katello? To be clear: I do think this could be a very good step, just trying to understand your goals.

cristifalcas:
My goal is to manage katello together with foreman and separately from pulp/qpid/etc (which we have on other machines).

I noticed now that the original module also has some providers for katello. Should I copy them here?

manifests/init.pp
```diff
@@ -2,193 +2,191 @@
 #
 # === Parameters:
 #
-# $foreman_url::                URL on which foreman is going to run
```
ekohl: This appears to be unrelated. Maybe it needs a rebase?

manifests/init.pp
```diff
@@ -201,7 +199,6 @@
   $passenger                 = $::foreman::params::passenger,
   $passenger_ruby            = $::foreman::params::passenger_ruby,
   $passenger_ruby_package    = $::foreman::params::passenger_ruby_package,
-  $plugin_prefix             = $::foreman::params::plugin_prefix,
```
ekohl: This is a desired change that you're reverting.

manifests/plugin/katello.pp
```diff
@@ -0,0 +1,134 @@
...
+  validate_bool($loggers_glue)
+  validate_bool($loggers_pulp_rest)
+  validate_bool($loggers_cp_rest)
+  validate_bool($loggers_cp_proxy)
+  validate_bool($loggers_action)
+  validate_bool($loggers_tire_rest)
+  validate_bool($loggers_manifest_import_logger)
+
+  if $configure_katello_repo {
+    case $::osfamily {
```
ekohl: Maybe this belongs in ```foreman::install::repos::katello```?

manifests/plugin/katello.pp
```diff
@@ -0,0 +1,134 @@
...
+  include ::foreman::plugin::tasks
+
+  foreman::config::passenger::fragment { 'katello':
+    content     => template("${module_name}/plugins/katello_httpd.erb"),
+    ssl_content => template("${module_name}/plugins/katello_httpd.erb"),
+  }
+
+  foreman::plugin { 'katello':
+    package     => $package,
+    config_file => '/etc/foreman/plugins/katello.yaml',
```
ekohl: You can reuse ```$::foreman::plugin_config_dir```

templates/plugins/katello_httpd.erb
```diff
@@ -0,0 +1,25 @@
+#
+# WARNING: THIS CONFIGURATION WAS GENERATED BY KATELLO-CONFIGURE TOOL,
+# CHANGES WILL LIKELY BE OVERWRITTEN.
+#
+
+SSLOptions +StdEnvVars +ExportCertData +FakeBasicAuth
+
+Alias /pub /var/www/html/pub
```
ekohl: Are sure Katello needs this? I think this belongs to pulp.

templates/plugins/katello_httpd.erb
```diff
@@ -0,0 +1,25 @@
...
+<Location /pub>
+Options +FollowSymLinks +Indexes
+</Location>
+
+<LocationMatch /rhsm|/subscription|/katello/api>
```
ekohl: Is rhsm and subscription part of the plugin? I thought that was a separate process which should configure its own apache.

mmoll:
1. What's the status here?
2. Personally I'm unsure if this should go here, especially with the whole scenarios work and its background of making a unified installer for Foreman and Katello more likely

ekohl:
Personally I'd like the case where Katello is just a plugin, but I think the rest of the ecosystem is not yet ready for this. I don't think it will be any time soon.

@ehelms what's your opinion on this?

ehelms:
I don't see it in the near future either, until we handle a few other pieces and figure out a strategy for how to have Katello but without it's services directly available or available on another box as a first class citizen.

ekohl:
When it's evolved enough, we should reconsider this but I'm going to close this for now.

PR #362, opened 2015-10-02
Prefer agent location of SSL CRL by default

lazyfrosch:
CRL under ssl/ca/ca_crl.pem should only be available on a Puppet CA.

So when Foreman is running on any other host, the Apache config is not
working.


lazyfrosch:
The travis test seems to have timed out...

mmoll:
:+1: - I did only have a look onto the ssldir on Puppet 3.8 machines, does this change work with 2.7 and 4.0 also?

fr34k8:
Yep

lazyfrosch:
@mmoll I think it will be the same, since `ca/` is the special path for CA content.

Could someone restart the travis test? I can't...

mmoll:
@lazyfrosch these two specific combinations of Ruby and Puppet take >2h since quite a while, so that's expected to error out.

:+1: for merging.

lazyfrosch:
@mmoll good to know

ekohl:
Thanks!

PR #365, opened 2015-10-09
add ExportCertData option needed by #12127

timogoebel:
This PR exports the client certificate to the rails env, needed by http://projects.theforeman.org/issues/12127


domcleal:
:+1:

timogoebel:
Tests fail for 1.8.7/ Puppet 3.5, but this looks like a travis issue.

```
Your test run exceeded 120 minutes. 
```

domcleal:
Yep, it's a known problem - sorry for the false positive, we'll just ignore it.

mmoll:
Shouldn't do any harm when not using the exported data, so merged. Danke @timogoebel!

PR #370, opened 2015-10-24
whitelist FreeBSD also in plugins

mmoll:
otherwise foreman-installer won't work, even with the foreman module disabled:
`Error: fibsd: ovirt_provision_plugin does not support osfamily FreeBSD at /usr/local/share/foreman-installer/modules/foreman/manifests/plugin/ovirt_provision/params.pp:28 on node fibsd.local`

### DIFF
--- a/manifests/plugin/ovirt_provision/params.pp
+++ b/manifests/plugin/ovirt_provision/params.pp
@@ -24,6 +24,9 @@ class foreman::plugin::ovirt_provision::params {
         }
       }
     }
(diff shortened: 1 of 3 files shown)


mmoll:
I'd like to see this in 1.10 RC2 :stars:

domcleal:
Merged as 6689e58 with a whitespace fix.  Thanks @mmoll!

I'll try and do a patch release of the module with that cherry picked.

PR #371, opened 2015-10-29
plugins:cockpit - Initial Cockpit support

dLobatog:


manifests/plugin/cockpit.pp
```diff
@@ -0,0 +1,5 @@
+# Installs foreman_bootdisk plugin
```
domcleal: Does it? :)

domcleal:
Could you add a spec too?  Copy one of the basic ones that only tests a single OS, please.

mmoll:
merged, gracias @dLobatog!

PR #368, opened 2015-10-18
Use rspec-puppet-facts where possible

ekohl:
This is pulled from https://github.com/theforeman/puppet-foreman/pull/332 and should be in a mergeable state.

PR #372, opened 2015-11-06
fix + refactor foreman::plugin::discovery test

brandonweeks:
screen shot 2015-11-06 at 1 46 20 am

The current test doesn't actually run because of `if facts[:os] == 'Fedora'`, I've included the refactored test from theforeman/puppet-foreman_proxy#203 that resolves the bug and improves the clarity.


ekohl:
You're right and I also discovered this, but the fix is hidden in https://github.com/theforeman/puppet-foreman/pull/368. I think it's fine to merge this and I'll rebase my PR on it once this is merged.

mmoll:
merged, thanks @brandonweeks!

PR #374, opened 2015-11-11
Fixes #12442 - Added a vhost to serve as internal infra port

parthaa:
Refer to this thread
https://groups.google.com/forum/#!topic/foreman-dev/9drbACTqbiw
and the attached image in this link in particular
https://groups.google.com/d/msg/foreman-dev/9drbACTqbiw/g67HXskJBwAJ

Basically we need to open up an infrastructure port so that components
like smart proxies can connect to it and communicate.

 This commit addresses the initial step in opening up a vhost. The next
step would be to get other services to talk to this.

### DIFF
--- a/manifests/config/passenger.pp
+++ b/manifests/config/passenger.pp
@@ -51,6 +51,10 @@ class foreman::config::passenger(
   $prestart            = $::foreman::passenger_prestart,
   $min_instances       = $::foreman::passenger_min_instances,
   $start_timeout       = $::foreman::passenger_start_timeout,
+  $infrastructure_port = 9070,
+  $client_ca           = $::foreman::client_ssl_ca,
+  $client_key          = $::foreman::client_ssl_key,
+  $client_cert         = $::foreman::client_ssl_cert,
 ) {
   # validate parameter values
   if $listen_on_interface {
@@ -148,7 +152,7 @@ class foreman::config::passenger(
         recurse => true,
       }
 
-      apache::vhost { 'foreman-ssl':
+      apache::vhost { 'foreman-infrastructure':
         add_default_charset     => 'UTF-8',
         docroot                 => $docroot,
         ip                      => $listen_interface,
@@ -158,10 +162,10 @@ class foreman::config::passenger(
         passenger_pre_start     => $https_prestart,
         passenger_start_timeout => $start_timeout,
         passenger_ruby          => $ruby,
-        port                    => 443,
+        port                    => $infrastructure_port,
         priority                => '05',
         servername              => $servername,
-        serveraliases           => ['foreman'],
+        serveraliases           => ['foreman_infra'],
         ssl                     => true,
         ssl_cert                => $ssl_cert,
         ssl_key                 => $ssl_key,
@@ -174,6 +178,67 @@ class foreman::config::passenger(
         ssl_verify_depth        => '3',
         custom_fragment         => template('foreman/_assets.conf.erb', 'foreman/_ssl_virt_host_include.erb',
                                             'foreman/_suburi.conf.erb'),
+      }
+
+      $bundle_cert_dir = '/usr/share/foreman/ssl'
+      $client_bundle = "${bundle_cert_dir}/client_bundle.pem"
+
+      file { $bundle_cert_dir:
+        ensure => 'directory',
+        owner  => 'root',
+        group  => $::foreman::group,
+        mode   => '0750',
+      }
+
+      concat {$client_bundle:
+        owner => 'root',
+        group => $::foreman::group,
+        mode  => '0640',
+      }
+
+      concat::fragment{ 'client_key':
+        target => $client_bundle,
+        source => $client_key,
+        order  => '01',
+      }
+
+      concat::fragment{ 'client_cert':
+        target => $client_bundle,
+        source => $client_cert,
+        order  => '02',
+      }
+
+      $infrastructure_url = "https://${servername}:${infrastructure_port}"
+      $infrastructure_redirection_path = '/'
+      apache::vhost { 'foreman-ssl':
+        add_default_charset => 'UTF-8',
+        docroot             => $docroot,
+        ip                  => $listen_interface,
+        options             => ['SymLinksIfOwnerMatch'],
+        port                => 443,
+        priority            => '05',
+        servername          => $servername,
+        serveraliases       => ['foreman'],
+        ssl                 => true,
+        ssl_cert            => $ssl_cert,
+        ssl_key             => $ssl_key,
+        ssl_chain           => $ssl_chain,
+        ssl_ca              => $ssl_ca,
+        ssl_crl             => $ssl_crl_real,
+        ssl_crl_check       => $ssl_crl_check,
+        ssl_verify_client   => 'optional',
+        ssl_options         => '+StdEnvVars +ExportCertData',
+        ssl_verify_depth    => '3',
+        ssl_proxyengine     => true,
+        proxy_pass          => [{
+          'path'         => $infrastructure_redirection_path,
+          'url'          => $infrastructure_url,
+          'reverse_urls' => [$infrastructure_redirection_path, $infrastructure_url]
+        }],
+        custom_fragment     => "
+        SSLProxyCACertificateFile ${client_ca}
+        SSLProxyMachineCertificateFile ${client_bundle}
+        ",
       }
     }
   } else {


domcleal:
Looks like this is a Katello change, perhaps make it there?

ekohl:
I agree at least partly with @domcleal. I'm not against it per se, but I don't see any benefit to this change in the current form. It looks like there's no service behind it and no way to extend this (from other modules or through parameters).

parthaa:
@domcleal @ekohl 
Maybe its a bad assumption, but this is more of a fundamental change from what I gather. From the email thread there the goal seems to be to move to the following architecture.
Newer Certs Arch

This is intended to be Stage 1. Assumption is that services like smart proxy will start using this once we get this in.

ekohl:
@parthaa I get that, but the current patch doesn't add any value: there's no service behind the vhost and no way for other modules to hook into it. I don't mind merging patches with partial functionality, but it has to add some value.

manifests/config/passenger.pp
```diff
@@ -108,6 +108,17 @@
       recurse => true,
     }
 
+    apache::vhost { 'foreman-infrastructure':
```
ehelms: This needs to be an SSL based vhost.

ehelms:
I think the initial value would come with a corresponding puppet-foreman_proxy PR that changes the default port to connect to this 9070 port. Then the web interfaces SSL certificates are de-coupled from the "infrastructure" certificates and easier for a user to change them out (i.e. they wouldn't have to change all their existing certificates for smart proxies down the road). This applies largely to greenfield deployments, however, it shouldn't break existing deployments as smart proxies connecting to 443 should still work as well (even if the user were to change out the web interface certificates).

ekohl:
I think that this PR should be changed to duplicate the SSL-vhost definition with a changed port. Since most users won't need this (I mostly see it as an advanced user/katello feature), IMHO it should be under a parameter that defaults to off. Maybe ```infrastructure_port``` and if it's not ```0``` then the vhost should be enabled.

domcleal:
Yeah, it would also increase out of the box memory usage by running another app instance.

ohadlevy:
I also support this change, and would allow a nice way to decouple from puppet ca, but we are missing the additional pr's to actually consume this from other services.
further, as @domcleal mention, this should be disabled by default, enabled probably by an explicit requirement, so that we don't increase the server load (number of passenger apps running) by default even when serving only one set of certificates.
I would rather introduce something like iptables DNAT (or apache proxy etc) vs have another app deployment that effect server load

ekohl:
@ohadlevy I also thought of some redirection, but think I'd prefer an apache proxy over iptables DNAT for simplicity of deployment.

ohadlevy:
@ekohl fine, how about mod_proxy then?

ehelms:
Roughly how much overhead would proxying vs another application instance produce?  

The downside to having it off by default is if a user doesn't deploy it and then spins up a bunch of smart proxies. Then if they want to make use of the separation they are going to have to go update all of their smart proxies which the idea was to create a network and certificate setup that wouldn't require the user to go manually update a bunch of their infrastructure (whether it was certificates or ports). That is to say, from my understanding of the discussions around this previously, the goals were:

1) Make custom certificates easier to change out by the user
2) Separate user facing network connections from "infrastructure" network and increase predictability
3) Make infrastructure connections (e.g. adding other components/services) extensible

ekohl:
@ohadlevy I think I'd start with that. Another benefit of mod_proxy is that you can use different certificates, but passing the correct certificate data might be interesting. That makes me think we should make the infrastructure port the actual passenger instance and proxy port 443 since most web users won't authenticate with client certificates.

@ehelms I think the overhead of mod_proxy is negligible so I think we could enable that by default. Running a full passenger stack (especially if we preload it like we do with our current foreman vhost) means quite a bit of memory, which makes it unsuitable for a default that won't be used by most users.

manifests/config/passenger.pp
```diff
@@ -175,6 +176,28 @@
         custom_fragment         => template('foreman/_assets.conf.erb', 'foreman/_ssl_virt_host_include.erb',
                                             'foreman/_suburi.conf.erb'),
       }
+
+      apache::vhost { 'foreman-ssl':
```
ehelms: This will require adding:

```
include apache::mod::proxy
include apache::mod::proxy_http
```

templates/_infrastructure_port_proxy.rb
```diff
@@ -0,0 +1,3 @@
+  ProxyPass / https://<%= @servername %>:9070
+  ProxyPassReverse / https://<%= @servername %>:9070
+  SSLProxyEngine On
```
ehelms: From some testing, this aspect of it will also require the following:

```
  ProxyPreserveHost on
  RequestHeader set X_FORWARDED_PROTO 'https'
  SSLProxyEngine On
  SSLProxyCACertificateFile "/var/lib/puppet/ssl/certs/ca.pem"
  SSLProxyMachineCertificateFile "/vagrant/foreman-installer/certs-bundle.pem"
```

This means we have to somewhat couple the 443 proxy port to the infrastructure port by making it aware of the CA that the infrastructure port is using (configured via  SSLProxyCACertificateFile). The other aspect is that we will have to generate a bundled key and certificate from the infrastructure CA and deploy it for the proxy to use (configured via SSLProxyMachineCertificateFile).

manifests/config/passenger.pp
ehelms: On second thought, you can configure everything without needing a template. Check out how we do the Capsule reverse proxy: https://github.com/Katello/puppet-capsule/blob/master/manifests/reverse_proxy.pp#L13

manifests/config/passenger.pp
```diff
@@ -175,6 +176,28 @@
...
+        add_default_charset     => 'UTF-8',
+        docroot                 => $docroot,
+        ip                      => $listen_interface,
+        options                 => ['SymLinksIfOwnerMatch'],
+        port                    => 443,
+        priority                => '06',
```
ekohl: I'd prefer keeping this at 05 so there's fewer changes for end users.

manifests/config/passenger.pp
```diff
@@ -51,7 +51,13 @@
   $prestart            = $::foreman::passenger_prestart,
   $min_instances       = $::foreman::passenger_min_instances,
   $start_timeout       = $::foreman::passenger_start_timeout,
+  $infrastructure_port = 9070,
+  $client_ca      = $::foreman::client_ssl_ca,
 ) {
+
+  $infrastructure_url = "https://${servername}:${infrastructure_port}"
+  $infrastructure_redirection_path = "/"
```
ehelms: I'd put these just above the vhost definition since thats where they are used (localizes them for context).

ehelms:
To create `  SSLProxyMachineCertificateFile "/vagrant/foreman-installer/certs-bundle.pem"`, I think you could use concat as the first step to combine the `ssl_cert` and `ssl_key` parameters into a bundle for it. As to where to deploy that file, we could either opt for introducing cert storage in:

/etc/foreman/ssl
/etc/foreman/
/etc/pki/foreman

Currently my prototype work with puppet-foreman_certs concept is using the latter -- https://github.com/ehelms/puppet-foreman_certs/blob/master/manifests/params.pp#L3

manifests/config/passenger.pp
```diff
@@ -175,6 +180,67 @@
         custom_fragment         => template('foreman/_assets.conf.erb', 'foreman/_ssl_virt_host_include.erb',
                                             'foreman/_suburi.conf.erb'),
       }
+
+      $pki_dir = '/etc/pki/foreman'
```
ekohl: `/etc/pki` is something that only exists on RH-based distros. I'm not sure what would be the best place for this that's generic for all distros.

manifests/config/passenger.pp
```diff
@@ -175,6 +180,67 @@
...
+        ssl_verify_depth        => '3',
+        ssl_proxyengine         => true,
+        proxy_pass              => [{
+          'path'         => $infrastructure_redirection_path,
+          'url'          => $infrastructure_url,
+          'reverse_urls' => [$infrastructure_redirection_path, $infrastructure_url]
+        }],
+        custom_fragment   => "
+          SSLProxyCACertificateFile ${client_ca}
+          SSLProxyMachineCertificateFile ${client_bundle}
```
ekohl: I'm not familiar with this, but am I reading it right that this does SSL client authentication against the infra vhost using this bundle? Wouldn't that open up this vhost rendering our authentication useless?

manifests/config/passenger.pp
ehelms: Good to know, searching around Debian has `/etc/ssl`. We could either set to path depending on the distro or follow the Puppet model and do something like `/usr/share/foreman/ssl` or `/var/lib/foreman/ssl` or `/etc/foreman/ssl`

manifests/config/passenger.pp
ehelms: It does - however, as I understand how it works (from our prior use case), the vhost itself will do authentication first using the SSL configuration for it. Then if that authentication passes, it will proxy the request to the infrastructure vhost using the client bundle.

ekohl: I think the setup would become something like:

```
Client <-> foreman-ssl vhost <-> foreman-infrastructure vhost
```

The ```Client``` can (but doesn't have to) present a client certificate. If it doesn't, won't that mean the ```foreman-ssl vhost``` will present a valid client certificate to ```foreman-infrastructure vhost```?

manifests/config/passenger.pp
ekohl: I think we should prefer distro locations if possible, but fall back to something generic.

manifests/config/passenger.pp
```diff
@@ -126,6 +130,7 @@
                                           'foreman/_suburi.conf.erb'),
     }
 
+
```
ekohl: Nitpick: please remove this line.

manifests/config/passenger.pp
```diff
@@ -158,10 +163,10 @@
         passenger_pre_start     => $https_prestart,
         passenger_start_timeout => $start_timeout,
         passenger_ruby          => $ruby,
-        port                    => 443,
+        port                    => 9070,
```
ekohl: Use ```$infrastructure_port``` here?

manifests/config/passenger.pp
ehelms: When communicating to the foreman-ssl vhost (443) ? or the instrastructure vhost (9070) ?

ekohl: In my scenario the client communicates with foreman-ssl vhost.

ehelms: If the client doesn't present a valid SSL certificate to the foreman-ssl vhost, it would get rejected by the SSL configuration present there or am I missing some case where a client can bypass the SSL configuration of 443?

ekohl: Yes, a normal web browser usually doesn't do client certificates. API authentication is also username/password or OAuth. Note the ssl_verify_client is set to optional.

I highly suspect the cert is also used for authorization so I think it's also very important that the right certificate (data) is presented to ```foreman-infrastructure```.

manifests/config/passenger.pp
```diff
@@ -175,6 +179,67 @@
...
+        order  => '01',
+      }
+
+      concat::fragment{ 'client_cert':
+        target => $client_bundle,
+        source => $client_cert,
+        order  => '02',
+      }
+
+      $infrastructure_url = "https://${servername}:${infrastructure_port}"
```
ehelms: This needs to have a trailing `/`

manifests/config/passenger.pp
ehelms: Upon further testing, I was perhaps wrong about needing `SSLProxyMachineCertificateFile` as you mentioned @ekohl . I was previously not seeing the smart proxy register correctly but it may have been to (https://github.com/theforeman/puppet-foreman/pull/374#discussion_r46178225). I am going to try not having that directive, and providing a completely different set of certificates for 443 than puppet on both sides to confirm.

manifests/config/passenger.pp
```diff
@@ -175,6 +179,67 @@
...
+        ssl_verify_depth    => '3',
+        ssl_proxyengine     => true,
+        proxy_pass          => [{
+          'path'         => $infrastructure_redirection_path,
+          'url'          => $infrastructure_url,
+          'reverse_urls' => [$infrastructure_redirection_path, $infrastructure_url]
+        }],
+        custom_fragment     => "
+        SSLProxyCACertificateFile ${client_ca}
+        SSLProxyMachineCertificateFile ${client_bundle}
```
ehelms: After some testing using (https://github.com/iNecas/ownca) to generate custom certificates for 443 port, I agree with @ekohl that this is not needed as I was able to do a full install which included a smart proxy registering to 443 successfully.

manifests/config/passenger.pp
```diff
@@ -175,6 +179,67 @@
...
+        ssl_options         => '+StdEnvVars +ExportCertData',
+        ssl_verify_depth    => '3',
+        ssl_proxyengine     => true,
+        proxy_pass          => [{
+          'path'         => $infrastructure_redirection_path,
+          'url'          => $infrastructure_url,
+          'reverse_urls' => [$infrastructure_redirection_path, $infrastructure_url]
+        }],
+        custom_fragment     => "
+        SSLProxyCACertificateFile ${client_ca}
```
ehelms: This has to be in quotes:

```
SSLProxyCACertificateFile \"${client_ca}\"
```

lazyfrosch:
Please consider #389 as well :)

ohadlevy:
is anyone still working on this?

mmoll:
what's the status here?

ehelms:
@lazyfrosch Looking at #389, that would seem to solve the verifying clients that have certificates signed by a CA different than what the web server is using; however, would it solve the reserve direction of the client verifying the server? Anytime the certificates changed for the we server, the clients would all need to be updated still I think. 

@mmoll We let this one sit unintentionally, but I still feel it is a change worth pursuing with respect to the overall ecosystem. I'd take over with a new PR but I don't want to lose the conversation hitsory. I'll talk to @parthaa to see if we can push an update for review.

lazyfrosch:
@ehelms I'm not sure if it is wise to fiddle with ssl_dir for such a feature.

I think an infrastructure vhost is a good idea, since it would allow the "user" vhost to have a cert that is "publicly" or company wide trusted, and not the Puppet CA.

ekohl:
Is there a plan to revisit this?

mmoll:
no reaction, closing.

PR #376, opened 2015-11-23
refs #8389 - lower case Foreman hostname to match foreman_proxy

domcleal:
Lower case FQDN used in the Foreman URL for apipie-bindings (in Hammer)
to work correctly and also to match foreman_proxy to ensure correct
ordering of foreman_smartproxy resources in init.pp's resource
collector.

---

Fixes the issue I noted in https://github.com/theforeman/puppet-foreman_proxy/pull/148#issuecomment-158368123, simply due to case sensitive comparisons in the init.pp resource collector.

### DIFF
--- a/manifests/params.pp
+++ b/manifests/params.pp
@@ -1,8 +1,9 @@
 # The foreman default parameters
 class foreman::params {
+  $lower_fqdn = downcase($::fqdn)
 
 # Basic configurations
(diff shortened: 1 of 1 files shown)


ekohl:
:+1:

mmoll:
merged, thanks @domcleal!

PR #377, opened 2015-12-01
fix discovery plugin specs

mmoll:
I'm not sure why this is not working as expected, but it's a start...

### DIFF
--- a/spec/classes/foreman_plugin_discovery_spec.rb
+++ b/spec/classes/foreman_plugin_discovery_spec.rb
@@ -5,14 +5,22 @@ describe 'foreman::plugin::discovery' do
     next if only_test_os() and not only_test_os.include?(os)
     next if exclude_test_os() and exclude_test_os.include?(os)
 
-    let(:facts) { facts }
-
     context "on #{os}" do
+      let(:facts) { facts }
+
+      case facts[:operatingsystem]
+        when 'Debian'
+          tftproot = '/srv/tftp'
+        when 'FreeBSD'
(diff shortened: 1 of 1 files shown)


domcleal:
Is this intermittent?  Or did one of today's merges somehow fix it by changing which tests we run on Ruby 1.8.7?  (https://travis-ci.org/theforeman/puppet-foreman/builds/94114145)

domcleal:
Curious, same failure on foreman_proxy's modulesync branch: https://travis-ci.org/theforeman/puppet-foreman_proxy/jobs/94114204

mmoll:
the code is almost the same, regarding `include ::tftp::params` and using `$::tftp::params::root`

spec/classes/foreman_plugin_discovery_spec.rb
```diff
@@ -7,12 +7,18 @@
 
     let(:facts) { facts }
```
domcleal: Got it!  This should be inside the `context "on #{os}"` block.

When you run the whole file you're getting facts from the wrong OS, probably whichever OS is defined first or last.  This might be why it was affecting Ruby 1.8.7 tests in particular as the hash order isn't deterministic - while on other Ruby versions it was probably a single OS every time (something with /var/lib).

mmoll:
indeed, thanks! should turn green now. :)

ekohl:
:+1: on green tests

mmoll:
:green_heart:

ekohl:
Thanks!

PR #378, opened 2015-12-02
Fixes #12657 - replace deprecated $::selinux fact

brandonweeks:


mmoll:
According to Facter docs this fact is even there in Facter 1.6, so :+1:

domcleal:
selinux_enforced unfortunately only shows whether SELinux is enforcing or not, so it returns false in permissive mode, preventing the installation of the SELinux policy when it should really be configured.

Also, it looks like selinux_enforced is deprecated too: https://docs.puppetlabs.com/facter/3.1/core_facts.html#selinuxenforced (I assume that's what "legacy" means).

domcleal:
> (I assume that's what "legacy" means).

Ah no, simply unstructured facts.  I can't see any replacement for the unstructured SELinux facts in Facter master right now.

Which version of Facter were you getting `permissive` on?  Could you have a custom fact from another module that's overwriting the built in one?

brandonweeks:
https://ask.puppetlabs.com/question/16601/why-is-selinux-fact-different-for-puppet-then-for-facter/

This is a pretty through description of the issue I saw. Facter always returned a boolean, however puppetserver got 'permissive' and err'd on str2bool. Either way this PR doesn't make much sense anymore given your clarification.

brandonweeks:
Some logs: https://gist.github.com/brandonweeks/507d8d8e2228d72d20e3

I'm fairly convinced this isn't related to foreman whatsoever, just happened to be the module that caught it.

domcleal:
Yeah indeed, I see what you mean, but I also can't understand it in the slightest.  Does the facts YAML cache in /var/lib/puppet/yaml/facts/ also show the string?  Does it happen with `puppet apply`?  (I quickly tested on 3.8.4 on EL7 and see a bool in both apply and agent with a regular Passenger-based master.)

I think we'd need to accept either the bool or a string (either 'enforcing' or 'permissive') for a fix in this module since a bool is still the usual value.

brandonweeks:
Managed to reproduce this in another OpenStack module... :cry: Closing.

alexjfisher:
@brandonweeks Did you ever figure out why the selinux fact isn't always a boolean?

theforeman/puppet is failing for me too.
```
Debug: Caching connection for https://devops-puppet4.example.com:8140
Error: Could not retrieve catalog from remote server: Error 400 on SERVER: Evaluation Error: Error while evaluating a Resource Statement, Evaluation Error: Error while evaluating a Function Call, str2bool(): Unknown type of boolean given at /etc/puppetlabs/code/environments/CORE_NG_foreman_webfrontend/modules/foreman/manifests/install.pp:23:30 on node devops20.example.com
```

For me, it's the string `disabled`, but only during a puppet agent run.  With puppet apply, it's `false`.
This is with puppet 4.5 and puppetserver 2.4 on OEL7.

alexjfisher:
https://github.com/theforeman/community-templates/issues/285

alexjfisher:
@brandonweeks Do you have `selinux` set as a parameter in Foreman?

PR #379, opened 2015-12-02
added the param  puppetrun

TomRitserveldt:
Added the param "puppetrun" to the foreman settings.yaml template (and params+init manifests) to toggle the puppetrun functionality in foreman. This enables the "Run Puppet" button on individual host pages. default value is false, since that was the default before.

### DIFF
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -4,6 +4,9 @@
 #
 # $foreman_url::                URL on which foreman is going to run
 #
+# $puppetrun::                  Should foreman be able to start puppetruns on nodes
+#                               type: boolean
+#
 # $unattended::                 Should foreman manage host provisioning as well
 #                               type:boolean
 #
@@ -192,6 +195,7 @@
 #
 class foreman (
   $foreman_url               = $::foreman::params::foreman_url,
+  $puppetrun                 = $::foreman::params::puppetrun,
   $unattended                = $::foreman::params::unattended,
   $authentication            = $::foreman::params::authentication,
   $passenger                 = $::foreman::params::passenger,
@@ -287,6 +291,7 @@ class foreman (
   if $email_delivery_method {
     validate_re($email_delivery_method, ['^sendmail$', '^smtp$'], "email_delivery_method can be either sendmail or smtp, not ${email_delivery_method}")
   }
+  validate_bool($puppetrun)
 
   class { '::foreman::install': } ~>
   class { '::foreman::config': } ~>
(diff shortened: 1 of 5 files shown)


manifests/init.pp
```diff
@@ -4,6 +4,8 @@
 #
 # $foreman_url::                URL on which foreman is going to run
 #
+# $puppetrun::                  Should foreman be able to start puppetruns on nodes
```
ekohl: You can add ```type:boolean``` like ```$unattended``` does below.

manifests/init.pp
```diff
@@ -192,6 +194,7 @@
 #
 class foreman (
   $foreman_url               = $::foreman::params::foreman_url,
+  $puppetrun                 = $::foreman::params::puppetrun,
```
ekohl: This can be validated using ```validate_bool```.

ekohl:
I think this is a good change. The only downside I see is users who use the module and enabled puppetrun. If they update the module, it will be turned off unless they change the parameter. This can be fixed by writing a changelog entry though.

TomRitserveldt:
I've added the type to the docs above, as well as the bool validation.
You do make a good point regarding the default "false" value, I suppose one of you guys can/should update the changelog for this?

ekohl:
Yes, the changelogs are only written when we release. @domcleal do we have some system in place to track those changes? Add a label?

domcleal:
No we don't, I normally just go over the commit log.  I'm quite happy if you'd like to add an "unreleased" section at the top of CHANGELOG.md in this PR with any notes you think should be in there.

mmoll:
@TomRitserveldt could you plase add a test to `spec/classes/foreman_config_spec.rb`? :+1: otherwise.

TomRitserveldt:
can this PR be accepted? the only check failing is the one that failed in the base branch as well..

TomRitserveldt:
Can this be accepted and merged then?

mmoll:
I'd like to see a test for the enabled case also, but it's also OK as-is. Please squash the commits together into one, then we're ready to go here, IMHO. :+1: 

BTW, GitHub does not notify anybody when commits are added to PRs, that's why this PR went unnoticed. :disappointed:

mmoll:
also, please rebase onto current master and it would be nice if you could add an entry to CHANGELOG.md

TomRitserveldt:
rebased with your current master and added an extra entry in the CHANGELOG.md,
travis.ci check are green as well.

mmoll:
:+1:

ekohl:
Thanks!

PR #380, opened 2015-12-04
Configure Passenger to use foreman-ruby symlink

domcleal:
Using the foreman-ruby symlink installed by our packages by default on
all Debian-based OSes will make upgrades easier, as users won't need to
reconfigure Apache if we change the Ruby version used by the packages.


ekohl:
:+1: though I didn't test it.

mmoll:
Working fine on Debian/jessie, will merge when https://travis-ci.org/domcleal/puppet-foreman/builds/94852294 turned green.

ekohl:
@mmoll it's green.

mmoll:
indeed, merged, thanks @domcleal!

PR #383, opened 2015-12-22
Allow puppetdb plugin config settings to be managed.

jyaworski:


mmoll:
Please see the discussion in #235. Maybe this can be solved using `foreman_config_entry`, but I didn't have a closer look.

jyaworski:
@mmoll good point. It may be better to move this to plugin itself. I'll look into it.

mmoll:
The thing is that it already was moved into the plugin, but if it's set in a config file, it will still work (it's just not the prefered way anymore).

jyaworski:
@mmoll are you meaning that it queries the database setting? I read that on the plugin site that this isn't preferred, but would we rather have a file in-place for puppet, or have puppet query postgres?

mmoll:
The PuppetDB plugin has settings in Foreman (visible in the UI under Administer -> Settings -> Puppetdb) and the theforman puppet module has the `foreman_config_entry` defined type to set such settings.

I think that's the method preferred over the old method of having the config in a YAML file, @dLobatog, could you clarify here?

jyaworski:
Oh, wow. OK. I was looked for a defined DSL type, not a Ruby type. That's probably the better solution. I can refactor since that's available.

jyaworski:
@mmoll refactored.

brandonweeks:
What about using localhost as the default value?

jyaworski:
@brandonweeks good point. I've set it to localhost and squashed.

mmoll:
Maybe it's just for me because of some history of that install, but when I using this via -installer, the values are read wrongly and set every time:
```
[ WARN 2015-12-26 17:39:22 main]  /Stage[main]/Foreman::Plugin::Puppetdb/Foreman_config_entry[puppetdb_address]/value: value changed 'https' to 'https://foreman.local:8081/v2/commands'
[ WARN 2015-12-26 17:39:33 main]  /Stage[main]/Foreman::Plugin::Puppetdb/Foreman_config_entry[puppetdb_dashboard_address]/value: value changed 'http' to 'http://localhost:8080/dashboard'
```

mmoll:
#384 should fix that, this PR is fine. :+1:

jyaworski:
Thanks! Can this be merged?

ekohl:
I think this should be added to the changelog since it may break existing installations. That shouldn't hold back the merge though.

mmoll:
@jyaworski if you want to add this to the changelog, please do. Otherwise ping here and I'll do it while merging.

jyaworski:
@mmoll could you do it this time and ping me? Next time I'll add the entry; I just want to see how it's done.

mmoll:
See 84c0f60, it's just extending the changelog file.
Dzięki @jyaworski!

PR #384, opened 2015-12-26
fix key-value splitting in foreman_config_entry

mmoll:
found while testing #383, where the value is an URL, containing a colon


ekohl:
:+1:

ekohl:
Thanks

PR #346, opened 2015-08-14
Don't qualified call to defined resource type

PierreR:
Resource names are not limited by scope.

PR #386, opened 2016-01-06
Don't qualified call to defined resource type

PierreR:
- resource names are not limited by scope.
- it is considered a bad practice in Puppet 3
- it isn't tolerated anymore in Puppet 4.

### DIFF
--- a/manifests/database/postgresql.pp
+++ b/manifests/database/postgresql.pp
@@ -16,7 +16,8 @@ class foreman::database::postgresql {
   }
 
   include ::postgresql::client, ::postgresql::server
-  ::postgresql::server::db { $dbname:
+  
+  postgresql::server::db { $dbname:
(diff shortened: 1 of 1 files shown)


PierreR:
It is actually the same kind of change as  #346 (which has been accepted)

ekohl:
Untested, but :+1: 

I do wonder why our tests didn't pick this up though.

PierreR:
IIRC Puppet 4 does not search for relative path anymore. So it is kind of pointless to use `::` (even for the inclusion of classes).
To be honest I just don't know if Puppet 4 fails hard when using `::` on defined resource type; maybe it just ignores them. Anyhow I can't think of any good reason to keep them ;-)

mmoll:
merged, with removed trailing whitespace, thanks @PierreR!

PR #387, opened 2016-01-08
Easy way to extract values from foreman api search

scitechfh:
With puppet 3.x without future parser it difficult access foreman search results. This parser function converts easily any search result to an array.

### DIFF
--- a/lib/puppet/parser/functions/foreman.rb
+++ b/lib/puppet/parser/functions/foreman.rb
@@ -19,7 +19,14 @@
 #            Foreman if you didn't specify anything.
 # 'foreman_url' is your actual foreman server address
 # 'foreman_user' is the username of an account with API access
-# 'foreman_pass" is the password of an account with API access
+# 'foreman_pass' is the password of an account with API access
+# 'filter_result' string or array with attribites zu return
+#                 if a string is given foreman() returns an array only with given attributes
+#                 in case of an array is given foreman() returns an array of hashes selecting only
+#                 attributes given in array
+#                 in case of an given hash foreman() returns an array of hashes selecting only
+#                 attribute keys given in hash renamed to values of given keys. This can be used
+#                 to rename keys in result
 #
 # Then, use a variable to capture its output:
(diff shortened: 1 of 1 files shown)


lib/puppet/parser/functions/foreman_extract_from_search.rb
```diff
@@ -0,0 +1,35 @@
...
+#
+# The foreman_extract_from_search() parser takes a hash value with query results and
+# a second parameter as key to extract
+#
+# To use foreman_extract_from_search(), apply results of foreman() and a key.
+#
+# $hosts = foreman_extract_from_search(foreman($f),'name')
+#
+
+require "net/http"
```
ekohl: Are these requires actually used?

scitechfh: No. This is an copy and paste error.

ekohl:
Looks like it makes sense, but I'd like @mmoll or @domcleal to have a look.

domcleal:
I guess it's needed now because the default API version switched to v2 in Foreman 1.9, so the foreman() function now returns the pagination and results data instead of just the results.  We could change foreman() to always return results directly, but I guess that could break manifests just as badly.

ekohl:
@domcleal and a parameter in foreman() to return just the results?

domcleal:
Oh, I see this is doing something further.  Given a response like this:

    {"results": [ {"name":"foo"}, {"name":"bar"} ]}

It's extracting a key from the hashes inside the results list - a bit like a map() function.

So probably best to ignore my comment above - it may be useful, but it's separate.

scitechfh:
Whats about add an result filter attribute to foreman()?

domcleal:
That'd be fine by me too.

scitechfh:
I have removed my old function and added an filter function to foreman(). It's possible convert the result into an array or an hash with limited key or renamed keys.

domcleal:
Works great, merged as a41fce0.  Thanks @scitechfh and sorry for the delay.

PR #388, opened 2016-01-14
Align yum repos with foreman-release

ekohl:


manifests/install/repos/yum.pp
```diff
@@ -20,23 +20,30 @@
...
   }
-  yumrepo { "${name}-source":
-    descr    => "Foreman ${repo} source repository",
-    baseurl  => "http://yum.theforeman.org/${repo_path}/${yumcode}/source",
+  yumrepo { "${name}-plugins-source":
+    descr    => "Foreman plugins ${repo} - source",
+    baseurl  => "http://yum.theforeman.org/${plugins_repo_path}/${yumcode}/source",
     gpgcheck => $gpgcheck_enabled,
-    gpgkey   => "http://yum.theforeman.org/${repo_path}/RPM-GPG-KEY-foreman",
+    gpgkey   => "http://yum.theforeman.org/${plugins_repo_path}/RPM-GPG-KEY-foreman",
```
domcleal: no GPG key for plugins, or their sources

domcleal:
Looks correct otherwise, compared to https://github.com/theforeman/foreman-packaging/blob/rpm/develop/foreman/foreman.repo.

manifests/install/repos/yum.pp
ekohl: I now see that the release files use ```gpgkey=file:///etc/pki/rpm-gpg/RPM-GPG-KEY-foreman``` in https://github.com/theforeman/foreman-packaging/blob/rpm/develop/foreman/foreman-plugins.repo. Should we just set the gpgkey to the http version but gpgcheck off?

domcleal: Yep, that'd be fine.  I'm not sure what the URL should be, but this seems fairly safe (it's like to be a plugin specific key rather than the main release one, if/when we did it).

mmoll:
this needs a rebase and the comments above addressed :)

ekohl:
Updated.

manifests/install/repos/yum.pp
ekohl: I decided to change gpgcheck to ```'0'``` since that matches the repos. I left out the gpgkey parameter since it's not used anyway and this way it won't change the value from foreman-release.

domcleal:
Merged, thanks @ekohl!

PR #389, opened 2016-01-14
fixes #11652: set ssl_certs_dir to '' by default

lazyfrosch:
This will not set SSLCACertificatePath by default. And that way only request and authenticate certificates by the configured CA and not any other present in the certs directory.

On the security site, this might be a problem, allowing any CA in the default certs dir, to sign a cert with a Puppet-known hostname CN.

We should discuss this :wink:


lazyfrosch:
My main issue with this, is that my browser always asked me to auth with a local client certificate.

I think it would be best practice to change it like this, should not affect any user.

Not exactly sure about the security meaning.

ekohl:
It looks like the solution that http://projects.theforeman.org/issues/11652 suggests, so I think you can add that issue number in the commit message.

Other than that, isn't the local client certificate because you can auth with it? node.rb, smart-proxy and other scripts use this so I think it will have to stay until we merge https://github.com/theforeman/puppet-foreman/pull/374 which splits that off into a separate port.

lazyfrosch:
@ekohl added the issue #

Regarding client auth, this still works, since this is still set, and is the effective setting. But we are no longer including the system trust store.

No need for an "internal" vhost.

```
SSLCACertificateFile    "/var/lib/puppet/ssl/certs/ca.pem"
```

ekohl:
@lazyfrosch I know this doesn't break the client auth, but I'd expect that your browser will still prompt you for a client certificate if available. This should be easy to test :)

lazyfrosch:
Not any more, I have a client cert from StartSSL, now I don't get the request anymore.

ekohl:
@lazyfrosch that's good to know because I have the same issue :)

manifests/config/passenger.pp
```diff
@@ -44,6 +46,7 @@
   $ssl_ca              = $::foreman::server_ssl_ca,
   $ssl_chain           = $::foreman::server_ssl_chain,
   $ssl_cert            = $::foreman::server_ssl_cert,
+  $ssl_certs_dir       = '',
```
ekohl: puppet-lint does not like this :(

lazyfrosch:
Oh right, but wouldn't `undef` use the default value of `apache::vhost` ?

lazyfrosch:
I'd suggest an lint ignore comment here, since we need to empty the default value.

ekohl:
I agree. I think it's clear to set the value in params and add the lint ignore there rather than init.pp.

lazyfrosch:
Updated my commit.

Is it intentional to not test default values from `foreman` (`::params`) in passenger spec?

ekohl:
@lazyfrosch not sure if it's intentional. I think it's mostly to save runtime and possibly because this was easier.

lazyfrosch:
Lint fix...

ekohl:
:+1: on green tests

mmoll:
tested, still works for me :+1:
@domcleal, could you also have a look? :eyes:

domcleal:
Makes sense, merged.  Thanks @lazyfrosch.

PR #392, opened 2016-01-26
ensure json_package is installed via ensure_packages

timogoebel:
This commit changes the installation of rubygem-json from an explicit
package definition to ensure_packages. This may prevent a duplicate
declaration of the package resource.

### DIFF
--- a/manifests/puppetmaster.pp
+++ b/manifests/puppetmaster.pp
@@ -25,9 +25,7 @@ class foreman::puppetmaster (
     default:  { $json_package = 'rubygem-json' }
   }
 
-  package { $json_package:
-    ensure  => installed,
-  }
+  ensure_packages([$json_package])
 
   file {"${puppet_etcdir}/foreman.yaml":
     content => template("${module_name}/puppet.yaml.erb"),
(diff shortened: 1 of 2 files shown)


manifests/puppetmaster.pp
```diff
@@ -25,9 +25,7 @@
     default:  { $json_package = 'rubygem-json' }
   }
 
-  package { $json_package:
-    ensure  => installed,
-  }
+  ensure_packages($json_package)
```
ekohl: The [documentation](https://github.com/puppetlabs/puppetlabs-stdlib#ensure_packages) says it should be a list of packages.

timogoebel: It is casted to an array in the code, but let me update this...

ekohl: In that case the documentation doesn't match the code :)

ekohl:
:+1:

mmoll:
@timogoebel please update the tests (see the output in travis for the needed change), then we're good to go here :)

timogoebel:
@ekohl @mmoll : Thanks for taking a look. Tests are green now! :-)

mmoll:
Merged, Danke @timogoebel!

PR #393, opened 2016-01-29
Pin puppetlabs/apache to 1.7.0

ehelms:
The update to puppetlabs/apache 1.8.0 included moving mod configuration
files to /etc/httpd/conf.modules.d on EL7. The mod_passenger package
currently places its configuration file in /etc/httpd/conf.d which
gets wiped out by puppet. The configuration file deployed by puppet
does not include the proper default directives to set the PassengerRoot
which results in Foreman being unaccessible. See:

https://github.com/puppetlabs/puppetlabs-apache/commit/e825422b0080e66b32cd05ed51b5ccae69325c74

### DIFF
--- a/metadata.json
+++ b/metadata.json
@@ -22,7 +22,7 @@
     },
     {
       "name": "puppetlabs/apache",
-      "version_requirement": ">= 1.2.0 < 2.0.0"
+      "version_requirement": ">= 1.2.0 < 1.8.0"
     },
     {
       "name": "puppetlabs/apt",


jlsherrill:
This is believe to be the cause of http://ci.theforeman.org/job/systest_foreman/8558/

the failures only seem to occur on el7 and show up because PassengerRoot is not defined any apache config files

mmoll:
I guess this is http://projects.theforeman.org/issues/13405 ?

domcleal:
I guess it is.  I spoke to the maintainers yesterday and they said it'll be a few weeks before a patch release is made with my fix, so I was planning to pin this today - though usually we'd do it in foreman-installer.

domcleal:
The other thing we can do is to point foreman-installer to the 1.8.x branch (which contains the fix) rather than pin to 1.7.

ehelms:
Apologies, I missed that issue entirely -- I have moved the change to here https://github.com/theforeman/foreman-installer/pull/164

PR #395, opened 2016-02-06
plugins:ansible - Initial Ansible support

dLobatog:


mmoll:
:+1:

dLobatog:
If I made the plugin optionally install Ansible as well, how likely to be accepted would it be? I think it'd be useful because:

* The smart-proxy-ansible plugin (to run remote execution) needs Ansible the executable with the proper Foreman callback. 
* Deploying the callback (a python script) and configuring /etc/ansible.cfg so people don't have to would make it easier.

mmoll:
I would need to delve a bit more into how the connection between Foreman, Foreman-Proxy and Ansible is made. For Salt and Chef, the **proxy** plugin packages depend onto their respective master's package and both have support in puppet-foreman_proxy to configure their integration config files. Would this touch puppet-foreman for Ansible?

dLobatog:
@mmoll Don't worry about proxy at all now - foreman_ansible itself doesn't need the proxy for anything yet. At least until https://github.com/theforeman/foreman_remote_execution/pull/134 is merged we don't have to worry about that bit.

mmoll:
allright.

mmoll:
@dLobatog can you please rebase onto current master, so all tests are green again?

As this is installing the foreman core plugin and other configurations have to be handled elsewhere, this is good to go, IMHO.

dLobatog:
Updated!

mmoll:
merged, gracias @dLobatog!

PR #397, opened 2016-02-10
pin rspec-puppet-facts for Ruby 1.8

mmoll:
closes GH-396

will open a PR for modulesync, once this is merged here.

Issue #396, opened 2016-02-09
Test failing with Ruby 1.8.7


mmoll:
Pinning rspec-puppet-facts to 1.3.0 fixes this, the only change for 1.4.x was to pull in mcollective.

mmoll:
thinking about it again, I think it would be OK to pin rspec-puppet-facts to 1.3.0 for Ruby 1.8, but leave it open ended for other Rubyies. Opinions?

domcleal:
Is this regarding the report processor failures?  I'm not sure I see the connection.  e.g.

<pre>
       -"metrics" => {"resources"=>{"changed"=>2, "failed"=>1, "out_of_sync"=>2, "restarted"=>1, "total"=>10}, "events"=>{"failure"=>1, "success"=>1, "total"=>2}, "changes"=>{"total"=>2}, "time"=>{"filebucket"=>7.9e-05, "schedule"=>0.000678, "service"=>0.00463, "config_retrieval"=>0.141527891159058, "package"=>0.189955, "file"=>0.090157}},

       +"metrics" => {"resources"=>{"changed"=>2, "failed"=>1, "out_of_sync"=>2, "restarted"=>1, "total"=>10}, "changes"=>{"total"=>2}, "events"=>{"failure"=>1, "success"=>1, "total"=>2}, "time"=>{"filebucket"=>7.9e-05, "schedule"=>0.000678, "service"=>0.00463, "config_retrieval"=>0.141527891159058, "package"=>0.189955, "file"=>0.090157}},

        "reported_at" => "2013-08-29 11:11:44 UTC",

       -"status" => {"applied"=>0, "failed_restarts"=>0, "skipped"=>0, "pending"=>0, "failed"=>1, "restarted"=>1},

       +"status" => {"applied"=>0, "failed_restarts"=>0, "skipped"=>0, "pending"=>0, "failed"=>2, "restarted"=>2},

       

     # ./spec/unit/foreman_report_processor_spec.rb:43
</pre>

The only danger with pinning that gem at different version is when the fact values change and we have to adapt tests, it'll be much more difficult if it varies per Ruby version too.

mmoll:
[...]

PR #398, opened 2016-02-10
Copy report processor to lib

GregSutcliffe:
For now this is *just* a copy of what's in files/

I was about to delete it from files/ and update the tests/code, but it occured to me we still need the module in the environment path somewhere. How shall we handle this? I can think of:

* Copy the foreman module to the first available env (can we figure this out if the user changes options?)
* Write a trivial module called foreman_reports into the first env, with just a lib dir
* Don't configure reports and document how to set it up (it's trivial after all)

In the meantime, this PR will make it easy to enable reporting on Puppet 4. Thoughts?

### DIFF
--- a/lib/puppet/reports/foreman.rb
+++ b/lib/puppet/reports/foreman.rb
@@ -1,0 +1,187 @@
+# copy this file to your report dir - e.g. /usr/lib/ruby/1.8/puppet/reports/
+# add this report in your puppetmaster reports - e.g, in your puppet.conf add:
+# reports=log, foreman # (or any other reports you want)
+# configuration is in /etc/puppet/foreman.yaml
+
+require 'puppet'
+require 'net/http'
+require 'net/https'
+require 'rbconfig'
+require 'uri'
+require 'yaml'
+begin
+  require 'json'
+rescue LoadError
+  # Debian packaging guidelines state to avoid needing rubygems, so
+  # we only try to load it if the first require fails (for RPMs)
+  begin
+    require 'rubygems' rescue nil
+    require 'json'
+  rescue LoadError => e
+    puts "You need the `json` gem to use the Foreman ENC script"
+    # code 1 is already used below
+    exit 2
+  end
+end
+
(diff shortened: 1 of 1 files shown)


GregSutcliffe:
No idea what's going with the tests there. Will leave it until we've had a chance to discuss the way forward.

mmoll:
@GregSutcliffe see #396 / #397

mmoll:
would it work to just leave it in files as-is and puppetize it as file resource into `/opt/puppetlabs/puppet/lib/ruby/vendor_ruby/puppet/reports`? we're handling the augeas trapperkeeper lens in puppet-puppet similarly.

domcleal:
Yeah, that would probably work.

It would be nice to put the report processor into a module one day, since it'd also make it easy to install onto an existing Puppet master (you'd just `puppet module install theforeman/foreman_reporter` or something, but it probably isn't a hard requirement for Puppet 4 support.

mmoll:
TBH, as already stated in #398 I'd rather move things like node.rb, foreman.rb and so on to puppet-puppet or at least use puppet-puppet for some variables so we don't have to reimplement the whole OS and AIO/non-AIO logic in puppet-foreman again.

GregSutcliffe:
Moving the YAML config file makes sense, but node.rb and foreman.rb are static files now, so putting them in their own module for easy module install makes sense to me. For now, I can code up a patch for @mmoll's suggestion over the weekend.

mmoll:
the static files also could go into some OS package (foreman-proxy or a seperate puppet-foreman or similar), I think that was the original plan when the yaml paremeters were rafctored out :)

ekohl:
I can confirm placing it in ```/opt/puppetlabs/puppet/lib/ruby/vendor_ruby/puppet/reports/foreman.rb``` works. I'll come up with a patch in puppet-puppet to place it in the correct directory.

mmoll:
fixed by theforeman/puppet-puppet#353

PR #399, opened 2016-02-11

igalic:
appear to have different levels of tolerance for such a certificate

this is a rebase of #337 against the stable-5.0 branch


mmoll:
@igalic this is looking good. However, please direct PRs into the master branch, thanks!

igalic:
@mmoll you think chances of travis being happy with this will be better when i base it off master?

mmoll:
@igalic Probably. I think 5.0 still has the monkey-patching against older rspec-puppet versions in (you'll run into #396 for sure, but that's known). Anyway, we always push to master first and would only backport if really needed.

mmoll:
see #401

PR #400, opened 2016-02-13
cherry-pick test fixes to 5.0-stable

mmoll:
Unsure if we need that and it's still broken for Puppet 2.7, but I think this is better than before.


domcleal:
Thanks @mmoll!

PR #402, opened 2016-02-17
Update min Puppet version and supported OSes

domcleal:


ekohl:
:+1: yay for dropping F19.

mmoll:
cherry-picked as 403152b with Ubuntu/xenial added, thanks @domcleal!

Issue #327, opened 2015-06-03
Installation of foreman-cli fails on CentOS 7


mattiasgiese:
I have tried to install foreman-proxy/foreman-cli using the foreman-installer and ran into this issue:

```
[...]
[ INFO 2015-06-03 14:11:54 main]            Requires: ruby193-ruby(abi) = 1.9.1
[ERROR 2015-06-03 14:11:54 main]  Package: ruby193-ruby-wrapper-0.0.2-5.el7.noarch (foreman)
[ INFO 2015-06-03 14:11:54 main]            Requires: ruby193-ruby
[ERROR 2015-06-03 14:11:54 main]  Package: ruby193-rubygem-foreigner-1.7.1-1.el7.noarch (foreman)
[ INFO 2015-06-03 14:11:54 main]            Requires: ruby193-rubygem(activerecord) >= 3.0.0
[ERROR 2015-06-03 14:11:54 main]  Package: foreman-1.8.1-1.el7.noarch (foreman)
[ INFO 2015-06-03 14:11:54 main]            Requires: ruby193-rubygem(json) < 2.0
[ERROR 2015-06-03 14:11:54 main]  Package: ruby193-rubygem-foreigner-1.7.1-1.el7.noarch (foreman)
[ INFO 2015-06-03 14:11:54 main]            Requires: ruby193-rubygems
[ INFO 2015-06-03 14:11:54 main]  You could try using --skip-broken to work around the problem
[ INFO 2015-06-03 14:11:54 main]  You could try running: rpm -Va --nofiles --nodigest
[...]
```

So, it is necessary to enable the ruby193 SCL repository on CentOS 7. 
The following quick hack fixes it:

```
diff --git a/manifests/cli.pp b/manifests/cli.pp
index 54fb998..f22e75e 100644
--- a/manifests/cli.pp
+++ b/manifests/cli.pp
@@ -36,6 +36,13 @@ class foreman::cli (
   validate_string($foreman_url_real, $username_real, $password_real)
   validate_bool($manage_root_config, $refresh_cache)

+  if $::operatingsystem == 'CentOS' and $::operatingsystemmajrelease == '7' {
+    package{'rhscl-ruby193-epel-7-x86_64':
+      ensure => installed,
+      before => Package['foreman-cli'],
+    }
+  }
+
   package { 'foreman-cli':
     ensure => installed,
   } ->
```

[...]

PR #404, opened 2016-02-23
refactor out a top repo class

mmoll:
first step for #327

This is not 100% ideal, i.e. as there's already the repo paramter in init/params, so this is mainly for comments now.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,10 @@
 # Changelog
 
+## 5.2.0
+* Compatibility warnings:
+    * `foreman::install::repos` has been moved to `foreman::repos`.
+      The old define has been deprecated and will issue a warning.
+
 ## 5.1.0
 * New classes to install Foreman plugins:
     * foreman::plugin::ansible to install Ansible support
--- a/manifests/repo.pp
+++ b/manifests/repo.pp
@@ -1,0 +1,20 @@
+# Configure the foreman repo
+class foreman::repo(
+  $custom_repo = $::foreman::custom_repo,
+  $repo        = $::foreman::repo,
+  $gpgcheck    = $::foreman::gpgcheck
+) {
+  anchor { 'foreman::repo::begin': }
+
+  if ! $custom_repo {
+    Anchor['foreman::repo::begin'] ->
+    foreman::repos { 'foreman':
+      repo     => $repo,
+      gpgcheck => $gpgcheck,
+    } -> Class['::foreman::repos::extra']
+  }
+
+  Anchor['foreman::repo::begin'] ->
+  class { '::foreman::repos::extra': } ->
+  anchor { 'foreman::repo::end': }
+}
--- a/manifests/repos.pp
+++ b/manifests/repos.pp
@@ -1,0 +1,39 @@
+# Set up a repository for foreman
+define foreman::repos(
+  $repo = stable,
+  $gpgcheck = true
+) {
+  include ::foreman::params
+
+  case $::osfamily {
+    'RedHat': {
+      foreman::repos::yum {$name:
+        repo     => $repo,
+        yumcode  => $::foreman::params::yumcode,
+        gpgcheck => $gpgcheck,
+      }
+    }
+    'Debian': {
+      foreman::repos::apt {$name:
+        repo => $repo,
+      }
+    }
+    'Linux': {
+      case $::operatingsystem {
+        'Amazon': {
+          foreman::repos::yum {$name:
+            repo     => $repo,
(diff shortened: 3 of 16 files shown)


manifests/repo.pp
```diff
@@ -0,0 +1,12 @@
+# Configure the foreman repo
+class foreman::repo(
+) {
+  if ! $::foreman::custom_repo {
```
ekohl: I think these should all be class parameters (defaulting to ```$::foreman::*```) so it can be used without the main foreman class, even if you have to manually set all parameters.

manifests/repo.pp
```diff
@@ -0,0 +1,12 @@
...
+    foreman::repos { 'foreman':
+      repo     => $::foreman::repo,
+      gpgcheck => $::foreman::gpgcheck,
+    }
+  }
+
+  class { '::foreman::repos::extra': }
```
ekohl: Include is nicer if you're not passing parameters.

ekohl: Maybe it is good to pass all parameters for the foreman-cli use case, but I'm leaning to pulling in ```::foreman::repos::extra``` in here.

manifests/repo.pp
```diff
@@ -0,0 +1,15 @@
...
+  $gpgcheck    = $::foreman::gpgcheck
+) {
+  if ! $custom_repo {
+    foreman::repos { 'foreman':
+      repo     => $repo,
+      gpgcheck => $gpgcheck,
+    }
+  }
+
+  include ::foreman::repos::extra
```
domcleal: I don't think this will get contained properly, you might need to add an anchor, use the contain keyword (and depend on Puppet >= 3.4(?)) or try `-> Class['foreman::repo']`.

domcleal:
Looks good, but I'm rather concerned about the ordering (one comment inline).  Some more `that_requires` tests would be good to prove it's all going to happen before foreman::install.

mmoll:
it's a bit of a kludge, but should work

manifests/repo.pp
```diff
@@ -0,0 +1,17 @@
...
+) {
+  if ! $custom_repo {
+    foreman::repos { 'foreman':
+      repo     => $repo,
+      gpgcheck => $gpgcheck,
+    }
+  }
+
+  class { '::foreman::repos::extra':
+    require => Class['foreman::repo'],
```
domcleal: Is that required?

domcleal:
I don't think it's fixed, it looks like extra *requires* repo, but won't be contained to run before foreman::install.

mmoll:
next try... however, I couldn't come up with a working test...

manifests/repo.pp
```diff
@@ -0,0 +1,20 @@
...
+
+  if ! $custom_repo {
+    Anchor['foreman::repo::begin'] ->
+    foreman::repos { 'foreman':
+      repo     => $repo,
+      gpgcheck => $gpgcheck,
+    } -> Class['::foreman::repos::extra']
+  }
+
+  class { '::foreman::repos::extra': } ->
```
domcleal: Nitpick, this doesn't depend on the begin anchor.

manifests/repos.pp
```diff
@@ -1,5 +1,5 @@
 # Set up a repository for foreman
-define foreman::install::repos(
```
domcleal: This is going to cause a problem in foreman_proxy, as it's calling this class directly: https://github.com/theforeman/puppet-foreman_proxy/blob/2.5.0/manifests/install.pp#L4-L7

Should we avoid renaming it?

ekohl: We could introduce a (deprecated) wrapper for compatibility.

domcleal: That'd work fine I think.  We would just issue a new release of foreman_proxy that depends on the newer foreman module and get rid of it here in the next major.

mmoll:
updated

domcleal:
:+1: tests well too.  @ekohl OK with you?

mmoll:
added version bump to metadata.json and changelog

CHANGELOG.md
```diff
@@ -1,5 +1,9 @@
 # Changelog
 
+## 5.2.0
+* Compatibility warnings:
+    * `foreman::install::repos` has been moved to `foreman::repos`
```
ekohl: 'The old define has been deprecated and will issue a warning.'

ekohl:
:+1: but I think the changelog can be tweaked. Fine either way.

mmoll:
added the comment to the changelog :bowtie:

domcleal:
Merged, thanks @mmoll!

PR #405, opened 2016-02-28
fixes #8489 - set KeepAlive On in foreman vhost

mmoll:
As it seems to be still a longer time to puppetlabs-apache 2.0, handle this in the vhost.
@lzap for your testing pleasure... :snail:

### DIFF
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -132,6 +132,17 @@
 #
 # $server_ssl_crl::             Defines the Apache mod_ssl SSLCARevocationFile setting in Foreman vhost conf file.
 #
+# $keepalive::                  Enable KeepAlive setting of Apache?
+#                               type:boolean
+#
+# $max_keepalive_requests::     MaxKeepAliveRequests setting of Apache
+#                               (Number of requests allowed on a persistent connection)
+#                               type:integer
+#
+# $keepalive_timeout::          KeepAliveTimeout setting of Apache
+#                               (Seconds the server will wait for subsequent requests on a persistent connection)
+#                               type:integer
+#
 # $oauth_active::               Enable OAuth authentication for REST API
 #                               type:boolean
 #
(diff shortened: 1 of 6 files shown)


lzap:
LGTM but I don't have enough confidence to merge Puppet stuff. Also, I think we should parametrize MaxKeepAliveRequests and KeepAliveTimeout options.

mmoll:
@lzap like that?
tests are still missing...

lzap:
That looks good, you can also parametrize keep-alive itself.

mmoll:
@lzap done :gift:
Now also with a test.

manifests/init.pp
```diff
@@ -246,6 +257,9 @@
   $server_ssl_certs_dir      = $::foreman::params::server_ssl_certs_dir,
   $server_ssl_key            = $::foreman::params::server_ssl_key,
   $server_ssl_crl            = $::foreman::params::server_ssl_crl,
+  $keepalive                 = $::foreman::params::keepalive,
```
ekohl: Call validate_*() on these parameters?

ekohl:
:+1: when the validation functions are used.

manifests/init.pp
mmoll: I do in manifests/config/passenger.pp

ekohl: Touche

manifests/init.pp
```diff
@@ -132,6 +132,17 @@
...
 # $server_ssl_crl::             Defines the Apache mod_ssl SSLCARevocationFile setting in Foreman vhost conf file.
 #
+# $keepalive::                  Enable KeepAlive setting of Apache?
+#                               type:boolean
+#
+# $maxkeepaliverequests::       MaxKeepAliveRequests setting of Apache
+#                               (Number of requests allowed on a persistent connection)
+#                               type:integer
+#
+# $keepalivetimeout::           KeepAliveTimeout setting of Apache
```
domcleal: Should these parameters contain underscores separating the words?  Other parameters do, e.g. `passenger_start_timeout`

mmoll: changed

domcleal:
:+1: when tests pass.  I tested the earlier version (prior to the parameter rename) successfully.

domcleal:
Merged, thanks @mmoll.

PR #408, opened 2016-03-10
fixes #14127 - turn on httpd server-status scoreboard

komidore64:


ekohl:
I don't think we want to force this on all our users. I can see this being useful for many of them, but I think it belongs in the profile layer. Not sure where to place that with the installer though.

komidore64:
What do you mean by "profile layer?"

domcleal:
A higher level class, i.e. http://www.craigdunn.org/2012/05/239/.

ekohl:
Given we don't want to manage Apache in our modules I'm closing this. A possible solution for this might be https://github.com/theforeman/foreman-installer/pull/204

PR #410, opened 2016-03-14
Adds timestamp to postgres logs

xprazak2:
Based on a redmine [issue](http://projects.theforeman.org/issues/5818) in installer

### DIFF
--- a/manifests/database/postgresql.pp
+++ b/manifests/database/postgresql.pp
@@ -25,5 +25,9 @@ class foreman::database::postgresql {
     locale   => 'en_US.utf8',
   }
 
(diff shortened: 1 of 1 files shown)


domcleal:
A bit of the same comment as #408, that this might be better in a profile layer as it could conflict with a user's own configuration or another module since it's server-wide.  Else perhaps submit this to the postgresql module project and have it parameterised with this as a default value?

xprazak2:
I submitted this to [puppetlabs-postgresql module](https://github.com/puppetlabs/puppetlabs-postgresql/pull/754) and it got merged today, therefore I am closing this.

domcleal:
Thanks!

PR #412, opened 2016-03-15
add 'puppetmaster' fact value

aleksey-novikov:
This fact allow to determine which puppet master is used by agent when using use_srv_records and srv_domain puppet settings. Furthermore it allow analize proxy load by adding this fact to trend counters.

### DIFF
--- a/files/external_node_v2.rb
+++ b/files/external_node_v2.rb
@@ -112,6 +112,13 @@ def build_body(certname,filename)
   puppet_facts = YAML::load(facts.gsub(/\!ruby\/object.*$/,''))
   hostname     = puppet_facts['values']['fqdn'] || certname
   
+  begin
+    require 'facter'
+    puppet_facts['values']['puppetmaster_fqdn'] = Facter.value(:fqdn).to_s
+  rescue LoadError => e
+    puppet_facts['values']['puppetmaster_fqdn'] = `hostname -f`.strip
+  end
+  
   # filter any non-printable char from the value, if it is a String
   puppet_facts['values'].each do |key, val|
     if val.is_a? String


discreet:
@aleksey-novikov Where does this Fact get executed? Is it on the Puppet master itself as part of the Puppet Smart-Proxy? I have a concern with the fact that it is using the FQDN which can be inaccurate when operating in a clustered environment. The Foreman parameter for Puppet Master would be more accurate in this sense.

brandonweeks:
I think this is a valuable addition for clustered environments, however a different name might be ideal. The "puppet master" nomenclature seems to be getting phased out with puppetserver becoming the default. Overall :+1:

discreet:
I agree with @brandonweeks, this could be very valuable when troubleshooting however it is the name that is confusing. I think there should be some type of distinction made between individual server responding vs the overall cluster.

aleksey-novikov:
@discreet, @brandonweeks I do not insist on this name. if you have any ideas how it is better to name this fact, I ready to rename it.

files/external_node_v2.rb
```diff
@@ -112,6 +112,9 @@ def build_body(certname,filename)
   puppet_facts = YAML::load(facts.gsub(/\!ruby\/object.*$/,''))
   hostname     = puppet_facts['values']['fqdn'] || certname
   
+  require 'facter'
```
domcleal: It's possible that Facter can't be loaded in this script as it depends on which Ruby is being used and how the master is installed - e.g. in a PE or AIO environment, this script might be running under the system Ruby, which may not have Facter installed.

I'd consider using a `begin`/`rescue LoadError` block and skipping this step if it can't be loaded.  Perhaps even use `hostname` as a fallback..

domcleal:
Perhaps `puppetmaster_fqdn`?

aleksey-novikov:
@domcleal Facter already has been used in this script in 'thread_count' function. I have nothing against `puppetmaster_fqdn`, but it can offer the other participants in the conversation other options.

domcleal:
> Facter already has been used in this script in thread_count function

Yes, but the thread_count function isn't called in ordinary usage - only when using the watch functionality.  This PR adds the facter require into the main code path for every fact upload, so it's far more likely to cause a problem.

aleksey-novikov:
@domcleal That is, you are encouraged to use \`hostname -f\`.strip instead of facter?

domcleal:
Yes, certainly as a fallback if Facter can't be loaded.

aleksey-novikov:
@domcleal Ok I'll rewrite it. But I don't have neither PE nor AIO environments, so I need help to test changes

aleksey-novikov:
@domcleal I rewrite code and rename fact. Review changes please.

domcleal:
Looks fine to me I think.

brandonweeks:
:+1:

discreet:
Looks good. :+1:

mmoll:
merged, Спасибо @aleksey-novikov!

aleksey-novikov:
@mmoll :+1:

PR #414, opened 2016-03-15
use different ssl dir on AIO

mmoll:
see https://docs.puppetlabs.com/puppet/4.3/reference/whered_it_go.html#ssldir-is-inside-confdir

### DIFF
--- a/manifests/params.pp
+++ b/manifests/params.pp
@@ -215,19 +215,32 @@ class foreman::params {
   $puppet_user = 'puppet'
   $puppet_group = 'puppet'
 
+  if versioncmp($::puppetversion, '4.0') < 0 {
+    $aio_package = false
+  } elsif $::osfamily == 'Windows' or $::rubysitedir =~ /\/opt\/puppetlabs\/puppet/ {
+    $aio_package = true
+  } else {
+    $aio_package = false
+  }
+
+  $puppet_ssldir = $aio_package ? {
+    true    => '/etc/puppetlabs/puppet/ssl',
+    default => "${puppet_home}/ssl"
+  }
+
   # If CA is specified, remote Foreman host will be verified in reports/ENC scripts
-  $client_ssl_ca   = "${puppet_home}/ssl/certs/ca.pem"
+  $client_ssl_ca   = "${puppet_ssldir}/certs/ca.pem"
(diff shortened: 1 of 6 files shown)


mmoll:
looking for comments on that one...

The `puppetssldir` setting is used to derive other paths in the Foreman settings, the symptom I saw was that Foreman could not load the certificate it uses to communicate with the proxy. I have no good ideas how to handle that...

manifests/params.pp
```diff
@@ -215,19 +215,32 @@
...
+  if versioncmp($::puppetversion, '4.0') < 0 {
+    $aio_package = false
+  } elsif $::osfamily == 'Windows' or $::rubysitedir =~ /\/opt\/puppetlabs\/puppet/ {
+    $aio_package = true
+  } else {
+    $aio_package = false
+  }
+
+  $puppet_ssldir = $aio_package ? {
+    true    => '/etc/puppetlabs/puppet/ssl',
```
ekohl: What happens on Windows here?

mmoll: Windows is ruled out by running into line 212 anyway, I just copypasted line 218-224 from puppet-puppet.

ekohl: You're right, we don't fully support that anyway. I was thinking of settings.yaml

mmoll:
yeah, I guess it would work (and I would also apply the usual logic with pulling `$puppet_ssldir` into init.pp to make it overridable), and it does not feel correct with current master anyway, as the Foreman settings will also get not updated when one overrides the respective parameters here...

mmoll:
updated and added a test

mmoll:
@brandonweeks BTW, if you want to test... :point_up:

mmoll:
rebased.

ekohl:
Untested, but in general :+1:. We need a similar fix in puppet-foreman_proxy.

mmoll:
-proxy worked for me as-is, because `$::puppet::params::ssldir` is used there, but maybe that only worked for me because I did an all-in-one setup.

brandonweeks:
Whoops, I thought I commented on this yesterday. Tested :+1: in production. :)

domcleal:
Merged, thanks @mmoll!

PR #416, opened 2016-03-22
Order foreman::cli after repo setup

domcleal:
1b552f0 seems to have triggered this condition in our nightly tests, where adding both foreman and foreman::cli classes causes the foreman-cli package to be installed before repos are configured.

I don't think any dependency has ever been present, and it's because foreman-cli now depends on SCL packages on EL OSes that it's especially relevant now.

### DIFF
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -332,6 +332,10 @@ class foreman (
   Package <| tag == 'foreman-compute' |> ~>
   Class['foreman::service']
 
+  Class['foreman::repo'] ~>
+  Package <| tag == 'foreman::cli' |> ~>
+  Class['foreman']
+
   # lint:ignore:spaceship_operator_without_tag
   Class['foreman::database']~>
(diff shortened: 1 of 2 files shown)


ekohl:
I'm wondering if this can't be solved inside ```foreman::cli```.

domcleal:
> I'm wondering if this can't be solved inside foreman::cli.

It's tricky to add a hard relationship I think, because foreman::cli can be used without the foreman class.  If using `defined()` to only add it when the other class exists would still require code in both classes as they may be defined in either order.

Using a collector to make it optional like I've done here wouldn't work the other way around (e.g. `Class <| name == 'foreman::repo' |> ~> Class['foreman::cli']`), since you can't collect on a class.  It's easier to specify it this way as there's only one package resource, but collecting on the contents of foreman::repo is hard with many yumrepo/package/file etc resources.

mmoll:
at least doesn't break anything for me, merged, thanks @domcleal!

PR #418, opened 2016-03-27
Allow user to select vhost priority

lazyfrosch:
Please let me know if you would accept this. I'd add a spec test if you like.

### DIFF
--- a/manifests/config/passenger.pp
+++ b/manifests/config/passenger.pp
@@ -9,6 +9,8 @@
 #
 # $ruby::                     Path to Ruby interpreter
 #
+# $priority::                 Apache vhost priority
+#
 # $servername::               Servername for the vhost.
 #
 # $serveraliases::            Serveraliases for the vhost.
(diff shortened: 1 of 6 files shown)


mmoll:
could you describe your use case?

lazyfrosch:
The idea is to have a server, that is not purely dedicated to Puppet or Foreman, so I still want a properly configured default vhost for the $::fqdn, while Foreman resides on a special servername like `foreman.example.com`.

The Puppet vhost doesn't matter, since its put on a extra port.

I just missed the flexibility here, and had to hand fiddle with the default vhosts.

mmoll:
Ah, mainly http, I'd suggerst to name the parameter `vhost_priority`, no objections otherwise.

lazyfrosch:
Yeah that makes sense, I updated my commit.

lazyfrosch:
Uhm the errornous spec tests dont seem related to my change.

ekohl:
It looks like you're right: https://travis-ci.org/theforeman/puppet-foreman/jobs/119093494 fails with the same error.

manifests/config/passenger.pp
```diff
@@ -55,6 +57,7 @@
   $app_root               = $::foreman::app_root,
   $listen_on_interface    = $::foreman::passenger_interface,
   $ruby                   = $::foreman::passenger_ruby,
+  $priority               = $::foreman::vhost_priority,
```
ekohl: Could you add a ```validate_string``` for this?

mmoll:
please rebase onto current master to get fixed tests.

lazyfrosch:
Rebased and with validate_string

ekohl:
:+1: on green tests.

mmoll:
merged, Danke @lazyfrosch!

PR #419, opened 2016-03-29
Fix function spec failures under rspec-puppet 2.4.0

domcleal:
Specs are rewritten to use the documented function testing helpers,
which fixes the errors seen under Puppet 4 with 2.4.0 due to changes
in initialisation.

    Puppet::DevError:
    Function ensure_value_in_string not defined despite being loaded!


ekohl:
:+1: on green tests

mmoll:
merged, thanks @domcleal!

PR #421, opened 2016-03-31
Fix foreman_smartproxy provider idempotency

alexjfisher:
Previously, if ::foreman_proxy::registered_name contained spaces, the
provider would not find the already registered proxy when searching
using the API.

This would result in errors like...

```
Error: /Stage[main]/Foreman_proxy::Register/Foreman_smartproxy[Puppet on
proxy.example.com]/ensure: change from absent to present failed: Proxy
Puppet on proxy.example.com cannot be registered (422 Unprocessable
Entity): Name has already been taken URL Only one declaration of a proxy
is allowed
```


domcleal:
:+1:

ekohl:
Thanks!

PR #422, opened 2016-03-31
Change openstack/rackspace compute resource packages

domcleal:
foreman-openstack/rackspace subpackages are provided on Foreman 1.12,
per ticket #14368. Compatibility with 1.11 and older is provided through
the $package parameter.


mmoll:
Adding an entry to the README would be nice, :+1: anyway

domcleal:
Good idea, updated.

mmoll:
merged, thanks @domcleal!

PR #423, opened 2016-03-31
allow plugins to be updated

cristifalcas:
This will allow the plugins to be kept updated.

In case that foreman is set to a fixed version, the version of plugins will need to be set also for all plugins.


mmoll:
https://github.com/theforeman/puppet-foreman_proxy/pull/218#issuecomment-179231342 applies also here I think.

mmoll:
:+1: on green tests

mmoll:
merged, thanks @cristifalcas!

PR #424, opened 2016-04-04
Add foreman::providers to install provider dependencies

domcleal:
This will replace the package resource in foreman_proxy and help when
using the providers here outside of foreman_proxy.

### DIFF
--- a/manifests/providers.pp
+++ b/manifests/providers.pp
@@ -1,0 +1,41 @@
+# = foreman_* providers support
+#
+# Installs dependencies to use the foreman_* types and providers.
+#
+# Default parameters should point to the latest packages required for
+# the current version of the providers.
+#
+# === Parameters:
+#
+# $apipie_bindings::          Install apipie-bindings dependency
+#                             type:boolean
+#
+# $apipie_bindings_package::  Name of apipie-bindings package
+#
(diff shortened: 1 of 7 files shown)


manifests/providers.pp
```diff
@@ -0,0 +1,38 @@
...
+#
+# $apipie_bindings_package::  Name of apipie-bindings package
+#
+# $foreman_api::              Install foreman_api dependency (deprecated)
+#                             type:boolean
+#
+# $foreman_api_package::      Name of foreman_api package
+#
+class foreman::providers(
+  $apipie_bindings         = $::foreman::providers::params::apipie_bindings,
```
ekohl: Validate the parameters using ```validate_bool``` and ```validate_string```?

domcleal: Thanks @ekohl, updated with the validation functions.

mmoll:
Maybe the old foreman_api provider can get removed and the modules major bumped? I'm also OK with it being kept for now...
:+1: otherwise

domcleal:
I'm not sure it's worth doing a major bump over... so I've added some docs to the README about the types/providers (long overdue) and noted there and in the provider header that it's deprecated so we can easily get rid of it later.

mmoll:
merged, thanks @domcleal!

PR #426, opened 2016-04-07
Fixes #14520 - upstream repo should be explicitly enabled

ares:


domcleal:
Default behaviour of the module throughout is to enable repos that are required to install Foreman, its plugins and dependencies, so this should remain true.

It could now be disabled on EL7 as the necessary dependency is in 7.2 (IIRC), but needs to be enabled for EL6 still.

ares:
@domcleal I think it's similar to [downloading discovery images](https://github.com/theforeman/puppet-foreman/blob/master/manifests/plugin/discovery/params.pp#L5)...

Anyway after digging a bit deeper, this repo is not require since foreman_openscap 0.5* does not use openscap. It relies on smart proxy to do it. It's not dependency to install foreman_openscap package, at least I don't see it [being dependency](https://github.com/theforeman/foreman-packaging/blob/rpm/1.11/rubygem-foreman_openscap/rubygem-foreman_openscap.spec) so we can remove this completely.

I'll move it to smart_proxy_openscap plugin manifest since it's needed there now. Should I write a migration to remove this parameter from answer file or do we want to keep manifest compatible with just no-oping for some time?

domcleal:
> I think it's similar to downloading discovery images...

I don't think so, installing discovery images is a very expensive process.  It's a package repository so should be treated like the other repositories.

> I'll move it to smart_proxy_openscap plugin manifest since it's needed there now.

Yep, thanks, with the default enabled for EL6.

> Should I write a migration to remove this parameter from answer file or do we want to keep manifest compatible with just no-oping for some time?

Removing the parameter would require a major version bump, so it should stay for now with this implementation too (not no-op).  Once added to foreman_proxy, this PR can be merged to change the default to false.

You could optionally add a migration to change the existing value in answer files to false, else wait until it's removed completely and delete it then.

ares:
> else wait until it's removed completely and delete it then.

should we put it on some list to remove it on next major bump so it won't last there forever?

> Yep, thanks, with the default enabled for EL6.

I opened https://github.com/theforeman/puppet-foreman_proxy/pull/228 that should address it on smart proxy side.

domcleal:
> should we put it on some list to remove it on next major bump so it won't last there forever?

Yeah, we're not good at tracking this.  Perhaps we should add some GitHub issues and milestones for the next major version.

ares:
>  Perhaps we should add some GitHub issues and milestones for the next major version.

Feel free to adjust name with [DNM] https://github.com/theforeman/puppet-foreman/pull/427 or whatever you prefer

domcleal:
Thanks.  

:+1: to this PR subject to tests.

ekohl:
Looks like the tests need to be updated.

ares:
@ekohl I don't think so, the parameter is still there and fully functional, it's just default value that is being changed

ekohl:
I think the tests failed and I didn't check why, sorry. Thanks!

Issue #428, opened 2016-04-12
[question] APT repo handling


lazyfrosch:
Is there a reason for us not using puppetlabs-apt in the module here?

I noticed it while APT notified me about an unknown key.

We should use ::apt there, and add the key id of the current key.

Whats your opinion? I could do a PR...

domcleal:
I can't think of a reason, I don't see why it couldn't be refactored.

mmoll:
:+1: for using pl-apt also here.

lazyfrosch:
How about updating pl-apt to 2.2.x ?

domcleal:
The metadata permits 2.2.x already, so probably fine to update the minimum version too if that's required.

lazyfrosch:
Please continue discussion in #429

PR #429, opened 2016-04-12
Updating repos::apt to puppetlabs-apt standards

lazyfrosch:
This will also require puppetlabs-apt > 2.0.0 because the provider has recommended new parameter style.

fixes GH-428

Things to discuss:
- [x] puppetlabs-apt update to > 2.0.0
- [x] key retrieval via keyserver by default? (key id is and should be mandatory with the current key)

Tests:
- [x] Debian jessie (Puppet Master + Foreman + Proxy)
- [x] Debian jessie (Proxy only)

### DIFF
--- a/metadata.json
+++ b/metadata.json
@@ -26,7 +26,7 @@
     },
     {
       "name": "puppetlabs/apt",
-      "version_requirement": "< 3.0.0"
+      "version_requirement": ">= 2.0.0 < 3.0.0"
     },
     {
       "name": "puppetlabs/concat",
--- a/.fixtures.yml
+++ b/.fixtures.yml
@@ -2,9 +2,7 @@ fixtures:
   repositories:
     alternatives:     'git://github.com/adrienthebo/puppet-alternatives'
     apache:           'git://github.com/puppetlabs/puppetlabs-apache'
-    apt:
-      repo:           'git://github.com/puppetlabs/puppetlabs-apt'
-      branch:         '1.8.x'
+    apt:              'git://github.com/puppetlabs/puppetlabs-apt'
     concat:           'git://github.com/puppetlabs/puppetlabs-concat'
     mysql:            'git://github.com/puppetlabs/puppetlabs-mysql'
     postgresql:       'git://github.com/puppetlabs/puppetlabs-postgresql'
--- a/manifests/repos/apt.pp
+++ b/manifests/repos/apt.pp
@@ -1,14 +1,26 @@
 # Install an apt repo
 define foreman::repos::apt ($repo) {
-  file { "/etc/apt/sources.list.d/${name}.list":
-    content => "deb http://deb.theforeman.org/ ${::lsbdistcodename} ${repo}\ndeb http://deb.theforeman.org/ plugins ${repo}\n",
-  } ->
-  exec { "foreman-key-${name}":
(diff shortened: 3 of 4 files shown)


lazyfrosch:
Change for discussion, especially since I'd like to update to puppetlabs-apt > 2.0.0

lazyfrosch:
(Live Server Tests still pending)

lazyfrosch:
Its maybe a bit ugly to have basically 4 repos on a Puppet Master + Foreman + Proxy, but working:

```
for i in foreman*; do echo $i; cat $i; done
foreman.list
# This file is managed by Puppet. DO NOT EDIT.
# foreman
deb http://deb.theforeman.org/ jessie stable
foreman-plugins.list
# This file is managed by Puppet. DO NOT EDIT.
# foreman-plugins
deb http://deb.theforeman.org/ plugins stable
foreman_proxy.list
# This file is managed by Puppet. DO NOT EDIT.
# foreman_proxy
deb http://deb.theforeman.org/ jessie stable
foreman_proxy-plugins.list
# This file is managed by Puppet. DO NOT EDIT.
# foreman_proxy-plugins
deb http://deb.theforeman.org/ plugins stable
```

Tested on my machines.

domcleal:
https://github.com/theforeman/puppet-foreman_proxy/pull/227 will fix the duplicate repos.

lazyfrosch:
Updated the Spec tests for 1.8.7

Please review :)

mmoll:
I think this is looking good in general. Regarding the key retrival, if we can make sure new keys are uploaded to the respective keyservers I think this is fine. If we should get a https URL to fetch the key from I'd slightly prefer that, but have no strong feelings about it.

lazyfrosch:
@mmoll technically we could use a key stored to the repo as well, but we should make sure not to just overwrite the key, but add like `foreman-2016.asc`.

Since it will matter it to match with the key id.

metadata.json
```diff
@@ -26,7 +26,7 @@
     },
     {
       "name": "puppetlabs/apt",
-      "version_requirement": "< 3.0.0"
+      "version_requirement": "> 2.0.0 < 3.0.0"
```
domcleal: Should this be `>= 2.0.0` rather than `>`?

lazyfrosch: Yes!

domcleal:
I'd marginally prefer putting the key in this repo as described above over using a keyserver, but the latter should always be available.  Otherwise :+1:.

lazyfrosch:
Well, if a used wants to use the keyserver, he should toggle custom_repo and make his own definition. This also counts probably for local repos and local available files.

@domcleal could you provide the asc on the repo and I'll change it here. I'd suggest `foreman-<shortid>.asc` (so its not depending on the year!)

domcleal:
Ah, I just meant that the keys we use should always be on the keyserver.

http://theforeman.org/security.html#GPGkeys has links to the full public keys that you can copy.

lazyfrosch:
So key retrieval via HTTP from keyserver it is?

lazyfrosch:
Updated, please have a look on the spec for the `apt_key`

domcleal:
:+1:

mmoll:
Merged. Danke @lazyfrosch!

PR #430, opened 2016-04-12
Change foreman::repo inclusion to 'include' keyword

domcleal:
Allows the class to be included with class parameters from
puppet-foreman_proxy.


mmoll:
merged, thanks @domcleal!

PR #431, opened 2016-04-13
Refresh smart proxy only when currently registered

domcleal:
Prevents misleading error messages due to unknown smart proxy IDs.


mmoll:
merged, thanks @domcleal!

PR #432, opened 2016-04-15
fixes #14455 - add rest_v3 smart proxy provider using OAuth gem

domcleal:
The new provider has fewer dependencies than rest_v2 (with
apipie-bindings), therefore it's easier to package the dependencies for
an AIO version of Puppet. puppet-agent-oauth will be provided by the
Foreman repositories to install the oauth gem into the AIO environment.

All Ruby 1.9+ installations, including Puppet AIO packages, won't need
a JSON package as they include a bundled 'json' library.

The provider's been split into a base type+provider (supported as per
PUP-2458) to make it easier to add and update other types in future.

The addition of the SSL CA in the type is required with the OAuth gem's
HTTP request support, as it enables verification by default when it
detects CA bundles in common locations.

Thanks to @liamjbennett, whose work this is partly based on.

### DIFF
--- a/manifests/providers/params.pp
+++ b/manifests/providers/params.pp
@@ -1,26 +1,48 @@
 # foreman::providers default parameters
 class foreman::providers::params {
   # Dependency packages for different providers supplied in this module
-  $apipie_bindings = true
+  $oauth = true
+  $json = (versioncmp($::rubyversion, '1.9') < 0)
+  $apipie_bindings = false
   $foreman_api = false
 
   # OS specific package names
   case $::osfamily {
     'RedHat': {
+      if versioncmp($::puppetversion, '4.0') >= 0 {
+        $oauth_package = 'puppet-agent-oauth'
+      } else {
+        $oauth_package = 'rubygem-oauth'
+      }
+      $json_package = 'rubygem-json'
       $apipie_bindings_package = 'rubygem-apipie-bindings'
       $foreman_api_package = 'rubygem-foreman_api'
     }
     'Debian': {
+      if versioncmp($::puppetversion, '4.0') >= 0 {
+        $oauth_package = 'puppet-agent-oauth'
+      } else {
+        $oauth_package = 'ruby-oauth'
+      }
+      $json_package = 'ruby-json'
       $apipie_bindings_package = 'ruby-apipie-bindings'
       $foreman_api_package = 'ruby-foreman-api'
     }
     'FreeBSD': {
+      $oauth_package = 'rubygem-oauth'
+      $json_package = 'rubygem-json'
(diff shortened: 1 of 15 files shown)


manifests/providers/params.pp
```diff
@@ -1,26 +1,48 @@
...
+  $oauth = true
+  $json = (versioncmp($::rubyversion, '1.9') < 0)
+  $apipie_bindings = false
   $foreman_api = false
 
   # OS specific package names
   case $::osfamily {
     'RedHat': {
+      if versioncmp($::puppetversion, '4.0') >= 0 {
+        $oauth_package = 'puppet-agent-oauth'
```
domcleal: Please note that these AIO related package names depend on https://github.com/theforeman/foreman-packaging/pull/1095 and a forthcoming PR for the Debian package.

mmoll:
Doesn't seem to work for me on a fresh Puppet 4 install:
```
Error: /Stage[main]/Foreman_proxy::Register/Foreman_smartproxy[puppetserver.kvedulv.de]: Could not evaluate: Exception SSL_connect returned=1 errno=0 state=error: certificate verify failed in get request to: https://puppetserver.kvedulv.de/api/v2/smart_proxies?search=name=%22puppetserver.kvedulv.de%22
Wrapped exception:
SSL_connect returned=1 errno=0 state=error: certificate verify failed
``

**Comment by mmoll** [comment]
Also happens on an existing installation on Debian/jessie.

**Comment by domcleal** [comment]
@mmoll you need to add an `ssl_ca => pick($::foreman_proxy::foreman_ssl_ca, $::foreman_proxy::ssl_ca)` to the foreman_smartproxy resource over in foreman_proxy/manifests/register.pp, as the oauth module enables SSL verification by default.  I've got a branch for the change, I'll open a PR next week, or after this is merged.

**Comment by mmoll** [comment]
@domcleal indeed. :+1:

**Comment by domcleal** [comment]
https://github.com/theforeman/puppet-foreman_proxy/pull/232 adds the new parameter to the resource in foreman_proxy.

**Comment by mmoll** [comment]
merged, thanks @domcleal!

**Comment by beav** [comment]
There may be some issue with this patch on CentOS 6, this error appears when running bats tests

```
Duplicate declaration: Package[rubygem-json] is already declared in file /usr/share/foreman-installer/modules/foreman/manifests/providers.pp:52; cannot redeclare at /usr/share/foreman-installer/modules/foreman/manifests/puppetmaster.pp:28 on node centos6-bats.example.com
```

domcleal:
#434 fixes it.

beav:
@domcleal ah, very cool :snowboarder: . Thanks!

PR #433, opened 2016-04-18
Fixes #14687 - Passenger module no longer loaded twice

xprazak2:


manifests/config/passenger.pp
```diff
@@ -81,6 +81,16 @@
   include ::apache
   include ::apache::mod::headers
   include ::apache::mod::passenger
+
+  #make sure Passenger module not loaded twice on RHEL/CenotOS 6
```
mmoll: "CentOS". Also, a space after the # would be nice.

mmoll:
I'm a bit unsure about this, is there anything that can solve this in the apache module?

domcleal:
Indeed, this should be fixed in puppetlabs-apache - Passenger isn't managed by this module.

xprazak2:
You are right, I just realized one thing. The file is not managed by [puppetlabs-apache](https://github.com/puppetlabs/puppetlabs-apache/blob/master/manifests/mod/passenger.pp#L37-L42), so this should probably go to [foreman-packaging](https://github.com/theforeman/foreman-packaging/blob/rpm/develop/rubygem-passenger/apache-passenger.conf.in#L1)

domcleal:
Probably not, because the packages work out of the box.  If it was removed from the packages then they wouldn't work.

xprazak2:
@domcleal, you are right, the package would not work out of the box without it. I do not think this can be fixed in puppetlabs-apache since they add [zpassenger.load](https://github.com/puppetlabs/puppetlabs-apache/pull/1060#issue-60672795) on purpose to load passenger with lower priority.
I see 2 possible solutions (apart from this):

1) We add `ensure => absent` for zpassenger.load
2) We move loading module from passenger.conf to zpassenger.load in our packaging (not sure this would work)

Your thoughts?

PR #435, opened 2016-04-22
Enable foreman-release-scl on RHEL

domcleal:
Foreman 1.12 requires it for access to the CentOS SCLo repositories
containing sclo-ror42.

### DIFF
--- a/manifests/params.pp
+++ b/manifests/params.pp
@@ -35,8 +35,8 @@ class foreman::params {
 
   # Additional software repos
   $configure_epel_repo      = ($::osfamily == 'RedHat' and $::operatingsystem != 'Fedora')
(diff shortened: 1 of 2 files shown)


ekohl:
:+1: on green tests

mmoll:
merged, thanks @domcleal!

ares:
Does it mean I have to set `--foreman-configure-scl-repo false` explicitly when I want to use RHEL and RHSCL?

domcleal:
RHSCL doesn't ship SCLo (sclo-ror42) packages, so you should only disable it if you've configured the repo another way.

ares:
oh I see, thanks for clarification

PR #438, opened 2016-04-28
use Timeout.timeout in node.rb

mmoll:
timeout alone is deprecated in Ruby 2.3:

`/etc/puppet/node.rb:343:in <main>: Object#timeout is deprecated, use Timeout.timeout instead.`

### DIFF
--- a/files/external_node_v2.rb
+++ b/files/external_node_v2.rb
@@ -340,7 +340,7 @@ if __FILE__ == $0 then
       # query External node
       begin
         result = ""
-        timeout(tsecs) do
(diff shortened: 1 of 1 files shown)


brandonweeks:
👍

domcleal:
Could you add a `require 'timeout'` too please?

mmoll:
there's already a require in line 78, or is it needed in the function specifically?

domcleal:
Ah thanks, I missed that.

Merged, thanks @mmoll!

PR #439, opened 2016-04-29
Revert default rest_v3 request data to nil

domcleal:
Fixes refreshing of smart proxies against Foreman, since the PUT
request data defaulted to `{}`. This causes the oauth gem to form
encode the empty hash data, overriding the Content-Type to application/
x-www-form-urlencoded and the request is rejected by Foreman with a 415
HTTP response.

https://github.com/oauth-xx/oauth-ruby/blob/v0.5.1/lib/oauth/consumer.rb#L362-L364 checks for a Hash overrides the content type.


mmoll:
damn, I think I ran into this when testing #436 yesterday, but couldn't reproduce it...
:+1: on green tests.

mmoll:
merged, thanks @domcleal!

PR #440, opened 2016-05-06
fixes #14942 - require puppet::server::install first for puppet group

domcleal:


ekohl:
:+1: I assume this is needed for the SSL-certificates. In particular the websockets proxy for the console.

domcleal:
Yeah, the certs are also required but I think it works OK today. To ensure they're generated in time for Apache to load, the `puppet::server::config` class adds a relationship between the cert generate exec and `Service[httpd]`.

For Foreman's use of the certs in websockets and smart proxy comms, the app loads them lazily. It would use them for the smart proxy registration during the installer run, but we have various relationships in the main foreman/proxy classes to ensure the registration only happens after Foreman and the smart proxy are loaded anyway.

ekohl:
> Edit: I misunderstood. Yes, the 'puppet' group is required for access to the Puppet SSL certs for websockets and smart proxy comms.

Yes, I was a bit unclear but this is what I meant.

mmoll:
merged, thanks @domcleal!

PR #444, opened 2016-05-18
Support timeout configuration in foreman report processor

treydock:
I have run into cases where a catalog and subsequent reports are so large that the default 60 second timeout in Net::HTTP is too short.  The reports are still uploaded but errors are thrown in puppetserver.


ekohl:
:+1:

domcleal:
Thanks @treydock!

PR #446, opened 2016-05-21
improve AIO detection for oauth provider

mmoll:
this helps to get the correct package with non-AIO Puppet 4 packages.

### DIFF
--- a/manifests/providers/params.pp
+++ b/manifests/providers/params.pp
@@ -9,7 +9,7 @@ class foreman::providers::params {
   # OS specific package names
   case $::osfamily {
     'RedHat': {
-      if versioncmp($::puppetversion, '4.0') >= 0 {
+      if $::rubysitedir =~ /\/opt\/puppetlabs\/puppet/ {
         $oauth_package = 'puppet-agent-oauth'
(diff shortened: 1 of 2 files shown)


manifests/providers/params.pp
```diff
@@ -9,7 +9,7 @@
   # OS specific package names
   case $::osfamily {
     'RedHat': {
-      if versioncmp($::puppetversion, '4.0') >= 0 {
+      if versioncmp($::puppetversion, '4.0') >= 0 and $::rubysitedir =~ /\/opt\/puppetlabs\/puppet/ {
```
ekohl: Is this version check even needed?

mmoll: Not really, but I left it in.

ekohl: One reason to leave it out is to potentially support old PE installs, but since nobody has asked for that I think the only real reason is simplicity.

domcleal:
:+1: when rebased and passing tests.

mmoll:
:green_heart:

ekohl:
Thanks!

PR #447, opened 2016-05-22
fix configure_brightbox_repo in test

mmoll:


ekohl:
It appears this broke the tests. I do recall that some versions of the facts used 12.04 while others use 12.

mmoll:
jep, it's depended on the version of facterdb as it seems, will have a look...

mmoll:
simply checking for both strings now

spec/classes/foreman_install_spec.rb
```diff
@@ -35,7 +35,7 @@
             it { should contain_package('tfm-rubygem-passenger-native') }
           end
         when 'Debian'
-          configure_brightbox_repo = os == 'ubuntu-12-x86_64'
+          configure_brightbox_repo = os == 'ubuntu-12.04-x86_64' || os == 'ubuntu-12-x86_64'
```
ekohl: How about checking if the string starts with ```ubuntu-12```?

mmoll: that should also work :)

domcleal:
Merged, thanks @mmoll.

PR #452, opened 2016-06-07
remove formerly deprecated parameters and code

mmoll:
- remove rest(v1) smart proxy provider and support for foreman_api gem
- remove foreman::install::repos define
- remove apipie_task parameter
- mark Fedora 24 supported instead of Fedora 21
- remove Debian 7 (wheezy) support
- remove Ubuntu 12.04 (precise) support
  - remove brightbox repository support
  - garbage-collect now unneeded alternatives dependency

This would, per definition, make the module incompatible with all Foreman releases older than 1.12, but the only real incompatibilty is the removal of Ubuntu 12.04 support. I'd open a seperate PR for documentation updates, once this got in.


domcleal:
rest_v1 should be removed from the README.md too.

It might be worth deprecating rest_v2 in this PR too, or even in 5.x so we get back down to one provider?

Otherwise :+1:.

ekohl:
In params.pp there's also a Fedora 19 conditional which I think can be removed. It then looks like the passenger_* params have the same value and can be moved out of the case statement.

ekohl:
And ubuntu 12 can be removed from travis as well.

domcleal:
> And ubuntu 12 can be removed from travis as well.

I think if we first remove it from metadata.json everywhere (including here) then we can remove it from modulesync to tidy up the config. Once removed from metadata it won't be running at least.

ekohl:
>> And ubuntu 12 can be removed from travis as well.
> 
> I think if we first remove it from metadata.json everywhere (including here) then we can remove it from modulesync to tidy up the config. Once removed from metadata it won't be running at least.

Oh that's a good point.

mmoll:
Removed the F19 conditional and move the passenger parameters out. Regarding rest_v1, as there's still the foreman_hostgriup provider using apipie, removing the v2 provider for foreman_smartproxy probably wouldn't buy us much?

README.md
```diff
@@ -49,34 +49,6 @@ Thus 'master' will support the upcoming major version and the current stable.
...
-
-    ruby193-rubygem-foreman-tasks
-    ruby193-rubygem-ovirt_provision_plugin
-    ruby193-rubygem-puppetdb_foreman
-
-### Foreman 1.7 compatibility notes
-
-* set `apipie_task => 'apipie:cache'` as Foreman 1.7 packages didn't have
-  precompiled API docs
-* `foreman::compute::ec2` needs `package => 'foreman-compute'` on Foreman 1.7,
```
domcleal: Should we delete these `package` parameters or just leave them be?  At a minimum you should probably also remove the class/param docs saying about the Foreman 1.7/11 package names.

domcleal:
> Regarding rest_v1, as there's still the foreman_hostgriup provider using apipie, removing the v2 provider for foreman_smartproxy probably wouldn't buy us much?

Hm, true, leave it for now I guess and perhaps we can update everything to v3 and deprecate v2 at a later date.

README.md
```diff
@@ -49,34 +49,6 @@ Thus 'master' will support the upcoming major version and the current stable.
 The latest release (git tag, Puppet Forge) should support current and the
 previous stable release.
 
-### Foreman 1.11 compatibility notes
```
ekohl: I think it would be helpful for users if we tell them which version they should use on 1.11.

mmoll: That's right... added a note there

README.md
mmoll: cleaned up the respective classes

manifests/compute/openstack.pp
```diff
@@ -1,20 +1,12 @@
-# = Foreman OpenStack compute resource support
```
domcleal: This is still useful?

manifests/compute/rackspace.pp
```diff
@@ -1,20 +1,12 @@
-# = Foreman Rackspace compute resource support
```
domcleal: ditto

manifests/compute/ec2.pp
```diff
@@ -1,20 +1,12 @@
-# = Foreman EC2 compute resource support
```
domcleal: ditto

mmoll:
Added the headline also to the other compute resouces.

domcleal:
:+1: all looks correct.

ekohl:
Thanks!

PR #454, opened 2016-06-13
support plugin foreman_expire_hosts

timogoebel:


ekohl:
:+1: on passing tests.

domcleal:
Merged, thanks @timogoebel!

domcleal:
If you could also send a PR over to foreman-installer which adds the new class to config/ and a config migration file, it'll show up in the installer itself then.

timogoebel:
It's on my todo list, @domcleal . Thanks for reminding me, though.

PR #455, opened 2016-06-13
restrict extlib version to be Ruby 1.8 compatible

mmoll:
should fix the failing tests


metadata.json
```diff
@@ -38,7 +38,7 @@
     },
     {
       "name": "puppet/extlib",
-      "version_requirement": ">= 0.10.4 < 1.0.0"
+      "version_requirement": ">= 0.10.4 <= 0.11.3"
```
domcleal: Should this be left, so users can use a newer version if they want (and aren't using 1.8.7)?  This is usually how we manage dependencies in other projects.  We might need to pin it in foreman-installer instead.

domcleal:
:+1: if green.

domcleal:
Thanks @mmoll!

PR #458, opened 2016-06-22
use absolute class name in require

mmoll:


manifests/init.pp
```diff
@@ -24,7 +24,7 @@
 #
 # $plugin_prefix::              String which is prepended to the plugin package names
 #
-# $use_vhost::                  Enclose apache configuration in <VirtualHost>...</VirtualHost>
+# $use_vhost::                  Enclose apache configuration in VirtualHost tags
```
mmoll: trying to ninja this in, related to theforeman/theforeman.org#636

manifests/install.pp
```diff
@@ -29,8 +29,8 @@
   if $::foreman::passenger and $::foreman::passenger_ruby_package {
     package { $::foreman::passenger_ruby_package:
       ensure  => installed,
-      require => Class['apache'],
-      before  => Class['apache::service'],
+      require => Class['::apache'],
```
domcleal: I think these can be left out, if the lint check is fixed.

domcleal:
:+1: to the first two changes.

manifests/install.pp
domcleal: It's fixed, this shouldn't be necessary.

mmoll:
corrected.

domcleal:
:+1: subject to tests.

domcleal:
Thanks @mmoll!

PR #459, opened 2016-06-25
Fix a missing usage of the priority parameter in fragments

lazyfrosch:
The change with GH-418 missed to change 2 vhost fragment templates. The problem only got noticed by apache.

In addition I include spec tests to check for a proper usage of the conf.d directories.

(( We should include this in 5.2-stable branch with the next bugfix release ))


domcleal:
:+1: looks correct.

ekohl:
Thanks!

PR #460, opened 2016-06-30
Fixes #15539 - Bubble up ENC net exception

dLobatog:
When any exception occurs when retrieving the node definition, we
should raise the error so that we can catch it and serve the cached yaml
instead.

This patch generates a generic error to be caught in case there was a
network problem when fetching the node classification. Notice the actual
error is included in the message as `res.class` is the name of the error.


ekohl:
Looks like it makes sense

files/external_node_v2.rb
```diff
@@ -196,8 +196,10 @@ def enc(certname)
   end
   res = http.start { |http| http.request(req) }
 
-  raise "Error retrieving node #{certname}: #{res.class}\nCheck Foreman's /var/log/foreman/production.log for more information." unless res.code == "200"
-  res.body
+  return res.body if res.code == '200'
+
+  raise NodeRetrievalError, "Error retrieving node #{certname}: #{res.class}"\
+        "Check Foreman's /var/log/foreman/production.log for more information."
```
domcleal: Is this message displayed anywhere? If the node retrieval fails on the first run, then the user is only likely to see an error retrieving the cache file, which won't exist. If it fails on subsequent runs, the script will now only use the cache.

files/external_node_v2.rb
```diff
@@ -344,7 +346,7 @@ def watch_and_send_facts(parallel)
           result = enc(certname)
           cache(certname, result)
         end
-      rescue TimeoutError, SocketError, Errno::EHOSTUNREACH, Errno::ECONNREFUSED
+      rescue TimeoutError, SocketError, Errno::EHOSTUNREACH, Errno::ECONNREFUSED, NodeRetrievalError
```
domcleal: It might be best to allow Net::HTTPServiceUnavailable (from the bug report) to be raised here and to use the cache, rather than _all_ node retrieval errors (e.g. a 40* error should be a hard failure, not cache).

dLobatog:
Updated, now it raises the class of the error produced, so we just rescue the one mentioned in the report, ` Net::HTTPServiceUnavailable`, and display the error for all others.

files/external_node_v2.rb
```diff
@@ -196,8 +196,10 @@ def enc(certname)
   end
   res = http.start { |http| http.request(req) }
 
-  raise "Error retrieving node #{certname}: #{res.class}\nCheck Foreman's /var/log/foreman/production.log for more information." unless res.code == "200"
-  res.body
+  return res.body if res.code == '200'
+
+  raise res.class, "Error retrieving node #{certname}: #{res.class}"\
```
domcleal: This line seems to cause the following error if executed:

<pre>
/etc/puppetlabs/puppet/node.rb:201:in `raise': exception class/object expected (TypeError)
        from /etc/puppetlabs/puppet/node.rb:201:in `enc'
        from /etc/puppetlabs/puppet/node.rb:346:in `block in <main>'
        from /usr/share/ruby/timeout.rb:66:in `timeout'
        from /etc/puppetlabs/puppet/node.rb:345:in `<main>'
</pre>

(on Ruby 2.0.0)

dLobatog: What's the class?

dLobatog: Perhaps I can make it raise a 'known' exception if the error is '5xx' and make it fail otherwise...

domcleal: Sorry!  That was a `Net::HTTPInternalServerError` (the first I was testing with).

dLobatog:
Updated, latest version just raises `Net::HTTPServerException` when the error is a 5xx, otherwise it raises a regular exception with the error message we used to have. Cache is only triggered for 5xx's now

files/external_node_v2.rb
```diff
@@ -196,8 +196,12 @@ def enc(certname)
   end
   res = http.start { |http| http.request(req) }
 
-  raise "Error retrieving node #{certname}: #{res.class}\nCheck Foreman's /var/log/foreman/production.log for more information." unless res.code == "200"
-  res.body
+  return res.body if res.code == '200'
+
+  raise Net::HTTPServerException if (500..599).include?(res.code)
+
+  raise "Error retrieving node #{certname}: #{res.class}"\
```
domcleal: The `\n` has been removed here, causing:

    Error retrieving node foreman.example.com: Net::HTTPInternalServerErrorCheck Foreman's /var/log/foreman/production.log for more information.

files/external_node_v2.rb
```diff
@@ -196,8 +196,12 @@ def enc(certname)
   end
   res = http.start { |http| http.request(req) }
 
-  raise "Error retrieving node #{certname}: #{res.class}\nCheck Foreman's /var/log/foreman/production.log for more information." unless res.code == "200"
-  res.body
+  return res.body if res.code == '200'
+
+  raise Net::HTTPServerException if (500..599).include?(res.code)
```
domcleal: This doesn't work for me, as `res.code` is a string but it's being compared against an integer range.

mmoll:
please rebase

dLobatog:
Latest update extends the reach from `TimeoutError` (leaf exception) to the parent `Timeout::Error` that catches more kinds of timeout errors, like the one mentioned in the ticket, `Net::ReadTimeout` or `Net::OpenTimeout`.

files/external_node_v2.rb
```diff
@@ -196,8 +196,12 @@ def enc(certname)
   end
   res = http.start { |http| http.request(req) }
 
-  raise "Error retrieving node #{certname}: #{res.class}\nCheck Foreman's /var/log/foreman/production.log for more information." unless res.code == "200"
-  res.body
+  return res.body if res.code == '200'
+
+  raise Net::HTTPServerException if (500..599).include?(res.code.to_i)
```
domcleal: This fails with:

<pre>
/usr/share/ruby/net/http/exceptions.rb:5:in `initialize': wrong number of arguments (0 for 2) (ArgumentError)
        from /etc/puppetlabs/puppet/node.rb:201:in `exception'
        from /etc/puppetlabs/puppet/node.rb:201:in `raise'
        from /etc/puppetlabs/puppet/node.rb:201:in `enc'
        from /etc/puppetlabs/puppet/node.rb:348:in `block in <main>'
        from /usr/share/ruby/timeout.rb:66:in `timeout'
        from /etc/puppetlabs/puppet/node.rb:347:in `<main>'
</pre>

on ruby 2.0.0p598, it's also reproducible in irb.

mmoll:
@dLobatog what's the status here? this also needs a rebase to pull in new test fixes with recent versions

mmoll:
ping?

ekohl:
There's a merge conflict now.

dLobatog:
Thanks for the reminder @ekohl @mmoll 

I've fixed the problem with the exception as per @domcleal 's comment

files/external_node_v2.rb
```diff
@@ -376,6 +382,7 @@ def watch_and_send_facts(parallel)
           cache(certname, result)
         end
       rescue TimeoutError, SocketError, Errno::EHOSTUNREACH, Errno::ECONNREFUSED, FactUploadError
+        Net::HTTPServerException
```
domcleal: This doesn't catch the error, it's on a new line with no comma on the preceding line. Currently it simply fails with a message stating there's a 500 erver error.

files/external_node_v2.rb
```diff
@@ -213,8 +213,14 @@ def enc(certname)
   end
   res = http.start { |http| http.request(req) }
 
-  raise "Error retrieving node #{certname}: #{res.class}\nCheck Foreman's /var/log/foreman/production.log for more information." unless res.code == "200"
-  res.body
+  return res.body if res.code == '200'
+
+  if (500..599).include?(res.code.to_i)
+    raise Net::HTTPServerException.new('500 error in Foreman', res)
```
domcleal: 1) "500" isn't correct for a non-500 error (e.g. 503), it should use the code.

<pre>
# node.rb $(hostname -f)
2016-12-22 05:51:30 56acff41 [app] [I] Started GET "/node/foreman.example.com?format=yml" for 127.0.0.1 at 2016-12-22 05:51:30 -0500
2016-12-22 05:51:30 56acff41 [app] [I] Processing by HostsController#externalNodes as YML
2016-12-22 05:51:30 56acff41 [app] [I]   Parameters: {"name"=>"foreman.example.com"}
2016-12-22 05:51:30 56acff41 [app] [I] Completed 503 Service Unavailable in 4ms (ActiveRecord: 0.8ms)
500 error in Foreman
</pre>

2) This doesn't seem to show any logging to stderr?

ekohl:
Any plan to revisit this?

PR #464, opened 2016-07-18
rename environment to rails_env

igalic:
rename foreman::environment parameter to rails_env
This expresses better what this parameter does. As an additional
benefit, it doesn't interfere with databindings (from hiera), when our
foreman server isn't running in the puppet environment 'production'.

n.b.: this is a breaking change.


manifests/init.pp
```diff
@@ -251,7 +251,7 @@
   $user                      = $::foreman::params::user,
   $group                     = $::foreman::params::group,
   $user_groups               = $::foreman::params::user_groups,
-  $environment               = $::foreman::params::environment,
+  $rails_env                 = $::foreman::params::rails_env,
```
igalic: does this use tabs anywhere?

mmoll: nope, I guess it's only GitHub's way to highlight the whitespace change

mmoll:
please remove the "BREAKING CHANGE" in the commit message,  maybe that's also what's keeping the travis tests from running (but I have no idea)?

spec/classes/foreman_spec.rb
```diff
@@ -42,4 +42,98 @@ class { 'foreman::providers':
       end
     end
   end
+
+  context 'with all parameters' do
+    let :facts do
+      default_facts.merge({
```
domcleal: This variable isn't defined, causing test errors.

spec/classes/foreman_spec.rb
```diff
@@ -42,4 +42,98 @@ class { 'foreman::providers':
...
+        :operatingsystem        => 'RedHat',
```
domcleal: If you place this in the loop above, OS facts will be set, rather than only test one OS.

igalic: this is all thanks to basing my patch on a stale repo, and then rebasing

igalic:
`This module does not support osfamily  at`.

Oh, yes, the  &nbsp;  system. Yeah, I wouldn't support that either.

spec/classes/foreman_spec.rb
```diff
@@ -41,5 +41,92 @@ class { 'foreman::providers':
...
+          :db_type                   => 'postgresql',
+          :db_adapter                => 'UNSET',
+          :db_host                   => 'UNSET',
+          :db_port                   => 'UNSET',
+          :db_database               => 'UNSET',
+          :db_username               => 'foreman',
+          :db_password               => 'secret',
+          :db_sslmode                => 'UNSET',
+          :db_pool                   => 5,
+          :apipie_task               => 'apipie:cache:index',
```
mmoll: The signature of init.pp changed, I guess that's a leftover of the old base you had that PR on initially.

igalic:
@mmoll thanks for the review, i've now updated my test to reflect the current init.pp

mmoll:
same with configure_ipa_repo it seems... maybe better to generate that list again from current master

igalic:
~_~ i thought i had caught them all

igalic:
i've literally copy-pasted the init params now, and its the same
i have no idea what's going, and which platform we're testing here that returns "" for `osfamily`

mmoll:
```diff
diff --git a/spec/classes/foreman_spec.rb b/spec/classes/foreman_spec.rb
index 72e8c18..c7f66ca 100644
--- a/spec/classes/foreman_spec.rb
+++ b/spec/classes/foreman_spec.rb
@@ -40,7 +40,6 @@ describe 'foreman' do
         it { is_expected.to compile.with_all_deps }
         it { should contain_package('apipie-bindings').that_subscribes_to('Class[foreman::repo]') }
       end
-    end
 
     describe 'with all parameters' do
       let :params do {
@@ -137,3 +136,4 @@ describe 'foreman' do
     end
   end
 end
+end
```
will fix that problem and leads to the next one :)

igalic:
fixed that and all the others (i think ;)

mmoll:
:+1: from me.

I know it's asked a bit much, but would you be willing to also create a migration for that param rename here? https://github.com/theforeman/foreman-installer/tree/develop/config/foreman.migrations

If not, just drop a line here and I'll create one after this PR got merged.

igalic:
I'll try, tomorrow

igalic:
https://github.com/theforeman/foreman-installer/pull/191

domcleal:
Merged, thanks @igalic!

PR #465, opened 2016-07-19
fix for ignoring new nodes while pushing facts with external_node_v2 …

bazilek:
external_node_v2.rb --push-facts-parallel --watch-facts does not push facts for the new nodes (nodes that were provisioned after script had been started ). please consider merging this fix.


domcleal:
Merged, thanks @bazilek!

PR #469, opened 2016-08-16
Improve logs when something goes wrong

ares:


files/external_node_v2.rb
```diff
@@ -206,8 +210,15 @@ def upload_facts(certname, req)
...
     res = initialize_http(uri)
     res.read_timeout = SETTINGS[:timeout]
-    res.start { |http| http.request(req) }
-    cache("#{certname}-push-facts", "Facts from this host were last pushed to #{uri} at #{Time.now}\n")
+    res.start do |http|
+      response = http.request(req)
+      if response.code.start_with?('2')
+        cache("#{certname}-push-facts", "Facts from this host were last pushed to #{uri} at #{Time.now}\n")
+      else
+        puts "Server responded with: #{response.code} #{response.message}"
```
domcleal: Sending output to stdout will probably interfere with the ENC output, should this be to stderr or a log file?

files/external_node_v2.rb
```diff
@@ -163,6 +165,8 @@ def generate_fact_request(certname, filename)
     rescue => e
       raise "Could not generate facts for Foreman: #{e}"
     end
+  else
+    raise 'Last facts are older than last upload time, rerun puppet agent on host to get newer facts'
```
domcleal: Raising an exception here might cause the watch process to exit entirely, as I don't think anything catches it. Shouldn't this be logged like other instances?

files/external_node_v2.rb
```diff
@@ -102,6 +102,8 @@ def process_all_facts(http_requests)
       elsif req
         upload_facts(certname, req)
       end
+    else
+      puts "Fact file #{f} does not contain any fact"
```
domcleal: ditto, concern about printing to stdout from the ENC

files/external_node_v2.rb
ares: there's a raise few lines above so I took the same approach, I'll change to `$stderr.puts`

ares:
thanks @domcleal for quick review, re-pushed with the changes

ares:
rebased

files/external_node_v2.rb
```diff
@@ -206,8 +210,15 @@ def upload_facts(certname, req)
...
     res = initialize_http(uri)
     res.read_timeout = SETTINGS[:timeout]
-    res.start { |http| http.request(req) }
-    cache("#{certname}-push-facts", "Facts from this host were last pushed to #{uri} at #{Time.now}\n")
+    res.start do |http|
+      response = http.request(req)
+      if response.code.start_with?('2')
+        cache("#{certname}-push-facts", "Facts from this host were last pushed to #{uri} at #{Time.now}\n")
+      else
+        $stderr.puts "Server responded with: #{response.code} #{response.message}"
```
domcleal: This should state that it's during the fact upload that the server responded this way, and possibly that the error will be ignored, execution continues?

domcleal:
Seems to be OK, the messages show up interactively and don't seem to interfere with the environment/node retrieval during the first step of the run (when the last facts warning will always print).

ares:
Thanks for the review, fixed.

domcleal:
:+1: looks OK.

files/external_node_v2.rb
```diff
@@ -163,6 +165,8 @@ def generate_fact_request(certname, filename)
     rescue => e
       raise "Could not generate facts for Foreman: #{e}"
     end
+  else
+    $stderr.puts 'Last facts are older than last upload time, rerun puppet agent on host to get newer facts'
```
mmoll: I'd need to debug this further, I'm getting this message quite often...

domcleal: You'd get it every time you run node.rb by hand, as the facts file wouldn't have been refreshed.

mmoll: Nope, that was when running `puppet agent --test` or even on normal Puppet runs:
```
# grep -c 'rerun puppet agent on host to get newer facts' puppetserver.log
149
```
That's from the last 10 hours or so.

domcleal: Right, you would also receive the message during the initial node info request, prior to pluginsync and the catalog request (containing the facts upload, the main ENC call). If it's logged by Puppet Server though, then it probably isn't a good idea to have this message.

ares: should i remove it then? or any idea how to detect how is the script triggered?

domcleal: I don't think you could tell how the script is triggered, it'd be identical to running the script standalone to fetch node information without a corresponding fact upload.

ares: thanks, removed :disappointed:

mmoll:
works fine, merged, díky @ares!

PR #472, opened 2016-08-25
Fix missing default parameters for strict variable tests

domcleal:
- relies upon defined() support for variables, present in Puppet 3.6
- removes unused ipa_authentication parameter from inner class
- defines defaults for foreman::config::passenger in many tests
- removes missing CRL parameter override test that relies on inheritance
- adds 'foreman' class to plugin tests to define version etc. parameters

---

The requirement on Puppet 3.6 will affect Ubuntu 14.04's default version of Puppet (3.4.3), but I can't think of another way to solve this reliably without disabling strict variable testing/support.


mmoll:
it wouldn't be the first time to tell users to use the PL repos, so :+1:

PR #474, opened 2016-08-29
external_node_v2.rb: generate_fact_request: don't bail out on nodes for which facts haven'…

pkranenburg:
…t been recorded yet; just ignore the request.

The classifier can encounter this situation when a puppet agent on a node contacts a puppet server instance for the first time. In such cases, the node classifier can get called before the agent has
submitted a set of facts.

See also issue #5925 (maybe this condition happens more frequently on the latest puppet 4 servers?)

### DIFF
--- a/files/external_node_v2.rb
+++ b/files/external_node_v2.rb
@@ -151,7 +151,7 @@ def generate_fact_request(certname, filename)
   # Temp file keeping the last run time
   stat = stat_file("#{certname}-push-facts")
   last_run = File.exists?(stat) ? File.stat(stat).mtime.utc : Time.now - 365*24*60*60
-  last_fact = File.exists?(filename) ? File.stat(filename).mtime.utc : Time.new(0)
+  last_fact = File.exists?(filename) ? File.stat(filename).mtime.utc : Time.at(0)
   if last_fact > last_run
     begin
       uri = URI.parse("#{url}/api/hosts/facts")
--- a/files/external_node_v2.rb
+++ b/files/external_node_v2.rb
@@ -151,7 +151,7 @@ def generate_fact_request(certname, filename)
   # Temp file keeping the last run time
   stat = stat_file("#{certname}-push-facts")
   last_run = File.exists?(stat) ? File.stat(stat).mtime.utc : Time.now - 365*24*60*60
-  last_fact = File.stat(filename).mtime.utc
+  last_fact = File.exists?(filename) ? File.stat(filename).mtime.utc : Time.new(0)
   if last_fact > last_run
     begin
       uri = URI.parse("#{url}/api/hosts/facts")
--- a/spec/unit/foreman_external_node_spec.rb
+++ b/spec/unit/foreman_external_node_spec.rb
@@ -55,7 +55,7 @@ describe 'foreman_external_node' do
   it "should NOT connect to the URL in the manifest" do
 
     enc.stubs(:stat_file).with('fake.host.fqdn.com-push-facts').returns("/tmp/fake.host.fqdn.com-push-facts.yaml")
-    # first :exists? call is for 'push-facts'; second one for non-existant facts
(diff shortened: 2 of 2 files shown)


domcleal:
This seems like a good change, thanks.

Could you please look at the failing test? It looks likely that the File stubs for `exists?` isn't specific enough and should only be stubbed to return `false` when called `.with(stat_file)`. You will probably then need to add a second stub for the fact file that returns `true` instead.

If you are able to extend the tests to check it runs successfully with no fact file, I'm sure that would be very useful too.

domcleal:
The test is still failing after the last push:

<pre>
  1) foreman_external_node should connect to the URL in the manifest
     Failure/Error: expect(webstub).to have_been_requested
       The request POST http://localhost:3000/api/hosts/facts with body {"fake"=>"data"} was expected to execute 1 time but it executed 0 times

       The following requests were made:
       No requests were made.
       ============================================================
     # ./spec/unit/foreman_external_node_spec.rb:39:in `block (2 levels) in <top (required)>'
</pre>

pkranenburg:
It turns out that using '.with(filename)' on File.stubs(:exists?) does  not have the desired effect. Now done by passing all desired return values in the order they are needed.

Also, a test is added for the case where no previously recorded facts are present.

Cheers,  --Paul Kranenburg
________________________________

ekohl:
Looks like the tests fail under ruby 1.8.7

pkranenburg:
Yeah, no idea why...

--pk
________________________________

files/external_node_v2.rb
```diff
@@ -151,7 +151,7 @@ def generate_fact_request(certname, filename)
   # Temp file keeping the last run time
   stat = stat_file("#{certname}-push-facts")
   last_run = File.exists?(stat) ? File.stat(stat).mtime.utc : Time.now - 365*24*60*60
-  last_fact = File.stat(filename).mtime.utc
+  last_fact = File.exists?(filename) ? File.stat(filename).mtime.utc : Time.new(0)
```
domcleal: Time.new [doesn't take any arguments in Ruby 1.8.7](https://ruby-doc.org/core-1.8.7/Time.html#method-c-new), it might be best to use `Time.at(0)` (returns Unix epoch, 1970).

pkranenburg: Ah! Thanks for pointing that out; i'll fix it later today
________________________________


Time.new doesn't take any arguments in Ruby 1.8.7<https://ruby-doc.org/core-1.8.7/Time.html#method-c-new>, it might be best to use Time.at(0) (returns Unix epoch, 1970).

spec/unit/foreman_external_node_spec.rb
```diff
@@ -51,6 +52,19 @@ class Enc
     expect(webstub).to have_been_requested.times(4)
   end
 
+  it "should NOT connect to the URL in the manifest" do
+
```
mmoll: superfluous newline, but we'll fix that when merging, as I'd like to avoid another hours long test run

spec/unit/foreman_external_node_spec.rb
```diff
@@ -51,6 +52,19 @@ class Enc
...
+    enc.stubs(:stat_file).with('fake.host.fqdn.com-push-facts').returns("/tmp/fake.host.fqdn.com-push-facts.yaml")
+    # first :exists? call is for 'push-facts'; second one for non-existent facts
+    File.stubs(:exists?).returns(false,false)
+    File.stubs(:stat).returns(stub(:mtime => Time.now.utc))
+    enc.stubs(:build_body).returns({'fake' => 'data'})
+
+    req = enc.generate_fact_request('fake.host.fqdn.com','non-existent-facts.yaml')
+    expect(req).to be_nil
+
```
mmoll: dito

mmoll:
:+1:

domcleal:
Merged as bbf6899, thanks @pkranenburg!

PR #475, opened 2016-08-30
Added SVG gzip explicit encoding

lzap:
We already precompile SVG files but it's not enabled here.

https://github.com/theforeman/foreman/blob/develop/config/initializers/assets.rb#L64


ekohl:
:+1:

mmoll:
merged, díky @lzap!

PR #477, opened 2016-09-05
Allow configuration of non-default ports for Apache

stdietrich:
Both ports for Apache are currently hard-coded in the module and set to 80 for HTTP and 443 for HTTPS.
This change allows to configure non-default ports for Apache through 2 new parameters.
Running Foreman on non-default ports is e.g. usable for running multiple instances behind a loadbalancer. Usually those instances are not running on the default ports.

### DIFF
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -120,6 +120,12 @@
 #
 # $vhost_priority::             Defines Apache vhost priority for the Foreman vhost conf file.
 #
+# $server_port::                Defines Apache port for HTTP requests
+#                               type:integer
+#
+# $server_ssl_port::            Defines Apache port for HTTPS reqquests
+#                               type:integer
+#
 # $server_ssl_ca::              Defines Apache mod_ssl SSLCACertificateFile setting in Foreman vhost conf file.
 #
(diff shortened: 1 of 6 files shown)


ekohl:
Looks good to me. :+1: on green tests.

stdietrich:
I had to update the PR, as I missed the passenger prestart configuration.
With the non-default ports configured, it would no longer prestart Foreman.

manifests/init.pp
```diff
@@ -120,6 +120,10 @@
 #
 # $vhost_priority::             Defines Apache vhost priority for the Foreman vhost conf file.
 #
+# $server_port::                Defines Apache port for HTTP requests
```
domcleal: Can you add `type:integer` below these two parameters please? It's used by Kafo to determine the data type.

mmoll:
FYI, to save resources on travis I cancelled all builds except one, which came back green.

manifests/init.pp
stdietrich: I have added the type and pushed a new version.

mmoll:
merged, danke @stdietrich!

PR #478, opened 2016-09-08
Prepare 6.0.0 changelog

domcleal:
[ci skip]

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,41 @@
 # Changelog
 
+## 6.0.0
+* New classes to install Foreman plugins:
+    * foreman::plugin::azure to install Azure compute resource plugin
+    * foreman::plugin::expire_hosts to install expire hosts plugin
+    * foreman::plugin::host_extra_validator to install hostname validator plugin
+* New or changed parameters:
+    * Add server_port, server_ssl_port parameters to change Apache vhost ports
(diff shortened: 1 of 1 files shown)


CHANGELOG.md
```diff
@@ -1,5 +1,41 @@
...
+    * Add server_port, server_ssl_port parameters to change Apache vhost ports
+    * Rename environment parameter to rails_env to fix compatibility with data
+      bindings
+* Other changes and fixes:
+    * node.rb: skip facts upload when facts file is missing, retrieves ENC
+      output anyway
+    * node.rb: improve logging for empty facts and failed fact uploads
+    * Change reports upload to use new config_reports API
+    * Change Yum GPG key URLs to HTTPS
+    * Fix missing default parameters for strict variables compatibility,
```
mmoll: #476 worked that around (at least for that one place), so requiring 4.6 should be ok, like it's stated in line 24?

domcleal: It only worked around it when strict variables is disabled on Puppet 3.6.0-3.7.4, unfortunately it'll still raise an error on the pick() line if strict vars is enabled on those versions. I was trying to say here that you can use the module with 3.6.0 or higher, but only 3.7.5 or higher if you have strict vars enabled.

mmoll: ah, I see... :+1:

mmoll:
merged, thanks @domcleal!

PR #481, opened 2016-09-17
Simplify specs

ekohl:


ekohl:
Note this is currently mostly to check if this works on all supported versions. If it does, I'll make a PR to modulesync first. It already includes https://github.com/theforeman/foreman-installer-modulesync/pull/39

ekohl:
Still the same build failure as before on 1.8.7 when moving to a newer rspec-puppet-facts :(

mmoll:
yeah, it's the same old problem... if all else fails, we still could update the expectation depending on the ruby version...

mmoll:
looks good otherwise :+1:

ekohl:
I'll rebase on a modulesync with https://github.com/theforeman/foreman-installer-modulesync/pull/39 and leave out the fact stuff. That's not that much anyway. I'll also do a PR for every module to simplify.

ekohl:
Given we're dropping 1.8.7 from master and can use the newer rspec-puppet-facts I'll revisit this. I think we should do a modulesync PR to add the custom facts we rely upon.

ekohl:
Rebased. The add_custom_fact part also needs a modulesync PR which I'll submit if these tests pass.

spec/classes/foreman_install_spec.rb
```diff
@@ -1,16 +1,9 @@
...
 describe 'foreman::install' do
-  on_supported_os.each do |os, facts|
-    next if only_test_os() and not only_test_os.include?(os)
-    next if exclude_test_os() and exclude_test_os.include?(os)
-
+  on_os_under_test.each do |os, facts|
     context "on #{os}" do
-      let :facts do
-        facts.merge({
-          :concat_basedir => '/tmp',
```
domcleal: There are two more instances of concat_basedir in this file that could be removed.

ekohl:
Updated to remove the two additional instances of concat_basedir. Also added another commit to clean up duplicate expectations.

mmoll:
merged, bedankt @ekohl!

Issue #482, opened 2016-09-23
Error 400 on SERVER: undefined method `has_key?' 


kalinux:
Hi,

I get some not usual error which i cant solve by my self, therefore please help me configure out whats going on.

manifests file:

```
  $f = { item        => 'fact_values',
        search       => 'fact = ipaddres',
        per_page     => '1000',
        foreman_url  => 'https://foreman.mycompany.com',
        foreman_user => 'apiuser',
        foreman_pass => 'password',
        filter_result => 'true' }


  $hosts = foreman($f)

  file {'/opt/karen09.conf':
    content => "$hosts",
  }
```

Error:

```
Error: Could not retrieve catalog from remote server: Error 400 on SERVER: undefined method `has_key?' for ["foreman.mycompany.com", {"fqdn"=>"foreman.mycompany.com"}]:Array at /etc/puppet/environments/production/modules/profiles/manifests/manifests.pp:13 on node node.mycompany.com
Warning: Not using cache on failed catalog
Error: Could not retrieve catalog; skipping run
```

Kindly Regards
Karen

ares:
This is probably not the best place to ask for support, try our $theforeman irc channel @freenode. First question that I'd have is what is "/etc/puppet/environments/production/modules/profiles/manifests/manifests.pp", it does not look like something we ship.

ekohl:
Which foreman version are you using? I wonder if this assumes a specific API version (I'm assuming v1) while Foreman responds with another API (I'd guess v2).

Karen09:
Hi @ares,

Case was indeed question , but it can actually cause to have changes in file, that`s why i had asked here.

"/etc/puppet/environments/production/modules/profiles/manifests/manifests.pp" - this is manifest file which is printed in above section of first comment . 

Mostly i will have a look at this file for solution : 
```
https://github.com/theforeman/puppet-foreman/blob/master/lib/puppet/parser/functions/foreman.rb
```

Hi @ekohl ,

Yes you are true , there is most of time api version problem , I tried CURL command with v1 and v2 and im getting the same output . (p.s. am getting with CURL all information which i need exactly ) . Foreman version : 1.12

Kindly Regards
Karen Harutyunyan

ekohl:
I think your use of ```filter_result``` is incorrect. I'd first try it without ```filter_result``` and otherwise set it to a correct value.

Karen09:
Hi @ekohl 

[...]

PR #483, opened 2016-09-30
Solving: undefined method `has_key?'

kalinux:
Hi all,

Using foreman 1.12.1 cause me to get this error, when im trying to get some facts from foreman using lib foreman() in manifest file. I had solving it with this pull request. Always open to hear any suggestions.

Regards
Karen

### DIFF
--- a/lib/puppet/parser/functions/foreman.rb
+++ b/lib/puppet/parser/functions/foreman.rb
@@ -36,6 +36,7 @@
 #
 # Happy Foreman API-ing!
 
+require "pry"
 require "net/http"
 require "net/https"
 require "uri"
@@ -71,7 +72,7 @@ module Puppet::Parser::Functions
       http = Net::HTTP.new(uri.host, uri.port)
       http.use_ssl = true if uri.scheme == 'https'
       http.verify_mode = OpenSSL::SSL::VERIFY_NONE if http.use_ssl?
-      results = Timeout::timeout(5) { PSON.parse http.request(req).body }
+      results = Timeout::timeout(25) { PSON.parse http.request(req).body }
     rescue Exception => e
       raise Puppet::ParseError, "Failed to contact Foreman at #{foreman_url}: #{e}"
     end


lib/puppet/parser/functions/foreman.rb
```diff
@@ -36,6 +36,7 @@
 #
 # Happy Foreman API-ing!
 
+require "pry"
```
ekohl: I don't think pry is available on every installation.

ekohl:
I'd expect that a timeout would throw an exception rather than returning some invalid data. I wonder if just using pry solves it. It may have some weird side effects that would trigger it.

lib/puppet/parser/functions/foreman.rb
Karen09: if its not available on every installation , how we gonna solve it? maybe adding docs in https://theforeman.org/2012/01/getting-foreman-search-results-into.html ? but other side , as i know "pry" is only debugging so it cant effect for others who`s using this module. is it not?

mmoll: What is the purpose of this line anyway? IMHO it's simply not possible to *require* a debugger gem for normal production installs (AFAICT that would even needed to get installed in the JRuby Puppetserver if running on that).

The fix for the problem is the increase of the timeout, no?

ekohl: Looking at the original problem (https://github.com/theforeman/puppet-foreman/issues/482) I find it hard to see how increasing the timeout solves it. That makes me wonder if pry has some sort of side effect.

mmoll: I see... maybe capturing the failing HTTP request to the foreman API leads to some other hint?

Karen09: i had tried to put 25 seconds for timeout, somehow it`s does`nt solve problem. after installing pry from first run its work perfectly. Section about  timeout  is less important.

ekohl: @domcleal any idea why this would happen?

ekohl:
What's the status here?

ekohl:
Closing for inactivity

PR #485, opened 2016-10-05
Fixes #16802 - enable hammer openscap plugin installation

ares:


mmoll:
@ares tests fail...

ares:
@mmoll thanks for warning, should be green now

manifests/plugin/openscap.pp
```diff
@@ -6,4 +6,10 @@
 #
 class foreman::plugin::openscap {
   foreman::plugin {'openscap': }
+
+  if $foreman::cli::manage_root_config {
```
domcleal: Isn't the presence of this variable (and hammer_plugin_prefix, below) dependent on the parse order? I don't think it's reliable code. I think it probably needs to be in its own class that's included directly, not here.

mmoll:
tests fail now because of `foreman::cli::manage_root_config`

manifests/plugin/openscap.pp
ares: Sorry, it has been a while since I've written puppet code. Does that mean that I'd have to create a new puppet class `foreman::cli::plugin::openscap` which would include both foreman and foreman::cli classes and check `$foreman::cli::manage_root_config` and `$foreman::hammer_plugin_prefix` and let users to explicitly instantiate this class? If that's the case, is there any way to keep the logic here so the openscap plugin would automatically install the hammer plugin if the plugin is enabled and foreman::cli manages the config? Maybe using some `defined("foreman::cli")` and `require`/`include`?

domcleal: It'd probably have to be in its own class, yes. I don't think using `defined("foreman:cli")` would be any more reliable, it's still dependent on the parse order.

ekohl: A ```foreman::cli::plugin::openscap``` or just ```foreman::cli::openscap``` sounds like the cleanest solution to me.

ares:
Ok guys, I hope this is what you meant. I moved the prefix logic into `foreman::cli`, I guess it belongs to this class, people might want to use this class without `foreman`. I tested with https://github.com/theforeman/foreman-installer/pull/210 and it seems to work well. So far there's no logic other than installing package which is why I didn't add something like `foreman::cli:plugin` resource. I guess other plugins will also only install the package.

manifests/cli.pp
```diff
@@ -12,26 +12,29 @@
...
+#                         type:boolean
 #
-# $request_timeout::    API request timeout, set -1 for infinity
-#                       type:integer
+# $request_timeout::      API request timeout, set -1 for infinity
+#                         type:integer
+#
+# $hammer_plugin_prefix:: Hammer plugin package prefix based, normally on platform
 #
 # $version::            foreman-cli package version, it's passed to ensure parameter of package resource
```
domcleal: Alignment of the text is out here compared to the ones above.

manifests/cli.pp
```diff
@@ -12,26 +12,29 @@
...
-#                       type:boolean
+# $refresh_cache::        Check API documentation cache status on each request
+#                         type:boolean
 #
-# $request_timeout::    API request timeout, set -1 for infinity
-#                       type:integer
+# $request_timeout::      API request timeout, set -1 for infinity
+#                         type:integer
+#
+# $hammer_plugin_prefix:: Hammer plugin package prefix based, normally on platform
```
domcleal: Remove the comma here.

manifests/cli/openscap.pp
```diff
@@ -0,0 +1,11 @@
+# = Hammer OpenSCAP plugin
+#
+# This class installs OpenSCAP plugin
```
domcleal: "[..] installs the OpenSCAP plugin for Hammer CLI" ?

spec/classes/foreman_cli_openscap.rb
```diff
@@ -0,0 +1,11 @@
+require 'spec_helper'
```
domcleal: This file's missing a _spec.rb extension, it won't run automatically.

spec/classes/foreman_cli_openscap.rb
```diff
@@ -0,0 +1,11 @@
...
+
+describe 'foreman::cli::openscap' do
+  let(:facts) do
+    on_supported_os['redhat-7-x86_64']
+  end
+
+  let(:pre_condition) { 'include foreman' }
```
domcleal: Probably foreman::cli?

domcleal:
Please also address @theforeman-bot's comment about the ticket, it's in the wrong project for Redmine to track the change.

ares:
I addressed all @domcleal's comments and moved the issue under installer project in Redmine.

domcleal:
Merged, thanks @ares.

Issue #473, opened 2016-08-29
foreman::rake {'apipie:cache:index'} not scheduled with db_managed=>false


mgriffin13:
In my highly available, distributed Foreman installation, we use an externally managed postgresql database, so all of my nodes are set to $db_manage=>false. If this parameter is false, though, then **foreman::rake { 'apipie:cache:index' }** is never scheduled to run, which causes downstream impacts that rely on the cache being present. I dont want foreman-rake to be scheduled every time puppet runs, but I need the cache to be built when Foreman is initially installed.

mmoll:
@domcleal @ekohl a very naive approach, could this work in general?
```diff
--- a/manifests/database.pp
+++ b/manifests/database.pp
@@ -1,9 +1,7 @@
 # Set up the foreman database
 class foreman::database {
-  if $::foreman::db_manage {
     validate_string($::foreman::admin_username, $::foreman::admin_password)
 
-    $db_class = "foreman::database::${::foreman::db_type}"
     $seed_env = {
       'SEED_ADMIN_USER'       => $::foreman::admin_username,
       'SEED_ADMIN_PASSWORD'   => $::foreman::admin_password,
@@ -20,7 +18,12 @@ class foreman::database {
       $foreman_service = Class['foreman::service']
     }
 
-    class { $db_class: } ~>
+  if $::foreman::db_manage {
+    $db_class = "foreman::database::${::foreman::db_type}"
+    class { $db_class:
+      notify => foreman_config_entry['db_pending_migration']
+    }
+  }
     foreman_config_entry { 'db_pending_migration':
       value => false,
       dry   => true,
@@ -38,5 +41,4 @@ class foreman::database {
     foreman::rake { 'apipie:cache:index':
       timeout => 0,
     }
-  }
 }
```

domcleal:
Yeah, it probably would, though I think a new parameter for whether to run db:migrate/seed is necessary as there are still cases (e.g. multi-host) where they should be disabled to avoid multiple systems running the same commands simultaneously.

PR #488, opened 2016-10-15
introduce db_managed_rake flag

mmoll:
This makes it possible to have the databases on an external server,
using db_manage=false, while still populating the database with the
required rake jobs.

closes GH-473

### DIFF
--- a/manifests/database.pp
+++ b/manifests/database.pp
@@ -1,9 +1,20 @@
 # Set up the foreman database
 class foreman::database {
   if $::foreman::db_manage {
-    validate_string($::foreman::admin_username, $::foreman::admin_password)
-
     $db_class = "foreman::database::${::foreman::db_type}"
+
+    $after_db_notify = $::foreman::db_manage_rake ? {
+      true    => Foreman_config_entry['db_pending_migration'],
+      default => undef,
+    }
+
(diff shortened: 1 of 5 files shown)


manifests/database.pp
```diff
@@ -1,9 +1,14 @@
 # Set up the foreman database
 class foreman::database {
   if $::foreman::db_manage {
-    validate_string($::foreman::admin_username, $::foreman::admin_password)
-
     $db_class = "foreman::database::${::foreman::db_type}"
+    class { $db_class:
+      notify => Foreman_config_entry['db_pending_migration'],
```
domcleal: This might be missing if `db_manage` is true, but `db_manage_rake` is false. I'd suggest either making it conditional on both (a separate conditional that chains the two if both parameters are set) or change the rake task block to `if $::foreman::db_manage or $::foreman::db_manage_rake` so db_manage_rake has no function when db_manage is set (I'm unsure if this is desirable).

ehelms:
@mmoll awesome! This will allow me to cleanup https://github.com/Katello/forklift/pull/269 to remove the orchestration hacks.

mmoll:
added some (light) tests.

ekohl:
Thanks!

PR #492, opened 2016-10-20
Bugfix: Use node cache, when fact upload fails

aptituz:
When running as an ENC not having a connection to the foreman
host makes the execution fail, instead of using the cache,
since the exception raised by upload_facts is not catched.
This patch does three things:
- introduce a new exception class for fact upload errors
  (so we don't have to catch the most generic exception
   and thus possibly catching to much)
- move the execution of that method during ENC run into the
  logic which handles connection problems by falling back
  to cached data
- additionally catch the new exception in the mentioned
  logic

### DIFF
--- a/files/external_node_v2.rb
+++ b/files/external_node_v2.rb
@@ -71,6 +71,8 @@ class Http_Fact_Requests
   end
 end
 
+class FactUploadError < StandardError; end
+
 require 'etc'
 require 'net/http'
 require 'net/https'
@@ -218,7 +220,8 @@ def upload_facts(certname, req)
       end
     end
   rescue => e
-    raise "Could not send facts to Foreman: #{e}"
+    $stderr.puts "During fact upload occured an exception: #{e}"
+    raise FactUploadError, "Could not send facts to Foreman: #{e}"
   end
 end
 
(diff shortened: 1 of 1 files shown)


files/external_node_v2.rb
```diff
@@ -342,22 +344,24 @@ def watch_and_send_facts(parallel)
...
+          if SETTINGS[:facts]
+            req = generate_fact_request certname, "#{puppetdir}/yaml/facts/#{certname}.yaml"
+            upload_facts(certname, req)
+          end
+
           result = enc(certname)
           cache(certname, result)
         end
-      rescue TimeoutError, SocketError, Errno::EHOSTUNREACH, Errno::ECONNREFUSED
+      rescue TimeoutError, SocketError, Errno::EHOSTUNREACH, Errno::ECONNREFUSED, FactUploadError
```
domcleal: Could you add some $stderr output to this rescue, or perhaps to the fact upload part? I like the behaviour of using the cache, but there's no indication on running node.rb that it's failing. See 8e86e15.

aptituz:
@mmoll Thanks for the hint of limited visibility. Dunno, how this happened, I just replied to the comment of @domcleal.

My question regarding the change request was, if the existing code wouldn't need that indication of problems, too.

Apart from that I'm not exactly sure what would be the right thing to do: Just write to stderr? Would puppet pick this up? And if done from within the upload_facts method: Would this in any way conflict with the other uses of the enc script, e.g. standalone fact upload?

(Sorry if the questions might seem odd, I'm still learning at a beginner knowledge of ruby :)

domcleal:
> Sure. Just wondering: Wouldn't the enc call itself need this, too? Otherwise the behaviour would be different, wouldn't it?

Indeed, it'd be best on both for sure. Since you were only changing one part, I wasn't adamant that you need to fix it in both requests.

> Apart from that I'm not exactly sure what would be the right thing to do: Just write to stderr? Would puppet pick this up?

No, I don't think it will capture it at all. In my tests of the existing logging, Puppet will just ignore content on stderr and only read from stdout, which is intentional here. It will though be useful for people running the script directly to later find out why facts aren't updating to see the errors.

> And if done from within the upload_facts method: Would this in any way conflict with the other uses of the enc script, e.g. standalone fact upload?

It'll appear on stderr but I don't think that would cause a problem. If anything I hope it'd be useful - e.g. people running node.rb from cron should then see errors on stderr.

aptituz:
> No, I don't think it will capture it at all. In my tests of the existing logging, Puppet will just ignore 
> content on stderr and only read from stdout, which is intentional here. It will though be useful for 
> people running the script directly to later find out why facts aren't updating to see the errors.

Yep, that matches my observation, too.

I have updated (amended) the patch to include a message when fact upload fails. I've not added a message for failing ENC lookup, though, but I think the current patch should already be an enhancement and the second change can happen sometime else.

files/external_node_v2.rb
```diff
@@ -342,22 +345,24 @@ def watch_and_send_facts(parallel)
...
-      if SETTINGS[:facts]
-        req = generate_fact_request certname, "#{puppetdir}/yaml/facts/#{certname}.yaml"
-        upload_facts(certname, req)
-      end
+
       #
       # query External node
       begin
         result = ""
         Timeout.timeout(tsecs) do
```
domcleal: I'm concerned about putting the fact upload into this timeout block without increasing the default timeout, as Foreman responds to fact uploads very slowly. I'd possibly prefer leaving it outside - between `begin` and `Timeout.timeout`?

domcleal: Timing out isn't wrong, but I thought the timeout was much shorter than it actually was - 60 seconds is the default, so I think this'll be OK for a fact upload and ENC retrieval.

mmoll:
merged, danke @aptituz!

PR #495, opened 2016-11-01
Add Archlinux packages to provider::params

GregSutcliffe:
I'm working on adding Archlinux to puppet-foreman_proxy, and the tests need to load some package names from here. The packages exist (https://aur.archlinux.org/packages/?O=0&SeB=nd&K=apipie+or+ruby-json+or+ruby-oauth&outdated=&SB=n&SO=a&PP=50&do_Search=Go)

 Archlinux isn't currently in the metadata.json, so I've not touched the tests.

### DIFF
--- a/manifests/params.pp
+++ b/manifests/params.pp
@@ -150,9 +150,14 @@ class foreman::params {
         }
       }
     }
-    /(ArchLinux|Suse)/: {
+    'Suse': {
       # Only the agent classes (cron / service) are supported for now, which
       # doesn't require any OS-specific params
+    }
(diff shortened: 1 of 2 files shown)


mmoll:
As preparation for potentially supporting puppet master setups with puppet-puppet, you could also add the three params needed in manifests/params.pp.
:+1: otherwise.

GregSutcliffe:
@mmoll good idea, done.

mmoll:
merged, thanks @GregSutcliffe!

PR #498, opened 2016-11-04
Fix false "The system does not seem to be IPA-enrolled" result.

adelton:


manifests/config.pp
```diff
@@ -65,7 +65,7 @@
     class { '::foreman::config::passenger': } -> anchor { 'foreman::config_end': }
 
     if $::foreman::ipa_authentication {
-      if !defined('$::default_ipa_server') or empty($::default_ipa_server) or !$::default_ipa_realm or empty($::default_ipa_realm) {
+      if !defined('$default_ipa_server') or empty($::default_ipa_server) or !defined('$default_ipa_realm') or empty($::default_ipa_realm) {
```
mmoll: as that is a fact `::` should stay, no?

ekohl: I assumed that @adelton tested and verified this works. This is ```defined``` and I'm not surprised that it works different from regular scoping. They are in quotes so it must have its own lookup code.

adelton: Things don't work with `::`, I have no idea why. But @ares suggested I drop them and things started to work for me, with Foreman develop.

PR #501, opened 2016-11-16
get environment fact value from puppet node file 

maduma:
form source  $puppetdir/yaml/node/server.example.com

Usefull if we just want to send fact with --push-facts option.
Otherwise the environment is the default production on foreman

### DIFF
--- a/files/external_node_v2.rb
+++ b/files/external_node_v2.rb
@@ -118,7 +118,7 @@ def build_body(certname,filename)
   
   # if there is no environment in facts
   # get it from node file ({puppetdir}/yaml/node/
-  unless puppet_facts.key?('environment')
+  unless puppet_facts['values'].key?('environment')
     node_filename = filename.sub('/facts/', '/node/')
     if File.exist?(node_filename)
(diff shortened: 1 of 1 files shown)


domcleal:
I think this is OK, it's certainly an improvement :+1:

The only issue is that the node YAML file isn't written until _after_ the ENC has been called, so the environment value will always update on the _subsequent_ run causing a lag. That seems unavoidable though.

mmoll:
is the environment still included there without a custom environment fact in Puppet 3+?

maduma:
I did not create a custom fact. just edit the agent config file (Puppet v3.6.2)

```
[agent]
environment = solaris113
logdest = /var/log/puppet/puppet-agent.log
server = newpuppet.svr.luxair

```
For the lag 'issue' we can only enable the feature only for --push-facts

mmoll:
I didn't test this, but it looks OK and I guess @maduma has this running.

maduma:
I confirm that i have it running

domcleal:
Merged, thanks @maduma.

PR #503, opened 2016-11-23
Fix changes to config not restarting service

### LINKED ISSUE #502: Changes to `email.yaml` don't always restart foreman
I updated the various smtp parameters, and ran puppet.  The email.yaml file was updated with the new config, but the foreman service wasn't restarted.

From init.pp
```puppet
Class['foreman::repo'] ~>
class { '::foreman::install': } ~>
class { '::foreman::config': } ~>
class { '::foreman::database': } ~>
class { '::foreman::service': } ->
Class['foreman'] ->
Foreman_smartproxy <| base_url == $foreman_url |>
```
`foreman::config` notifies `foreman::database` but in my case I have `db_manage => false`, so no resources under `foreman::database` are actually refreshed and as a result `foreman::service` doesn't get notified.  At least that what I *think* is happening!

I guess this would be easily fixed by adding an extra relationship. ie
`Class['foreman::config'] ~> Class['foreman::service']`

I'll do some testing and put forward a PR.

alexjfisher:
When `db_manage` and `db_manage_rake` are both set to `false`,
there are no resources in `Class['foreman::database']` that get
refreshed when configuration changes. As a result,
`Class['foreman::service']` doesn't get notified and foreman isn't
restarted.

This commit adds an explicit relationship between
`Class['foreman::config']` and `Class['foreman::service']`.

Fixes #GH-502

### DIFF
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -350,6 +350,9 @@ class foreman (
   Class['foreman'] ->
   Foreman_smartproxy <| base_url == $foreman_url |>
 
+  # When db_manage and db_manage_rake are false, this extra relationship is required.
+  Class['foreman::config'] ~> Class['foreman::service']
+
   # Anchor these separately so as not to break
   # the notify between main classes
   Class['foreman::install'] ~>
--- a/spec/classes/foreman_spec.rb
+++ b/spec/classes/foreman_spec.rb
@@ -8,7 +8,11 @@ describe 'foreman' do
       it { is_expected.to compile.with_all_deps }
       it { should contain_class('foreman::repo').that_notifies('Class[foreman::install]') }
       it { should contain_class('foreman::install') }
-      it { should contain_class('foreman::config') }
+      it do
+        is_expected.to contain_class('foreman::config').that_notifies(
+          ['Class[foreman::database]', 'Class[foreman::service]']
+        )
+      end
       it { should contain_class('foreman::database') }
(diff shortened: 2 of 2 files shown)


domcleal:
1. Can you add an example to spec/foreman_spec.rb? Probably stating that foreman::config notifies foreman::service.
2. Regarding the notification from @theforeman-bot, refer to the ticket as "GH-502" so our two ticketing systems aren't confused.

alexjfisher:
@domcleal I've updated one of the existing examples instead of creating a new one.  Hope that works for you.

alexjfisher:
Commit message title fixed.  It's not under 50 chars anymore, but that becomes quite tricky when the message also has to start with the issue number in a particular format! :)

domcleal:
Sorry! I meant the "Fixes #GH-502" at the end of the commit message should have been "Fixes GH-502". I'm not too worried about the beginning, but this'll work fine - as long as "Fixes GH-*" is in there somewhere, GitHub will close the ticket.

Prefixing the commit message with "Fixes" is more for Redmine tickets. Sorry for the confusion.

alexjfisher:
> Sorry for the confusion.

@domcleal - No worries.  I spend most of my life confused lately! :)
Anyway, I fixed up the commit again and build is green.

mmoll:
merged, thanks @alexjfisher!

PR #504, opened 2016-11-23
Fixes #17467 - Configure tasks with core

iNecas:


iNecas:
Closed due to https://github.com/theforeman/foreman/pull/4217

PR #506, opened 2016-11-24
Change existing Kafo type definitions to Puppet 4 types

domcleal:
Supported since Kafo 1.0.0, it performs more precise validation of
parameter values and can later be moved to the parameter definitions
when the module is made Puppet 4-only.


ekohl:
Thanks!

PR #508, opened 2016-11-29
Support configuring email settings via foreman_config_entry

domcleal:
Foreman 1.14+ stores email settings in its database when email.yaml
is absent. This can be configured by setting `email_config_method`
to 'database', defaulting to the file method for 1.13 compatibility.

---

Support in Foreman not quite merged, but likely to be shortly (http://projects.theforeman.org/issues/12156).

I'll also change the default value to `database` in foreman-installer's develop branch.


ekohl:
Thanks!

PR #510, opened 2016-12-06
Prepare 7.0.0 changelog

domcleal:
[ci skip]


domcleal:
> Some are GitHub IDs where some are redmine issues. Maybe use GH-### and #redmine?

Thanks, fixed - preview releases used GH- as well, now consistent.

PR #511, opened 2016-12-15
Disable the docroot management

binford2k:
Rather than a weird and troublesome workaround, just tell `apache::vhost` not to manage the `docroot`.


domcleal:
Could you resubmit the patch against master please? It seems applicable there rather than 6.0-stable. You should be able to edit the base branch via the GitHub UI.

PR #514, opened 2016-12-27
Prevent Apache to manage the docroot for foreman-ssl

peimanja:


ekohl:
Looks like our tests should be adjusted for this as well.

spec/classes/foreman_config_passenger_spec.rb
```diff
@@ -81,10 +81,6 @@
           http_dir = '/etc/apache2'
         end
 
-        it 'should contain the docroot' do
-          should contain_file("#{params[:app_root]}/public")
```
ekohl: Would you mind changing this to ``should_not```? That would be the proper test now.

spec/classes/foreman_config_passenger_spec.rb
```diff
@@ -81,10 +81,10 @@
...
+        it 'should not contain the docroot' do
+          should_not contain_file("#{params[:app_root]}/public")
         end
-
+
```
mmoll: trailing whitespace here

mmoll:
While we can do that in merge, removing the trailing whitespace and squashing together the commits into one would be nice. 👍 otherwise.

ekohl:
Thanks!

PR #515, opened 2016-12-29
don't provide defaults in the docstrings of parameters

evgeni:
Kafo will already do that for us, and having defaults also in the docstring renders the output duplicated:

    --foreman-manage-user         Controls whether foreman module will manage the user on the system. (default true) (default: true)

### DIFF
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -29,7 +29,7 @@
 # $email_smtp_address::         SMTP server hostname, when delivery method is SMTP
 #                               type:Optional[String]
 #
-# $email_smtp_port::            SMTP port, defaults to 25
+# $email_smtp_port::            SMTP port
 #                               type:Integer[0, 65535]
 #
 # $email_smtp_domain::          SMTP HELO domain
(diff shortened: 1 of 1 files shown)


manifests/init.pp
```diff
@@ -5,7 +5,7 @@
 # $admin_username::             Username for the initial admin user
 #                               type:String
 #
-# $admin_password::             Password of the initial admin user, default is randomly generated
```
ekohl: I think this is still valuable so users know that it's just for their installation.

manifests/init.pp
```diff
@@ -147,7 +147,7 @@
 # $db_username::                Database 'production' user (e.g. foreman)
 #                               type:Optional[String]
 #
-# $db_password::                Database 'production' password (default is random)
```
ekohl: Same as above

evgeni:
Right, thanks.
Adjusted both strings to say "default is randomly generated".

ekohl:
Thanks!

PR #516, opened 2017-01-06
Use apache mod classes

ekohl:
Requires https://github.com/puppetlabs/puppetlabs-apache/pull/1588

### DIFF
--- a/manifests/config/passenger.pp
+++ b/manifests/config/passenger.pp
@@ -44,6 +44,9 @@
 #
 # $start_timeout::            Amount of seconds to wait for Ruby application boot.
 #
+# $foreman_url::              The URL Foreman should be reachable under. Used for loading the application
+#                             on startup rather than on demand.
+#
 # $keepalive::                Enable KeepAlive setting of Apache?
 #
 # $max_keepalive_requests::   MaxKeepAliveRequests setting of Apache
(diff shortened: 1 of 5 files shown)


ekohl:
The pl-apache PR has been merged and rebased. Do we want to wait for a pl-apache release though?

manifests/config/passenger.pp
```diff
@@ -95,6 +95,13 @@
   include ::apache::mod::headers
   include ::apache::mod::passenger
 
+  if $::foreman::ipa_authentication {
```
domcleal: This should be added to the local class parameters list so it can be set in foreman::config::passenger tests.

ekohl: Updated with tests

spec/classes/foreman_config_passenger_spec.rb
```diff
@@ -314,6 +319,41 @@
           should contain_apache__vhost('foreman-ssl').with_passenger_pre_start("https://#{facts[:fqdn]}:8443")
         end
       end
+
+      describe 'with different ports set' do
```
domcleal: `with ipa_authentication => true` ?

ekohl: It's still early. Fixed

ekohl:
Rebased now that puppetlabs-apache 2.0.0 is out.

mmoll:
merged, bedankt @ekohl!

PR #517, opened 2017-01-07
Clean up specs

ekohl:
I'm not sure what's the best solution for the shared examples and modulesync. I think puppetlabs has some optional local spec helper they include in spec_helper.rb so they can still use modulesync, https://relishapp.com/rspec/rspec-core/docs/example-groups/shared-examples suggests placing them in spec/support. It'd nice to have some common convention because puppet-foreman_proxy has the same structure with plugins.


domcleal:
> I think puppetlabs has some optional local spec helper they include in spec_helper.rb so they can still use modulesync

The spec_helper.rb template has an `extra_code` variable, but it would be better in other files.

> https://relishapp.com/rspec/rspec-core/docs/example-groups/shared-examples suggests placing them in spec/support

Sounds fine to me - just add a require into spec_helper.rb in modulesync (or perhaps extra_code).

ekohl:
Updated, implemented through ```extra_code```

domcleal:
Merged, thanks @ekohl.

PR #519, opened 2017-01-13
Puppetdb4 support

netman2k:
It is related to this
https://github.com/theforeman/puppetdb_foreman/commit/57f782c5502b4bdc7afcd56436bb303e182ddb45

Recently, I modified codes on the Puppetdb plugin for supporting PuppetDB 4.x.
But the foreman-installer does not reflect these changes.

I added few params on the puppetdb.pp file. these are all related to the client SSL certificate.

Please apply this request.

### DIFF
--- a/manifests/plugin/puppetdb.pp
+++ b/manifests/plugin/puppetdb.pp
@@ -4,19 +4,35 @@
 #
 # === Parameters:
 #
-# $package:: Package name to install, use ruby193-rubygem-puppetdb_foreman on Foreman 1.8/1.9 on EL
-#            type:String
+# $package::           Package name to install, use ruby193-rubygem-puppetdb_foreman on Foreman 1.8/1.9 on EL
+#                      type:String
 #
-# $address:: Address of puppetdb API. Defaults to 'https://localhost:8081/v2/commands'
-#            type:Stdlib::HTTPUrl
-#
-# $dashboard_address:: Address of puppetdb dashboard. Defaults to 'http://localhost:8080/dashboard'
+# $address::           Address of puppetdb API.
+#                      Defaults to 'https://localhost:8081/pdb/cmd/v1'
 #                      type:Stdlib::HTTPUrl
 #
+# $dashboard_address:: Address of puppetdb dashboard.
+#                      Defaults to 'http://localhost:8080/pdb/dashboard'
+#                      type:Stdlib::HTTPUrl
+#
+# $ssl_ca_file::       CA certificate file which will be used to connect to the PuppetDB API.
+#                      Defaults to client_ssl_ca
+#                      type:String
+#
+# $ssl_certificate::   Certificate file which will be used to connect to the PuppetDB API.
+#                      Defaults to client_ssl_cert
+#                      type:String
+#
+# $ssl_private_key::   Private key file which will be used to connect to the PuppetDB API.
+#                      Defaults to client_ssl_key
+#                      type:String
 class foreman::plugin::puppetdb (
   $package           = $foreman::plugin::puppetdb::params::package,
   $address           = $foreman::plugin::puppetdb::params::address,
   $dashboard_address = $foreman::plugin::puppetdb::params::dashboard_address,
+  $ssl_ca_file       = $foreman::plugin::puppetdb::params::ssl_ca_file,
+  $ssl_certificate   = $foreman::plugin::puppetdb::params::ssl_certificate,
+  $ssl_private_key   = $foreman::plugin::puppetdb::params::ssl_private_key,
 ) inherits foreman::plugin::puppetdb::params {
 
   validate_string($package, $address, $dashboard_address)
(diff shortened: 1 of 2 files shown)


manifests/plugin/puppetdb.pp
```diff
@@ -13,10 +13,25 @@
 # $dashboard_address:: Address of puppetdb dashboard. Defaults to 'http://localhost:8080/dashboard'
 #                      type:Stdlib::HTTPUrl
 #
+# $ssl_ca_file:: CA certificate file which will be used to connect PuppetDB API.
+#                Defaults to client_ssl_ca
+#                type:String
+#
+# $ssl_certificate:: Certificate file which will be used to connect PuppetDB API.
+#                Defaults to client_ssl_cert
```
domcleal: Indentation should match the previous line. (Line it up with "Certificate".)

manifests/plugin/puppetdb.pp
```diff
@@ -13,10 +13,25 @@
 # $dashboard_address:: Address of puppetdb dashboard. Defaults to 'http://localhost:8080/dashboard'
 #                      type:Stdlib::HTTPUrl
 #
+# $ssl_ca_file:: CA certificate file which will be used to connect PuppetDB API.
```
domcleal: Nitpick, "[..]used to connect _to the_ PuppetDB API", ditto below.

manifests/plugin/puppetdb.pp
```diff
@@ -13,10 +13,25 @@
...
+#                Defaults to client_ssl_ca
+#                type:String
+#
+# $ssl_certificate:: Certificate file which will be used to connect PuppetDB API.
+#                Defaults to client_ssl_cert
+#                type:String
+#
+#
+# $ssl_private_key:: Private key file which will be used to connect PuppetDB API.
+#                Defaults to client_ssl_key
```
domcleal: ditto, line it up

manifests/plugin/puppetdb/params.pp
```diff
@@ -31,6 +31,9 @@
       fail("${::hostname}: puppetdb_foreman does not support osfamily ${::osfamily}")
     }
   }
-  $address = 'https://localhost:8081/v2/commands'
-  $dashboard_address = 'http://localhost:8080/dashboard'
+  $address = 'https://localhost:8081/pdb/cmd/v1'
+  $dashboard_address = 'http://localhost:8080/pdb/dashboard'
+  $ssl_ca_file       = $::foreman::client_ssl_ca
```
domcleal: `$::foreman` can't be used as defaults here, at least with foreman-installer. You could include foreman::params and then use `$::foreman::params::client_ssl_ca` though.

manifests/plugin/puppetdb.pp
```diff
@@ -13,10 +13,25 @@
...
+#                type:String
 class foreman::plugin::puppetdb (
   $package           = $foreman::plugin::puppetdb::params::package,
   $address           = $foreman::plugin::puppetdb::params::address,
   $dashboard_address = $foreman::plugin::puppetdb::params::dashboard_address,
+  $ssl_ca_file       = $foreman::plugin::puppetdb::params::client_ssl_ca,
```
domcleal: The names here don't match, causing test failures (it's called $ssl_ca_file in params, not client_ssl_ca.)

netman2k:
Oh, you are totally right. I would change the code and request it.

manifests/plugin/puppetdb.pp
netman2k: Thanks for correcting my poor English

manifests/plugin/puppetdb/params.pp
netman2k: You're right. thanks

manifests/plugin/puppetdb.pp
```diff
@@ -13,10 +13,25 @@
 # $dashboard_address:: Address of puppetdb dashboard. Defaults to 'http://localhost:8080/dashboard'
```
ekohl: This (and ```$address``` above) is no longer correct. Maybe we should just remove the Defaults to part since the installer will show the default value anyway.

netman2k: I updated this URL with new one.

manifests/plugin/puppetdb/params.pp
```diff
@@ -31,6 +31,9 @@
       fail("${::hostname}: puppetdb_foreman does not support osfamily ${::osfamily}")
     }
   }
-  $address = 'https://localhost:8081/v2/commands'
-  $dashboard_address = 'http://localhost:8080/dashboard'
+  $address = 'https://localhost:8081/pdb/cmd/v1'
+  $dashboard_address = 'http://localhost:8080/pdb/dashboard'
+  $ssl_ca_file       = $::foreman::params::client_ssl_ca
```
domcleal: Also needs an `include foreman::params` in this class to ensure the parameters are defined (evaluation order may vary).

netman2k: Thanks for the advice. I added that you gave me.

I think It is more complicated than modifying the actual ruby code.

mmoll:
merged, thanks @netman2k!

PR #521, opened 2017-01-26
Extend gzip file serving to /public/webpack

domcleal:
With compression-webpack-plugin, precompressed files are also available
in `~foreman/public/webpack/` and may be served if available.


mmoll:
merged, thanks @domcleal!

PR #526, opened 2017-02-01
Remove YAML.load_file stubbing, preset $settings_file

domcleal:
Internally Puppet uses YAML.load_file to load the Hiera config (set to
/dev/null under rspec-puppet), but 4.9.0 uses Pathnames instead of
Strings causing the existing stub for "/dev/null" to fail.

Setting the value of $settings_file prior to evaluating the ENC/report
scripts is more reliable than stubbing the YAML constant.


mmoll:
merged, thanks @domcleal!

PR #528, opened 2017-02-02
Restrict gzip asset serving to known extensions

domcleal:
Prevents responses with incorrect content types being served if
different, unknown types of assets are later added to the directory
(similar to #18371 where SVG was added, if 568f278 was not applied).


mmoll:
merged, thanks @domcleal!

PR #530, opened 2017-03-01
Fixes #18757 - Move foreman-tasks service to Foreman

dLobatog:
As per #18618, the dynflow executor is moving over to Foreman, in order
to support ActiveJob. Tasks will provide an UI and persistence layer for
Dynflow, but Foreman should be able to run jobs without it too.

Part of this effort includes moving the Dynflow executor over to
Foreman, which at the moment is managed by the foreman-tasks service.

https://github.com/theforeman/foreman/pull/4316 has to be merged before this

### DIFF
--- a/README.md
+++ b/README.md
@@ -51,6 +51,7 @@ previous stable release.
 
 ### Foreman version compatibility notes
 
+For Foreman 1.16 or older, please set the 'dynflow_in_core' parameter to false.
 For Foreman 1.11 or older, please use the 5.x release series of this module.
 
 ## Types and providers
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -201,6 +201,9 @@
 #
 # $dynflow_pool_size::          How many workers should Dynflow use
 #
+# $jobs_service::               Name of the service for running the background Dynflow executor
+#
+# $dynflow_in_core::            Whether the Dynflow executor service is provided by Foreman or tasks
 class foreman (
   Stdlib::HTTPUrl $foreman_url = $::foreman::params::foreman_url,
   Boolean $puppetrun = $::foreman::params::puppetrun,
@@ -295,6 +298,8 @@ class foreman (
   Optional[String] $email_smtp_user_name = $::foreman::params::email_smtp_user_name,
   Optional[String] $email_smtp_password = $::foreman::params::email_smtp_password,
   Integer[0, 65535] $dynflow_pool_size = $::foreman::params::dynflow_pool_size,
+  String $jobs_service = $::foreman::params::jobs_service,
+  Boolean $dynflow_in_core = $::foreman::params::dynflow_in_core,
 ) inherits foreman::params {
   if $db_adapter == 'UNSET' {
     $db_adapter_real = $::foreman::db_type ? {
--- a/manifests/params.pp
+++ b/manifests/params.pp
@@ -94,6 +94,8 @@ class foreman::params {
 
   # Configure how many workers should Dynflow use
   $dynflow_pool_size = 5
+  # Defines whether Foreman or the tasks plugin provides the Dynflow executor
+  $dynflow_in_core = true
 
   # OS specific paths
   case $::osfamily {
@@ -102,6 +104,8 @@ class foreman::params {
       $init_config_tmpl = 'foreman.sysconfig'
       $puppet_etcdir = '/etc/puppet'
       $puppet_home = '/var/lib/puppet'
+      $jobs_service = $dynflow_in_core ? { true => 'dynflow-executor',
+                                            false => 'foreman-tasks' }
 
       case $::operatingsystem {
         'fedora': {
@@ -134,6 +138,9 @@ class foreman::params {
       $plugin_prefix = 'ruby-foreman-'
       $init_config = '/etc/default/foreman'
       $init_config_tmpl = 'foreman.default'
+      $jobs_service = $dynflow_in_core ? { true => 'ruby-dynflow-executor',
+                                            false => 'ruby-foreman-tasks' }
+
     }
     'Linux': {
       case $::operatingsystem {
@@ -257,4 +264,7 @@ class foreman::params {
   $server_port     = 80
   $server_ssl_port = 443
 
+  # Define job processing service properties
+  $jobs_service_ensure = 'running'
+  $jobs_service_enable = true
 }
--- a/manifests/plugin/ansible.pp
+++ b/manifests/plugin/ansible.pp
@@ -1,9 +1,8 @@
 # Installs foreman_ansible plugin
 class foreman::plugin::ansible {
-
   include ::foreman::plugin::tasks
 
   foreman::plugin {'ansible':
-    notify => Service['foreman-tasks'],
+    notify => Class['foreman::service::jobs'],
   }
 }
--- a/manifests/plugin/remote_execution.pp
+++ b/manifests/plugin/remote_execution.pp
@@ -1,9 +1,8 @@
 # Installs foreman_remote_execution plugin
 class foreman::plugin::remote_execution {
-
   include ::foreman::plugin::tasks
 
   foreman::plugin {'remote_execution':
-    notify => Service['foreman-tasks'],
+    notify => Class['foreman::service::jobs'],
   }
 }
--- a/manifests/plugin/tasks.pp
+++ b/manifests/plugin/tasks.pp
@@ -6,25 +6,20 @@
 #
 # $package:: Package name to install, use ruby193-rubygem-foreman-tasks on Foreman 1.8/1.9 on EL
 #
-# $service:: Service name
-#
 # $automatic_cleanup:: Enable automatic task cleanup using a cron job
 #
 # $cron_line:: Cron line defining when the cleanup cron job should run
 #
 class foreman::plugin::tasks(
   String $package = $::foreman::plugin::tasks::params::package,
-  String $service = $::foreman::plugin::tasks::params::service,
   Boolean $automatic_cleanup = $::foreman::plugin::tasks::params::automatic_cleanup,
   String $cron_line = $::foreman::plugin::tasks::params::cron_line,
 ) inherits foreman::plugin::tasks::params {
+  include ::foreman::service::jobs
+
   foreman::plugin { 'tasks':
     package => $package,
-  }
-  ~> service { 'foreman-tasks':
-    ensure => running,
-    enable => true,
-    name   => $service,
+    notify  => Class['foreman::service::jobs'],
   }
   $cron_state = $automatic_cleanup ? {
     true    => 'present',
--- a/manifests/service.pp
+++ b/manifests/service.pp
@@ -5,6 +5,7 @@ class foreman::service(
   $ssl       = $::foreman::ssl,
 ) {
   anchor { ['foreman::service_begin', 'foreman::service_end']: }
+  contain ::foreman::service::jobs
 
   if $passenger {
     exec {'restart_foreman':
--- a/manifests/service/jobs.pp
+++ b/manifests/service/jobs.pp
@@ -1,0 +1,35 @@
+# = Jobs service
+#
+# Service to start a Dynflow executor for background job
+# processing in Foreman.
+#
+# On foreman 1.16+ this runs the dynflow-executor. This class could be
+# renamed to dynflow-executor when this module has to be compatible with
+# 1.16 and 1.17 only
+#
+# === Parameters:
+#
+# $service:: Service name
+#
+# $dynflow_in_core:: Whether Foreman has a dynflow executor
+#                    without the foreman-tasks plugin
+#
+# $ensure:: State the service should be in
+#
+# $enable:: Whether to enable the service or not
+
+class foreman::service::jobs(
+  String $service = $::foreman::jobs_service,
+  Boolean $dynflow_in_core = $::foreman::dynflow_in_core,
+  Enum['running', 'stopped'] $ensure  = $::foreman::jobs_service_ensure,
+  Boolean $enable = $::foreman::jobs_service_enable,
+) {
+  if $dynflow_in_core and ($service in ['ruby-dynflow-executor', 'dynflow-executor']) or
+    !$dynflow_in_core and ($service in ['ruby-foreman-tasks', 'foreman-tasks']) {
+    service { $service:
+      ensure => $ensure,
+      enable => $enable,
+      name   => $service,
+    }
+  }
+}
--- a/spec/classes/foreman_service_spec.rb
+++ b/spec/classes/foreman_service_spec.rb
@@ -43,7 +43,14 @@ describe 'foreman::service' do
     end
 
     let :pre_condition do
-      'include ::apache'
+      'include ::apache
+      class {
+      "::foreman::service::jobs":
+        service => "dynflow-executor",
+        dynflow_in_core => true,
+        ensure => "running",
+        enable => true,
+      }'
     end
 
     it { is_expected.to compile.with_all_deps }
@@ -80,12 +87,26 @@ describe 'foreman::service' do
   end
 
   context 'without passenger' do
+    let :facts do
+      on_supported_os['redhat-7-x86_64']
+    end
+
     let :params do
       {
         :passenger => false,
         :app_root  => '/usr/share/foreman',
         :ssl => true,
       }
+    end
+
+    let :pre_condition do
+      "class {
+      'foreman::service::jobs':
+        service => 'dynflow-executor',
+        dynflow_in_core => true,
+        ensure => 'running',
+        enable => true,
+      }"
     end
 
     it { is_expected.to compile.with_all_deps }
--- a/spec/classes/plugin/remote_execution_spec.rb
+++ b/spec/classes/plugin/remote_execution_spec.rb
@@ -7,7 +7,12 @@ describe 'foreman::plugin::remote_execution' do
       let(:pre_condition) { 'include foreman' }
 
       it { should compile.with_all_deps }
-      it { should contain_foreman__plugin('remote_execution').that_notifies('Service[foreman-tasks]') }
+      it { should contain_foreman__plugin('remote_execution').that_notifies("Class[foreman::service::jobs]") }
+    end
+
+    context "in 1.16 or lower, tasks is included" do
+      let(:facts) { facts }
+      let(:pre_condition) { 'include foreman' }
       it { should contain_foreman__plugin('tasks') }
     end
   end
--- a/manifests/plugin/tasks/params.pp
+++ b/manifests/plugin/tasks/params.pp
@@ -4,7 +4,6 @@ class foreman::plugin::tasks::params {
   $cron_line = '45 19 * * *'
   case $::osfamily {
     'RedHat': {
-      $service = 'foreman-tasks'
       case $::operatingsystem {
         'fedora': {
           $package = 'rubygem-foreman-tasks'
@@ -16,7 +15,6 @@ class foreman::plugin::tasks::params {
     }
     'Debian': {
       $package = 'ruby-foreman-tasks'
-      $service = 'ruby-foreman-tasks'
     }
     /^(FreeBSD|DragonFly)$/: {
       # do nothing to not break foreman-installer
--- a/spec/classes/plugin/tasks_spec.rb
+++ b/spec/classes/plugin/tasks_spec.rb
@@ -17,20 +17,16 @@ describe 'foreman::plugin::tasks' do
                        else
                          'tfm-rubygem-foreman-tasks'
                        end
-        service_name = 'foreman-tasks'
       when 'Debian'
         package_name = 'ruby-foreman-tasks'
-        service_name = 'ruby-foreman-tasks'
       else
         package_name = 'foreman-tasks'
-        service_name = 'foreman-tasks'
       end
 
       it { should compile.with_all_deps }
 
       it 'should call the plugin' do
         should contain_foreman__plugin('tasks').with_package(package_name)
-        should contain_service('foreman-tasks').with('ensure' => 'running', 'enable' => 'true', 'name' => service_name)
         should contain_file('/etc/cron.d/foreman-tasks').
           with_path('/etc/cron.d/foreman-tasks').
           with_ensure('absent')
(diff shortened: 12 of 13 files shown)


ekohl:
Any reason it has to live in ```tasks_service``` rather than ```service```? I'm guessing that the service needs a restart if any settings change which it currently won't.

Other than that we generally keep compatibility with older Foreman versions (current and current - 1 is what we at least aim for) or bump the major version. The current implementation breaks on 1.14 without foreman-tasks so it needs to deal with this.

manifests/tasks_service.pp
```diff
@@ -0,0 +1,20 @@
...
+# === Parameters:
+#
+# $service:: Service name
+#            type:String
+
+class foreman::tasks_service(
+  $service = $::foreman::params::tasks_service,
+) inherits foreman::params {
+  service { 'foreman-tasks':
+    ensure => running,
```
timogoebel: I'd suggest pulling the variables out so you can overwrite them if you want.
Just move them to the "signature" of the class, similar to service.

dLobatog:
@ekohl Actually no need to restart the tasks service AFAIK - that's not a problem.  I see a small problem though if foreman 1.14 has to be supported. The situation is the following:

On 1.14:
* foreman-tasks ships the foreman-tasks service. This service starts the dynflow executor. 
* foreman 'core' does not ship the foreman-tasks service. it does not have anything to do with dynflow

On 1.15:
* foreman-tasks does not ship the foreman-tasks service. It needs foreman to run the dynflow executor
* foreman 'core' ships 'foreman-tasks' to start the dynflow executor. 

To keep this module compatible with both versions it would require:
* having a generic 'foreman-tasks' that can be triggered from either the foreman-tasks or the foreman class https://github.com/theforeman/puppet-foreman/pull/530/files#diff-bb391d4e28641499e1a968d496bde646R2
* on 1.15: trigger foreman-tasks service from foreman
* on 1.14: trigger foreman-tasks service from foreman-tasks
So it'd require some kind of `if version then'

We thought about changing the `foreman-tasks` service to a "fake" one that simply calls 'foreman-jobs' (a renamed foreman-tasks task in Foreman core) for everything. That would mean:

* Keeping the foreman-tasks module as-is
* On 1.15: foreman-jobs is started automatically with the foreman service.

dLobatog:
@mmoll What you suggested was having a parameter `dynflow_in_core = true`  that triggers the foreman-tasks service from foreman or foreman-tasks depending on its value, correct?

mmoll:
@dLobatog exactly.

dLobatog:
Updated to reflect the changes asked for @ekohl @mmoll & @timogoebel  - I have to add tests for this `dynflow_in_core` option to ensure it works properly. I suppose I'd have to add `dynflow_in_core: true` to the default options in the installer if https://github.com/theforeman/foreman/pull/4316 gets merged on time

manifests/plugin/tasks.pp
```diff
@@ -7,19 +7,17 @@
...
-#            type:String
+# $dynflow_in_core:: Whether Foreman ships the Dynflow executor service (true on 1.15+)
+#                    type:String
 #
 class foreman::plugin::tasks(
   $package = $foreman::plugin::tasks::params::package,
-  $service = $foreman::plugin::tasks::params::service,
+  $dynflow_in_core = $::foreman::dynflow_in_core,
 ) inherits foreman::plugin::tasks::params {
+  if $dynflow_in_core == false {
```
ohadlevy: nitpick, cant you just use
```puppet
unless $dynflow_in_core {
```

manifests/params.pp
```diff
@@ -99,6 +99,7 @@
       $init_config_tmpl = 'foreman.sysconfig'
       $puppet_etcdir = '/etc/puppet'
       $puppet_home = '/var/lib/puppet'
+      $tasks_service = 'foreman-tasks'
```
ohadlevy: IMHO we should not call it foreman-tasks

manifests/tasks_service.pp
```diff
@@ -0,0 +1,28 @@
+# = foreman-tasks service
```
ohadlevy: nitpick about the name, imho dynflow or background-jobs makes sense not to confuse with task plugin.

manifests/tasks_service.pp
```diff
@@ -0,0 +1,28 @@
...
+#
+# $enable:: Whether to enable the service or not
+#           type:String
+
+class foreman::tasks_service(
+  $service = $::foreman::params::tasks_service,
+  $ensure  = $::foreman::params::tasks_service_ensure,
+  $enable  = $::foreman::params::tasks_service_enable,
+) inherits foreman::params {
+  service { 'foreman-tasks':
```
ohadlevy: nitpick: use the parameter?

mmoll:
please check the test errors

manifests/plugin/tasks.pp
```diff
@@ -7,19 +7,17 @@
...
 ) inherits foreman::plugin::tasks::params {
   foreman::plugin { 'tasks':
     package => $package,
-  } ~>
-  service { 'foreman-tasks':
-    ensure => running,
-    enable => true,
-    name   => $service,
+  }
+  unless $dynflow_in_core {
```
ekohl: I'm not sure this parameter is actually needed. Given there's an include in ```foreman::service``` if it's true that means it's going to be included anyway.

You're also losing the notify on the service if the plugin changes.

spec/classes/foreman_tasks_service_spec.rb
```diff
@@ -0,0 +1,13 @@
+require 'spec_helper'
+
+describe 'foreman::tasks_service' do
+  let :facts do
+    on_supported_os['redhat-7-x86_64']
```
ekohl: Given this code is relevant for every distro where we run Foreman, I'd feel more comfortable if this was tested on more platforms. We've already seen that there are subtle problems with dependencies if we take these shortcuts.

manifests/plugin/tasks.pp
dLobatog: This one is precisely for the opposite thing - if it's false, I don't want to include the tasks service as it's not necessary to restart it when foreman-tasks was installed.

ekohl: My point is: if it's true, it's included anyway (through ```foreman::service```). Here you also ensure it's included if it's false. Doesn't that mean you can unconditionally include it?

dLobatog:
@ekohl @mmoll mind to double check? I think I've implemented the changes you asked for. Thanks!

manifests/plugin/remote_execution.pp
```diff
@@ -1,9 +1,21 @@
...
-  include ::foreman::plugin::tasks
+  if $dynflow_in_core {
+    include ::foreman::tasks_service
+  } else {
+    include ::foreman::plugin::tasks
+  }
 
   foreman::plugin {'remote_execution':
-    notify => Service['foreman-tasks'],
+    notify => Service[$::foreman::params::tasks_service],
```
mmoll: what happens if that parameter is overridden?

mmoll:
what's the name of service? it'll come with the RPM/DEB package, correct?

dLobatog:
@mmoll It's still called foreman-tasks, distributed with the package - https://github.com/theforeman/foreman-packaging/pull/1549

manifests/tasks_service.pp
```diff
@@ -0,0 +1,28 @@
...
+#            type:String
+#
+# $ensure:: State the service should be in
+#           type:Enum['absent', 'present']
+#
+# $enable:: Whether to enable the service or not
+#           type:String
+
+class foreman::tasks_service(
+  $service = $::foreman::params::tasks_service,
```
dLobatog: I got kind of stuck here - if I remove the params and use `$::foreman::tasks_service`, the spec for `foreman_service_spec` fails - because that spec tests when `$::foreman`  is not set - and this manifest needs it

ekohl: You can either explicitly override it in the test or add a ```let :pre_condition do 'include ::foreman' end```. The former leads to faster tests which I'd slightly prefer.

manifests/plugin/tasks.pp
```diff
@@ -7,19 +7,10 @@
...
-# $service:: Service name
-#            type:String
-#
 class foreman::plugin::tasks(
   $package = $foreman::plugin::tasks::params::package,
-  $service = $foreman::plugin::tasks::params::service,
 ) inherits foreman::plugin::tasks::params {
   foreman::plugin { 'tasks':
     package => $package,
-  } ~>
```
ekohl: I still think this is incorrect. It should include the service or else you break 1.14 installations.

dLobatog:
@ekohl Yes, it will stay a package, but no longer have its own service. The foreman-tasks service manages the dynflow executor, and all this work is to move the dynflow executor management to foreman. 

foreman-tasks will remain a plugin, as Dynflow by itself does not have an UI or data structure to show the tasks.

ekohl:
@dLobatog there's a merge conflict now. Given the main foreman PR has not yet been merged, can I assume this won't be ready in time for 1.15 and we don't need to include this in the next release?

ekohl:
So with the 1.16 branching approaching I'm wondering what the plan is with this.

manifests/plugin/remote_execution.pp
```diff
@@ -1,9 +1,21 @@
...
-  include ::foreman::plugin::tasks
+  if $dynflow_in_core {
+    include ::foreman::tasks_service
+  } else {
+    include ::foreman::plugin::tasks
+  }
 
   foreman::plugin {'remote_execution':
-    notify => Service['foreman-tasks'],
+    notify => Service[$::foreman::tasks_service],
```
ekohl: You could just notify ```Class['foreman::tasks_service']```.

manifests/plugin/remote_execution.pp
```diff
@@ -1,9 +1,21 @@
 # Installs foreman_remote_execution plugin
-class foreman::plugin::remote_execution {
+#
+# === Parameters:
+#
+# $dynflow_in_core:: Whether Foreman ships the Dynflow executor service (true on 1.15+)
+#                    type:Boolean
+#
+class foreman::plugin::remote_execution(
+  $dynflow_in_core = $::foreman::dynflow_in_core
```
ekohl: We use trailing commas

manifests/plugin/remote_execution.pp
```diff
@@ -1,9 +1,21 @@
 # Installs foreman_remote_execution plugin
-class foreman::plugin::remote_execution {
+#
+# === Parameters:
+#
+# $dynflow_in_core:: Whether Foreman ships the Dynflow executor service (true on 1.15+)
+#                    type:Boolean
```
ekohl: This is the old style. When you want to enforce the type you can now use Puppet 4 types in the class definition: ```Boolean $dynflow_in_core = $::foreman::dynflow_in_core```. Then Puppet will actually check the type instead of relying on Kafo to do it instead.

manifests/plugin/remote_execution.pp
```diff
@@ -1,9 +1,21 @@
 # Installs foreman_remote_execution plugin
-class foreman::plugin::remote_execution {
+#
+# === Parameters:
```
ekohl: I think this can be ```Advanced``` so ```foreman-installer --help``` doesn't show it. It's rather internal and I'd expect no foreman-installer user to ever need this.

manifests/plugin/ansible.pp
```diff
@@ -1,9 +1,16 @@
...
+#                  type:String
+#
+class foreman::plugin::ansible(
+  $tasks_service = $::foreman::tasks_service,
+) {
   include ::foreman::plugin::tasks
 
   foreman::plugin {'ansible':
-    notify => Service['foreman-tasks'],
+    notify => Service[$tasks_service],
```
ekohl: You can notify ```Class['foreman::tasks_service']``` and remove the parameter.

manifests/init.pp
```diff
@@ -197,6 +197,11 @@
 #
 # $email_conf::                 Email configuration file under /etc/foreman
 #
+# $tasks_service::              Name of the service for running the background Dynflow executor
+#                               type:String
```
ekohl: These types can be left out since you're using Puppet 4 types already which means this is redundant.

manifests/init.pp
```diff
@@ -197,6 +197,11 @@
...
+#
+# $dynflow_in_core::            Whether the Dynflow executor service is provided by Foreman or tasks
+#                               type:Boolean
```
ekohl: Also redundant now.

manifests/tasks_service.pp
ekohl: I think I'd prefer to name it ```foreman::service::tasks``` (or jobs).

manifests/tasks_service.pp
```diff
@@ -0,0 +1,40 @@
...
+#
+# $ensure:: State the service should be in
+#           type:Enum['absent', 'present']
+#
+# $enable:: Whether to enable the service or not
+#           type:String
+
+class foreman::tasks_service(
+  $service = $::foreman::tasks_service,
+  $dynflow_in_core= $::foreman::dynflow_in_core,
```
ekohl: nit: missing space before ```=```.

manifests/tasks_service.pp
```diff
@@ -0,0 +1,40 @@
...
+# foreman-tasks plugin.
+#
+# On foreman 1.16+ this runs the dynflow-executor. This class could be
+# renamed to dynflow-executor when this module has to be compatible with
+# 1.16 and 1.17 only
+#
+# === Parameters:
+#
+# $service:: Service name
+#            type:String
```
ekohl: Here you can move the types to Puppet 4 types as well.

manifests/tasks_service.pp
```diff
@@ -0,0 +1,40 @@
...
+#
+# $dynflow_in_core:: Whether Foreman has a dynflow executor
+#                  without the foreman-tasks plugin
+#            type:Boolean
+#
+# $ensure:: State the service should be in
+#           type:Enum['absent', 'present']
```
ekohl: I'd expect this to be ```Enum['running', 'stopped']```.

manifests/tasks_service.pp
```diff
@@ -0,0 +1,40 @@
...
+#
+# $enable:: Whether to enable the service or not
+#           type:String
```
ekohl: Should be a ```Boolean```

spec/classes/foreman_service_spec.rb
```diff
@@ -80,6 +80,10 @@
   end
 
   context 'without passenger' do
+    let :facts do
```
ekohl: I'm surprised this worked before.

spec/classes/plugin/remote_execution_spec.rb
```diff
@@ -7,7 +7,17 @@
       let(:pre_condition) { 'include foreman' }
 
       it { should compile.with_all_deps }
-      it { should contain_foreman__plugin('remote_execution').that_notifies('Service[foreman-tasks]') }
+      if os =~ /redhat/i
+        it { should contain_foreman__plugin('remote_execution').that_notifies("Service[dynflow-executor]") }
```
ekohl: When you change the notification in the class this can be static again.

manifests/params.pp
```diff
@@ -92,13 +92,18 @@
   $email_smtp_user_name      = undef
   $email_smtp_password       = undef
 
+  # Defines whether Foreman or the tasks plugin provides the Dynflow executor
+  $dynflow_in_core = true
```
ekohl: This will mean a major version bump. Not a problem per se, but important we don't forget this.

manifests/init.pp
```diff
@@ -316,6 +323,7 @@
   ~> class { '::foreman::install': }
   ~> class { '::foreman::config': }
   ~> class { '::foreman::database': }
+  ~> class { '::foreman::tasks_service': }
```
ekohl: I think that if you do this inside the tasks class with ```contain ::foreman::tasks_service``` then you can just notify the service class and plugins automatically restart the tasks service as well. I think that's a good thing in general even if it's not always needed, it can save some headache. Especially if more plugins start to use background jobs. It will also mean you can remove the inclusion from plugins that only need the tasks service and not bother with the notification.

ekohl:
Btw, don't be discouraged by the large amount of comments: overall it looks good.

manifests/init.pp
dLobatog: I can certainly turn the include in `plugin/tasks.pp` into a `contain ::foreman::service::jobs`. However here I'm supposed to make it start (and return OK) before starting Foreman itself. Do you want me to move this class declaration to contain too (if so, how?)?

ekohl: I meant that you should remove this line here and add the contain to ```manifests/service.pp```, but wasn't very clear in that.

dLobatog:
Repushing, stalled build in Travis.

dLobatog:
@ekohl @mmoll Could you take another look please 😄 ?

manifests/plugin/remote_execution.pp
```diff
@@ -1,9 +1,9 @@
 # Installs foreman_remote_execution plugin
-class foreman::plugin::remote_execution {
```
ekohl: nit: please keep the old style to keep the diff smaller

manifests/plugin/ansible.pp
```diff
@@ -1,9 +1,8 @@
 # Installs foreman_ansible plugin
-class foreman::plugin::ansible {
```
ekohl: nit: please keep the old style to keep the diff smaller.

manifests/plugin/tasks.pp
```diff
@@ -6,18 +6,11 @@
...
 ) inherits foreman::plugin::tasks::params {
   foreman::plugin { 'tasks':
     package => $package,
   }
-  ~> service { 'foreman-tasks':
-    ensure => running,
-    enable => true,
-    name   => $service,
-  }
+  contain ::foreman::service::jobs
```
ekohl: I think this needs to be an ```include``` and a ```notify => Class['foreman::service::jobs']```

manifests/service.pp
```diff
@@ -5,6 +5,7 @@
   $ssl       = $::foreman::ssl,
 ) {
   anchor { ['foreman::service_begin', 'foreman::service_end']: }
+  include ::foreman::service::jobs
```
ekohl: I think this needs to be ```contain ::foreman::service::jobs``` so you restart the jobs service as well when you ```notify => Class['foreman::service']```.

manifests/params.pp
```diff
@@ -92,13 +92,18 @@
...
+
   # OS specific paths
   case $::osfamily {
     'RedHat': {
       $init_config = '/etc/sysconfig/foreman'
       $init_config_tmpl = 'foreman.sysconfig'
       $puppet_etcdir = '/etc/puppet'
       $puppet_home = '/var/lib/puppet'
+      $jobs_service = $dynflow_in_core ? { true => 'dynflow-executor',
```
ekohl: This is tricky. I get what you want to achieve, but currently this is equal to ```$jobs_service = 'dynflow-executor'``` because ```$dynflow_in_core``` is always true. It doesn't look at init.pp.

I'd recommend to drop the ```dynflow_in_core``` parameter altogether and let users override ```$jobs_service```. Some additional documentation in the README can suffice for this. I think this keeps the API and code simpler while allowing users of 1.15 and 1.16 to use it.

The alternative is to default ```jobs_service``` to ```undef``` and determine it in ```foreman::service::jobs``` if undef. Don't forget to change all types to ```Optional[String]```.

README.md
```diff
@@ -51,6 +51,7 @@ previous stable release.
 
 ### Foreman version compatibility notes
 
+For Foreman 1.15 or older, please set the 'dynflow_in_core' parameter to false.
```
mmoll: 1.16

manifests/service/jobs.pp
```diff
@@ -0,0 +1,35 @@
+# = Jobs service
+#
+# Service to start a Dynflow executorfor background job
```
mmoll: `excutor for`

manifests/service/jobs.pp
```diff
@@ -0,0 +1,35 @@
...
+# processing in Foreman.
+#
+# On foreman 1.16+ this runs the dynflow-executor. This class could be
+# renamed to dynflow-executor when this module has to be compatible with
+# 1.16 and 1.17 only
+#
```
mmoll: version updates needed also here

spec/classes/plugin/remote_execution_spec.rb
```diff
@@ -7,7 +7,12 @@
       let(:pre_condition) { 'include foreman' }
 
       it { should compile.with_all_deps }
-      it { should contain_foreman__plugin('remote_execution').that_notifies('Service[foreman-tasks]') }
+      it { should contain_foreman__plugin('remote_execution').that_notifies("Class[foreman::service::jobs]") }
+    end
+
+    context "in 1.15 or lower, tasks is included" do
```
mmoll: 1.16

mmoll:
looking good, but I first want to have the change done in nightly packages (after 1.16 RC1 would be good, I think).

ekohl:
What's the status of the packaging? Can this be merged?

ekohl:
So now it's only pending https://github.com/theforeman/foreman-packaging/pull/1549?

ares:
@dLobatog sadly this needs another rebase

dLobatog:
Tests 🍏  cc @ekohl @mmoll

README.md
```diff
@@ -51,6 +51,7 @@ previous stable release.
 
 ### Foreman version compatibility notes
 
+For Foreman 1.16 or older, please set the 'dynflow_in_core' parameter to false.
```
ekohl: If we want to get this in 1.16, this should be ```1.15 or older```

manifests/service.pp
```diff
@@ -5,6 +5,7 @@
   $ssl       = $::foreman::ssl,
 ) {
   anchor { ['foreman::service_begin', 'foreman::service_end']: }
+  contain ::foreman::service::jobs
```
ekohl: I think this should be surrounded by ```if $dynflow_in_core {``` and added added to the class parameters.

manifests/params.pp
```diff
@@ -102,6 +104,8 @@
...
```
ekohl: Since here ```$dynflow_in_core``` is always true, you can just write ```$jobs_service = 'dynflow-executor```. If you want to dynamically determine this it should be in the ```foreman::service::jobs``` class. I'd recommend ```$jobs_service = undef``` here and then determine it later.

ekohl:
Closing since the replacement PR was merged.

PR #533, opened 2017-03-11
add SLES AIO defaults

mmoll:
This is only for some theforeman-puppet tests to succeed, as in real life, theforeman-puppet does set these on its own.


ekohl:
Actually theforeman-puppet only sets it in case of AIO, these values would be used in non-AIO configs.

PR #536, opened 2017-03-21
Add param for timeout to foreman() parser function

dae1804:
Add parameter for timeout to foreman() parser function.  If unspecified, keep the default at 5sec.

This is useful if you are using the foreman() parser function to automatically populate templates, and you are using queries that occasionally take more than five seconds to complete.


dae1804:
Don't merge this; my test had an error, and the timeout param isn't properly converted to an integer.

ekohl:
In that case ```timeout.to_i``` might be the fix.

dae1804:
Done, and added a unit test this time.

domcleal:
Merged, thanks @dae1804.

PR #537, opened 2017-03-23
Remove ruby 1.8 pinning

ekohl:
We no longer test on 1.8 so this code is redundant.

### DIFF
--- a/.sync.yml
+++ b/.sync.yml
@@ -7,11 +7,6 @@ Gemfile:
   extra:
     - gem: webmock
       version: '< 2.0'
-    - gem: addressable
-      version: '< 2.4'
-      options:
-        platforms:
-        - 'ruby_18'
     - gem: oauth
 Rakefile:
   param_docs_pattern:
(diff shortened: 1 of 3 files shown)


.sync.yml
```diff
@@ -7,11 +7,6 @@ Gemfile:
   extra:
     - gem: webmock
```
domcleal: Also webmock 2.x can be used for Ruby 1.9 and above.

ekohl: https://github.com/bblimke/webmock only lists MRI 2.0 and higher. Soon we'll drop Puppet 3 support and that means we can drop Ruby 1.9 as well since the PL AIO install includes a ruby 2+ (for better or worse). Should we address webmock then?

domcleal: That's master, but the CHANGELOG for 2.x and the published gems state 1.9.3 (https://github.com/bblimke/webmock/blob/master/CHANGELOG.md). I think that may become 3.x, so the version here could be raised to permit 2.x.

ekohl:
Rebased and added the upgrade to webmock 2.x.

mmoll:
merged, bedankt @ekohl!

Issue #390, opened 2016-01-18
Ship report processor as a module


GregSutcliffe:
According to https://docs.puppetlabs.com/guides/reporting.html#writing-custom-reports it's now expected for reports to live in /lib/puppet/reports within a module on the master to make it available to the master. This is probably a good thing (no more hunting around in /usr for the correct path).

I guess we need to keep the copy in /files for older puppet versions - should we symlink? I don't know how puppet will treat a symlink here, but keeping two actual copies in sync sounds annoying...

domcleal:
This has been possible for quite a while, it should be fine with 3.x (at least).

GregSutcliffe:
Good to know, I'll try to test and get a PR up.

ekohl:
Given we dropped support for 2.7, it should simplify our module.

mmoll:
For now fixed by theforeman/puppet-puppet#353, I suppose? Maybe it makes sense to open a tracker issue to collect stuff that's fixed but should be revistied...

GregSutcliffe:
@mmoll yeah, that could make sense - this fix works for us, but I do like the idea of having a trivial module that people can add to their environment rather than figuring out the path to the reports dir in the puppet codebase.

Happy to close this if we don't think that's going to happen - but I'll leave it open for now to link in to such a tracker.

ekohl:
I don't mind a tracker, but IMHO we should rename it and decouple it from puppet 4 since it's no longer a blocker.

ekohl:
Rephrased the title

jyaworski:
Has there been any progress on this?

ekohl:
Not that I'm aware of, but I think @GregSutcliffe tested it locally and it worked.

GregSutcliffe:
It does work, it's what I do in my own infra to ensure the report processor is available. @jyaworski is it not working with https://github.com/theforeman/puppet-puppet/pull/353? @mmoll seemed to think that would be enough, so I left this alone for a while...

rendhalver:
So do I need to use the foreman puppet module to get this working?
I also have a module called puppet for managing puppet in my environment so that won't work for me.
I was going to prep a PR for getting this setup as a plugin in this module.
Is that acceptable?

ekohl:
[...]

PR #539, opened 2017-03-24
[WIP] Fix running under puppet 4

rendhalver:
Add support for puppet 4
This also moves the report into the plugins so it is synced to the right place.

### DIFF
--- a/manifests/puppetmaster.pp
+++ b/manifests/puppetmaster.pp
@@ -37,16 +37,19 @@ class foreman::puppetmaster (
 
   if $reports {   # foreman reporter
 
-    exec { 'Create Puppet Reports dir':
-      command => "/bin/mkdir -p ${puppet_basedir}/reports",
-      creates => "${puppet_basedir}/reports",
-    }
-    file {"${puppet_basedir}/reports/foreman.rb":
-      mode    => '0644',
-      owner   => 'root',
-      group   => '0',
-      source  => "puppet:///modules/${module_name}/foreman-report_${report_api}.rb",
-      require => Exec['Create Puppet Reports dir'],
+    if ( $::foreman::params::aio_package == false ) {
+      # this should already be in place from pluginsync
+      exec { 'Create Puppet Reports dir':
+        command => "/bin/mkdir -p ${puppet_basedir}/reports",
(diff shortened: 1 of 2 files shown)


manifests/puppetmaster.pp
```diff
@@ -37,16 +37,19 @@
...
-      command => "/bin/mkdir -p ${puppet_basedir}/reports",
-      creates => "${puppet_basedir}/reports",
-    }
-    file {"${puppet_basedir}/reports/foreman.rb":
-      mode    => '0644',
-      owner   => 'root',
-      group   => '0',
-      source  => "puppet:///modules/${module_name}/foreman-report_${report_api}.rb",
-      require => Exec['Create Puppet Reports dir'],
+    if ( $::foreman::params::aio_package == false ) {
```
ekohl: This can be written as ```unless $::foreman::params::aio_package``` but I think it still breaks our use case. Like [I mentioned](https://github.com/theforeman/puppet-foreman/issues/390#issuecomment-289026125), we deploy it normally through puppet-puppet and set the correct ruby path there. In our installer use case we don't actually deploy the puppet module onto the puppet server (we've always stayed away from managing the users puppet environment) so no reporting ends up being installed.

rendhalver: So how do we want to deal with this?
If this get's split out into it's own module this change will not be necessary.
If I migrate this to copy the file into the lib directory then it will be writing a file to a directory that get's synced from the server.

ekohl: I'd prefer users of this call it with the correct parameters. The most logical one would be to set ```$reports``` to false.

ekohl:
Currently you should call it as follows with puppet AIO installs:

```puppet
class {'::foreman::puppetmaster':
  puppet_home    => '/opt/puppetlabs/server/data/puppetserver',
  puppet_basedir => '/opt/puppetlabs/puppet/lib/ruby/vendor_ruby/puppet',
  puppet_etcdir  => '/etc/puppetlabs/puppet',
}
```

Would you be willing to modify the PR to use this and drop the aio_package conditional or should we close this?

mmoll:
closing due to inactivity

PR #540, opened 2017-03-29
Move chaining arrows to right operand line

domcleal:
To match style guide and for compatibility with puppet-lint 2.2.0.


ekohl:
Thanks!

PR #541, opened 2017-03-29
Set ssl_ca_file in Hammer system-wide config file

domcleal:
Defaults when used with main 'foreman' class to the server's CA file,
automatically ensuring Hammer CLI will trust the CA used.

### DIFF
--- a/manifests/cli.pp
+++ b/manifests/cli.pp
@@ -24,6 +24,9 @@
 # $request_timeout::      API request timeout, set -1 for infinity
 #                         type:Integer[-1]
 #
+# $ssl_ca_file::          Path to SSL certificate authority
+#                         type:Optional[Stdlib::Absolutepath]
+#
 # $hammer_plugin_prefix:: Hammer plugin package prefix based normally on platform
 #                         type:String
 #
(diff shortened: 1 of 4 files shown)


manifests/cli.pp
```diff
@@ -49,12 +53,14 @@
...
     $password_real    = pick($password, $::foreman::admin_password)
+    $ssl_ca_file_real = pick($ssl_ca_file, $::foreman::server_ssl_ca)
   } else {
     $foreman_url_real = $foreman_url
     $username_real    = $username
     $password_real    = $password
+    $ssl_ca_file_real = $ssl_ca_file
   }
-  validate_string($foreman_url_real, $username_real, $password_real)
+  validate_string($foreman_url_real, $username_real, $password_real, $ssl_ca_file_real)
```
ekohl: This may not be needed but the documentation states it should be an absolute path and here it's only validated as a string.

domcleal: Thanks, updated to use the correct validation function.

mmoll:
merged, thanks @domcleal!

PR #544, opened 2017-03-31
Workaround puppet server ruby

mrwulf:
While originally working around this (https://tickets.puppetlabs.com/browse/MODULES-4256) puppet server issue, it does seem like this foreman installed script should be using the foreman installed ruby instance. Failing build seems to be related to a new parallels requirement for ruby >= 2.

I cleaned up some "hidden" spaces too...

### DIFF
--- a/files/external_node_v2.rb
+++ b/files/external_node_v2.rb
@@ -1,4 +1,4 @@
-#!/usr/bin/env ruby
+#!/usr/bin/env tfm-ruby
 
 # Script usually acts as an ENC for a single host, with the certname supplied as argument
 #   if 'facts' is true, the YAML facts for the host are uploaded
@@ -115,7 +115,7 @@ def build_body(certname,filename)
   facts        = File.read(filename)
   puppet_facts = YAML::load(facts.gsub(/\!ruby\/object.*$/,''))
   hostname     = puppet_facts['values']['fqdn'] || certname
-  
+
   # if there is no environment in facts
   # get it from node file ({puppetdir}/yaml/node/
   unless puppet_facts['values'].key?('environment')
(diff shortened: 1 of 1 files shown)


mmoll:
tfm-ruby is only there on RedHat boxes with the Ruby from the tfm SCL, so this is not a general solution. I guess what's really needed is that the ENC script uses the same Ruby as the puppetmaster/puppetserver process, correct?

ekohl:
Is this still an issue?

mmoll:
I guess so.

jamoflaw:
Just to note that the following works:

Edit /etc/puppetlabs/puppet/node.rb

Replace the #! with 

#!/opt/puppetlabs/puppet/bin/ruby

Effectively forcing the script to use the included puppet ruby environment. I have tested this in test with no adverse effects so far.

domcleal:
I'd be very hesitant to change this without understanding why. The linked ticket has no information other than the `json` gem is failing to load for some reason, and I'm unable to reproduce it. If the cause was clearer then we could better consider which Ruby shebang it should use.

mrwulf:
I agree that this is a hack at best and would love to get to the root of this problem. @mmoll, I thought tfm-ruby **was** effectively a symlink managed by the installer. I just wanted to point the script (since it's part of foreman) to the "foreman approved" ruby.

@domcleal, I updated the linked ticket with the requested details. Unfortunately I don't have the cycles to track this down myself, but I'll do my best to facilitate.

ekohl:
Any update on this?

mrwulf:
I've had 4  servers (each with hundreds of clients) running this since I posted it and haven't noticed any ill effects. Albeit we only use CentOS and windows...

ekohl:
But we can't really accept this right now and it looks like your system ruby is broken since it can't load json. I wonder if you run ```/etc/puppetlabs/puppet/node.rb``` manually on the shell if you also get the error. When that's the case, you can try a minimal script to reproduce:
```ruby
#!/usr/bin/env ruby
require 'json'
```

You could also verify if the package is still in tact: ```rpm -qV rubygem-json```.

mrwulf:
`node.rb` works fine from the command line- this isn't a problem with my ruby. This is really a problem with a different ruby running inside the puppetserver process with the rabbitmq module.

I understand this PR isn't really a fix for the issue and is just a workaround until [the real issue](https://tickets.puppetlabs.com/browse/MODULES-4256) is fixed. Hopefully puppet v5 will handle conflicting gem versions better.

PR #545, opened 2017-04-02
Added pgsql cluster option

sean797:
This adds the ability to setup a HA pgSQL cluster using Streaming Replication.

On a master, you can do:
`--foreman-db-node-type master --foreman-db-cluster-hostnames node2.example.com --foreman-db-host foreman-vip.example.com`

One a slave, you can do:
`--foreman-db-node-type slave --foreman-db-cluster-hostnames node1.example.com --foreman-db-host foreman-vip.example.com --foreman-db-username <SAME AS MASTER> --foreman-db-replication-password <SAME AS MASTER>`

Failover should be just as simple as touching the trigger file. e.g `touch /var/lib/pgsql/data/failover`. You can then set `--foreman-db-node-type master` for consistency.
If the master then recovers you can do `rm -rf /var/lib/pgsql/data/` and then set the machine up as a slave.

By default slaves are asynchronous, slaves can be set to synchronous using the `--foreman-db-synchronous-names` option, though don't do this during initial setup as if the slaves aren't running yet it will result in a error.

### DIFF
--- a/manifests/database/postgresql.pp
+++ b/manifests/database/postgresql.pp
@@ -10,18 +10,27 @@ class foreman::database::postgresql {
     default => postgresql_password($::foreman::db_username, $::foreman::db_password),
   }
 
+  $replication_password = postgresql_password($::foreman::db_replication_username, $::foreman::db_replication_password)
+
+  $port = $::foreman::db_port ? {
+    'UNSET' => 5432,
+    default => $::foreman::db_port,
+  }
+
   # Prevents errors if run from /root etc.
   Postgresql_psql {
     cwd => '/',
   }
 
-  include ::postgresql::client, ::postgresql::server
+  class { "foreman::database::postgresql::${::foreman::db_node_type}": }
 
-  postgresql::server::db { $dbname:
-    user     => $::foreman::db_username,
-    password => $password,
-    owner    => $::foreman::db_username,
-    encoding => 'utf8',
-    locale   => 'en_US.utf8',
+  unless $::foreman::db_host == 'UNSET' {
+    postgresql::validate_db_connection { 'validate postgres connection':
+      database_host     => $::foreman::db_host,
+      database_username => $::foreman::db_username,
+      database_password => $::foreman::db_password,
+      database_name     => $dbname,
+      database_port     => $port,
+    }
   }
 }
--- a/manifests/database/postgresql/master.pp
+++ b/manifests/database/postgresql/master.pp
@@ -1,0 +1,48 @@
+# Set up the foreman master database using postgresql
+class foreman::database::postgresql::master(
+  $wal_keep_segments = 256
+) {
+  $db_all_cluster_hostnames = unique(concat($::foreman::db_cluster_hostnames, $::fqdn))
+
+  class {'::postgresql::server':
+    listen_addresses => '*',
+  }
+
+  if $::foreman::db_synchronous_names {
+    postgresql::server::config_entry { 'synchronous_standby_names':
+      value => join($::foreman::db_synchronous_names, ','),
+    }
+  }
+
+  postgresql::server::db { $::foreman::database::postgresql::dbname:
+    user     => $::foreman::db_username,
+    password => $::foreman::database::postgresql::password,
+    owner    => $::foreman::db_username,
+    encoding => 'utf8',
+    locale   => 'en_US.utf8',
+  }
+
+  foreman::database::postgresql::db_access { $db_all_cluster_hostnames: }
+
+  postgresql::server::role { $::foreman::db_replication_username:
+    password_hash => $::foreman::database::postgresql::replication_password,
+    replication   => true,
+  }
+
+  postgresql::server::config_entry { 'wal_level':
+    value => 'hot_standby',
+  }
+
+  postgresql::server::config_entry { 'max_wal_senders':
+    value => count($db_all_cluster_hostnames),
+  }
+
+  postgresql::server::config_entry { 'wal_keep_segments':
+    value => $wal_keep_segments,
+  }
+
+  postgresql::server::config_entry { 'hot_standby':
+    value => 'on',
+  }
+
+}
--- a/manifests/database/postgresql/slave.pp
+++ b/manifests/database/postgresql/slave.pp
@@ -1,0 +1,60 @@
+# Set up the foreman slave database using postgresql
+class foreman::database::postgresql::slave(
+  $wal_keep_segments = 256
+) {
+  require ::postgresql::params
+
+  $db_all_cluster_hostnames = unique(concat($::foreman::db_cluster_hostnames, $::fqdn))
+
+  file { 'Write .pgpass':
+    content => "*:*:*:${::foreman::db_replication_username}:${::foreman::db_replication_password}\n",
+    path    => "${::foreman::params::postgres_home}/.pgpass",
+    owner   => $::postgresql::params::user,
+    group   => $::postgresql::params::group,
+    mode    => '0600',
+  }
+  -> exec { 'Initialize new postgresql cluster':
+    command => "${::postgresql::params::bindir}/pg_basebackup -w -c fast -X stream -h ${::foreman::db_host} -p ${::foreman::database::postgresql::port} -U ${::foreman::db_replication_username} -D ${::postgresql::params::datadir}",
+    creates => "${::postgresql::params::datadir}/base",
+    user    => $::postgresql::params::user,
+  }
+  -> class {'::postgresql::server':
+    listen_addresses     => '*',
+    manage_recovery_conf => true,
+    needs_initdb         => false,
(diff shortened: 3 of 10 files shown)


manifests/database/postgresql.pp
```diff
@@ -10,20 +13,111 @@
...
   # Prevents errors if run from /root etc.
   Postgresql_psql {
     cwd => '/',
   }
 
-  include ::postgresql::client, ::postgresql::server
+  if $::foreman::db_cluster {
+    $db_all_cluster_hostnames = concat($::foreman::db_cluster_hostnames, $::fqdn)
+    if $::foreman::db_node_type == 'master' {
+      class {'postgresql::server':
```
ekohl: We try very hard not to manage external classes so users can override them in profiles.

manifests/database/postgresql.pp
```diff
@@ -10,20 +13,111 @@
...
+    }
 
-  postgresql::server::db { $dbname:
-    user     => $::foreman::db_username,
-    password => $password,
-    owner    => $::foreman::db_username,
-    encoding => 'utf8',
-    locale   => 'en_US.utf8',
+  } else {
+    class {'::postgresql::client':}
```
ekohl: We try hard not to manage external classes so users can override them in profiles.

mmoll:
I don't think this should be a part of this module, but rather using the puppetlabs-postgresql module to set up the cluster and then conenct it as external DB via puppet-foreman feels the more correct way.

manifests/database/postgresql.pp
```diff
@@ -10,20 +13,111 @@
     default => postgresql_password($::foreman::db_username, $::foreman::db_password),
   }
 
+  $rep_password = $::foreman::db_replication_password ? {
+    'UNSET' => false,
```
ekohl: I strongly dislike the UNSET we already have but dropping them has been too much of an effort until now. Let's at least try to avoid adding more.

manifests/params.pp
```diff
@@ -145,6 +157,7 @@
           $plugin_prefix = 'tfm-rubygem-foreman_'
           $init_config = '/etc/sysconfig/foreman'
           $init_config_tmpl = 'foreman.sysconfig'
+          $postgres_home = "/var/lib/pgsql${::postgresql::params::package_version}"
```
ekohl: Won't this fail if the params aren't included?

manifests/database/postgresql.pp
```diff
@@ -10,20 +13,111 @@
...
+        user     => $::foreman::db_username,
+        password => $password,
+        owner    => $::foreman::db_username,
+        encoding => 'utf8',
+        locale   => 'en_US.utf8',
+      }
+    } elsif $::foreman::db_node_type == 'slave' {
+      file { 'Write .pgpass':
+        content => "*:*:*:${::foreman::db_replication_username}:${::foreman::db_replication_password}\n",
+        path    => "${::foreman::params::postgres_home}/.pgpass",
```
ekohl: It surprises me a little that there's no variable for this in puppetlabs-postgresql.

manifests/database/postgresql.pp
```diff
@@ -10,20 +13,111 @@
...
+    postgresql::server::db { $dbname:
+      user     => $::foreman::db_username,
+      password => $password,
+      owner    => $::foreman::db_username,
+      encoding => 'utf8',
+      locale   => 'en_US.utf8',
+    }
   }
 
-  Postgresql::Server::Role[$::foreman::db_username] -> Postgresql::Server::Database[$dbname]
```
ekohl: This chaining was added in puppetlabs-postgresql 4.1.0. Please also increase the minimum version in metadata.json.

manifests/init.pp
```diff
@@ -159,6 +159,24 @@
...
+#                               type:Boolean
+#
+# $db_node_type::               Cluster node type, can be one of 'master' or 'slave'.
+#                               type:Optional[String]
+#
+# $db_cluster_hostnames::       List of hostnames in the cluster. Excluding this one.
+#                               type:Optional[Array]
+#
+# $db_synchronous_names::       List of hostnames in the cluster using synchronous replication.
+#                               type:Optional[Array]
```
ekohl: I think this could even be ```Optional[Array[String]]```

manifests/init.pp
```diff
@@ -159,6 +159,24 @@
...
 #                               type:Boolean
 #
+# $db_cluster::                 Setup a database cluster. PostgreSQL only.
+#                               type:Boolean
+#
+# $db_node_type::               Cluster node type, can be one of 'master' or 'slave'.
+#                               type:Optional[String]
+#
+# $db_cluster_hostnames::       List of hostnames in the cluster. Excluding this one.
+#                               type:Optional[Array]
```
ekohl: I think this could even be ```Optional[Array[String]]```

manifests/init.pp
```diff
@@ -159,6 +159,24 @@
 # $db_manage_rake::             if enabled, will run rake jobs, which depend on the database
 #                               type:Boolean
 #
+# $db_cluster::                 Setup a database cluster. PostgreSQL only.
+#                               type:Boolean
+#
+# $db_node_type::               Cluster node type, can be one of 'master' or 'slave'.
```
ekohl: Would it make sense to have a an explicit name for the standalone version?

manifests/database/postgresql.pp
```diff
@@ -10,20 +13,111 @@
...
+
   # Prevents errors if run from /root etc.
   Postgresql_psql {
     cwd => '/',
   }
 
-  include ::postgresql::client, ::postgresql::server
+  if $::foreman::db_cluster {
+    $db_all_cluster_hostnames = concat($::foreman::db_cluster_hostnames, $::fqdn)
+    if $::foreman::db_node_type == 'master' {
```
ekohl: Does it make sense to split this into ```foreman::database::postgresql::master``` and ```foreman::database::postgresql::slave```? Maybe just ```foreman::database::postgresql::cluster```?

manifests/database/postgresql.pp
ekohl: Created https://github.com/theforeman/puppet-foreman/pull/546 to address this. It was also 4.2.0.

sean797:
@mmoll I totally agree the better way to scale is using an external DB, but I think is defiantly valuable to be able to deploy a full self-maintained foreman cluster. Not everyone wants to build/manage/maintain an external DB.

@ekohl I think I addressed all your comments, also added some tests. Thanks for the quick review 👍

manifests/database/postgresql.pp
sean797: Yes, I was also - really annoying. I'll send a PR to them this week.

manifests/database/postgresql/master.pp
```diff
@@ -0,0 +1,46 @@
+# Set up the foreman master database using postgresql
+class foreman::database::postgresql::master {
+  $db_all_cluster_hostnames = concat($::foreman::db_cluster_hostnames, $::fqdn)
```
ekohl: Maybe a ```unique()``` around it as well to be sure?

manifests/database/postgresql/master.pp
```diff
@@ -0,0 +1,46 @@
...
+
+  postgresql::server::db { $::foreman::database::postgresql::dbname:
+    user     => $::foreman::db_username,
+    password => $::foreman::database::postgresql::password,
+    owner    => $::foreman::db_username,
+    encoding => 'utf8',
+    locale   => 'en_US.utf8',
+  }
+
+  ::foreman::database::postgresql::db_access { $db_all_cluster_hostnames: }
```
ekohl: Defines don't need the ```::``` prefix.

manifests/database/postgresql/slave.pp
```diff
@@ -0,0 +1,58 @@
...
+  postgresql::server::config_entry { 'wal_level':
+    value => 'hot_standby',
+  }
+
+  postgresql::server::config_entry { 'max_wal_senders':
+    value => count($db_all_cluster_hostnames),
+  }
+
+  postgresql::server::config_entry { 'wal_keep_segments':
+    value => '256',
```
ekohl: Would it make sense to make this a class parameter (with the same default)?

manifests/database/postgresql/slave.pp
```diff
@@ -0,0 +1,58 @@
+# Set up the foreman slave database using postgresql
+class foreman::database::postgresql::slave {
+  require ::postgresql::params
+
+  $db_all_cluster_hostnames = concat($::foreman::db_cluster_hostnames, $::fqdn)
```
ekohl: Use a ```unique()``` as well?

sean797:
Humm a `puppet-foreman_profiles` module could be useful in the future..

Anyway I've updated this and tests passed now 👍

PR #548, opened 2017-04-05
Move existing Puppet 4 data types to class parameters

domcleal:
1. Moves existing Puppet 4 data types from the class/parameter docs into the class parameters themselves.
1. Converts some older Kafo-only data types into the Puppet 4 style and moves them into the class parameters.
1. Removes the validate_* family of functions, updates a few instances where they had better validation than the data types.

Requires https://github.com/theforeman/foreman-installer-modulesync/pull/58.


manifests/compute/ec2.pp
```diff
@@ -5,9 +5,8 @@
 # === Parameters:
 #
 # $version::  Package version to install, defaults to installed
-#             type:Optional[String]
 #
-class foreman::compute::ec2($version = 'installed') {
+class foreman::compute::ec2(Optional[String] $version = 'installed') {
```
ekohl: I know it was already the case that it was optional, but since we have a default is it even possible to pass in undef somehow?

domcleal: No, it wasn't really - I'll revert it back to `String`. There was a bug (http://projects.theforeman.org/issues/6158) in Kafo which was why this was made `Optional` in 3f5238c.

manifests/init.pp
```diff
@@ -3,379 +3,289 @@
...
+  Boolean $configure_scl_repo = $::foreman::params::configure_scl_repo,
+  Optional[Boolean] $selinux = $::foreman::params::selinux,
+  Boolean $gpgcheck = $::foreman::params::gpgcheck,
+  String $version = $::foreman::params::version,
+  Enum['installed', 'present', 'latest'] $plugin_version = $::foreman::params::plugin_version,
+  Boolean $db_manage = $::foreman::params::db_manage,
+  Enum['mysql', 'postgresql', 'sqlite'] $db_type = $::foreman::params::db_type,
+  Optional[Enum['mysql2', 'postgresql', 'sqlite3', 'UNSET']] $db_adapter = 'UNSET',
+  Optional[String] $db_host = 'UNSET',
+  Variant[Undef, Enum['UNSET'], Integer[0, 65535]] $db_port = 'UNSET',
```
ekohl: What's the difference between ```Undef``` in the ```Variant``` and ```Optional```?

domcleal: No practical difference, it was just a little shorter and had fewer brackets.

ekohl: Makes sense, learned something new :)

mmoll:
could you rebase and bump the major version?

domcleal:
PR rebased, module version increased.

mmoll:
merged, thanks @domcleal!

PR #549, opened 2017-04-06
Limit number of parallel_test processes

domcleal:
Prevent test jobs being killed for excessive resource consumption on
Travis CI.

---

Example failure: https://api.travis-ci.org/jobs/219217285/log.txt?deansi=true (see "Killed" part way down, running 32 processes).

Corresponding modulesync change made in https://github.com/theforeman/foreman-installer-modulesync/pull/59.

### DIFF
--- a/.sync.yml
+++ b/.sync.yml
@@ -3,6 +3,9 @@
   beaker_sets:
     - docker/centos-7
     - docker/debian-8
+  env:
+    global:
+      - PARALLEL_TEST_PROCESSORS=16
 Gemfile:
   extra:
     - gem: webmock
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,10 +4,13 @@
 rvm:
   - 2.1.5
 env:
-  # First test the major distros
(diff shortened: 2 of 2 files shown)


ekohl:
Is this something we should limit globally in the modulesync?

domcleal:
I'm not sure, because it isn't a limit (a maximum). It's overriding the number of processes used, so a fixed value might not be appropriate for smaller modules.

ekohl:
By default it always uses the number of cores available so for smaller modules it may even be beneficial to limit it, but let's start with this and see how it goes.

lazyfrosch:
I've encountered something similar. I only had to limit on on a certain ruby version. Maybe that helps.

https://github.com/Icinga/puppet-icinga2/commit/9768f71956c6080494a413659d1f96a05ee7c5c3

And yes, I'd recommend to avoid setting it globally. Especially for local tests :wink:

mmoll:
merged, thanks @domcleal!

PR #553, opened 2017-04-11
Replace tftp::params::root use with Puppet lookup

domcleal:


mmoll:
merged, thanks @domcleal!

PR #556, opened 2017-04-24
Add features property to foreman_smartproxy to check enabled features

domcleal:
Implemented on the rest_v3 provider to validate that the enabled
features found by Foreman include all of the features listed in the
'features' property on the resource.

This should detect configuration errors on the smart proxy, or missing
Foreman plugins for the registered smart proxy.

### DIFF
--- a/lib/puppet/type/foreman_smartproxy.rb
+++ b/lib/puppet/type/foreman_smartproxy.rb
@@ -1,6 +1,8 @@
 Puppet::Type.newtype(:foreman_smartproxy) do
   desc 'foreman_smartproxy registers a smartproxy in foreman.'
 
+  feature :feature_validation, "Enabled features can be validated", methods: [:features, :features=]
+
   ensurable
 
   newparam(:name, :namevar => true) do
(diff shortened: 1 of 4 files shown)


lib/puppet/type/foreman_smartproxy.rb
```diff
@@ -23,6 +25,17 @@
     desc 'Foreman oauth consumer_secret'
   end
 
+  newproperty(:features, required_features: :feature_validation, array_matching: :all) do
+    desc 'Features enabled on smart proxy. Setting this validates that they all
```
ekohl: I'm not exactly sure what this means. Should the word they be removed?

domcleal:
> If I understand it correct this list of features should be manually maintained and isn't automatically detected by calling the proxy /features API, right?

Yes, it's manually maintained on the resource. So if you'd set up a smart proxy instance in a module with the DHCP and TFTP features, and were then registering it via `foreman_smartproxy`, you'd set `features => ['DHCP', 'TFTP']` and the resource would ensure they were active.

It's detected by the Foreman smart proxy API responses, so when creating or refreshing the smart proxy resource in Foreman only - it doesn't call the smart proxy directly. The data is already present in the Foreman responses. Both Foreman and the smart proxy's /features responses only list features that were successfully enabled and recognised, so this detects configuration failures.

> The tests also fail, but not sure they're related.

This is due to https://github.com/puppetlabs/puppetlabs-stdlib/pull/765, should be fixed soon.

lib/puppet/type/foreman_smartproxy.rb
domcleal: Yes, it should, thanks. I've re-worded it slightly too.

domcleal:
> It'd be nice if we could somehow collect the features that should be enabled within the foreman_proxy module, but I'm not sure that's even possible within puppet.

Yes, this is possible - I intend to make that change once this PR's complete/merged. It's trivial to do for built-in features and for plugins too with datacat.

ehelms:
Will this fix the following error:

```
# [ERROR 2017-05-03 23:25:24 verbose]  Proxy centos7-bats-ci.example.com cannot be registered: Features "logs, pulp, puppet, puppetca, and tftp" in this proxy are not recognized by Foreman. If these features come from a Smart Proxy plugin, make sure Foreman has the plugin installed too.
```

mmoll:
merged, thanks @domcleal!

PR #558, opened 2017-05-08
mark compatible with recent puppetlabs-concat

mmoll:


mmoll:
Will rekick tests, once puppetlabs-stdlib is fixed.

PR #560, opened 2017-05-10
Validate feature list on foreman_smartproxy refresh

domcleal:
When refreshing the feature list on foreman_smartproxy via a notify,
validate it against the expected features to ensure they're all
configured correctly still.

The cached proxy data is replaced with the refresh response where
possible (since #19476) and is otherwise cleared to validate against
the new feature list (including via the `#features=` setter).


mmoll:
merged, thanks @domcleal!

PR #561, opened 2017-05-10
Improve formatting of foreman_smartproxy feature lists

domcleal:
Originally:

    change from ["Logs", "Puppet CA", "TFTP"] to Puppet Puppet CA TFTP Logs failed

Now:

    change from [Logs, Puppet CA, TFTP] to [Logs, Puppet, Puppet CA, TFTP] failed

### DIFF
--- a/lib/puppet/type/foreman_smartproxy.rb
+++ b/lib/puppet/type/foreman_smartproxy.rb
@@ -35,6 +35,11 @@ Puppet::Type.newtype(:foreman_smartproxy) do
     def insync?(current)
       (@should.sort - current.sort).empty?
     end
+
+    def is_to_s(value)
+      "[#{value.sort.map(&:inspect).join(', ')}]"
+    end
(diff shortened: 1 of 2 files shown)


ekohl:
I doubt it'll happen, but is ```,``` valid in a feature name? I wonder if we should somehow restrict feature names to ```[A-Za-z0-9 _-]``` or if that'll be an annoying restriction somewhere.

domcleal:
Yes, I think a comma would be valid in Foreman and in the Puppet type. If Foreman allows it then the type shouldn't prevent them, however to use one would certainly be unusual - these are hard-coded strings by developers rather than user editable.

I think I'm OK with the possible future ambiguity, if only to make the output more human-readable. But I can add the quotation marks back in if you'd like.

domcleal:
PR updated with quote marks (using #inspect) around each item in the list.

mmoll:
merged, thanks @domcleal!

PR #564, opened 2017-05-24
Fixes #19550 - support smart card certificates

ares:


ekohl:
The code looks correct but I'm unable to test it since I don't have a smart card.

mmoll:
same here

ares:
I asked the reporter whether they could test it. I at least verified that the installer adds this correctly to vhost and apache2 starts. When I get some update from reporter, I'll update this PR. In worst case scenario, if it does not work, it should not do any harm...

ares:
User confirms it works for them (see https://bugzilla.redhat.com/show_bug.cgi?id=1449088)

mmoll:
merged, díky @ares!

PR #567, opened 2017-06-07
Updated oauth syntax, support for disabling SSL Validation

flyinbutrs:
Added an optional parameter to toggle SSL verification, and updated to reflect Apipie's preferred oauth config.


mmoll:
merged, thanks @flyinbutrs!

PR #568, opened 2017-06-13
Made it possible to disable passenger repo in apache

enemarke:
Added option to disable passenger repo i apache from foreman.


enemarke:
After committing I came to think that this probably does not belong in puppet-foreman, but should be configured directly in the apache module?

mmoll:
normally things like that would be set in hiera, because of the puppetlabs-apache module design.

PR #571, opened 2017-06-21
Fixes #20079 - SSL secured and verified Pg connection

mbacovsky:
To setup DB with SSL and verification use params:

  db_sslmode => 'verify-full',
  db_root_cert => 'ca_bundle.pem'

Default DB sslmode is 'prefer' - non-verified SSL with fallback to
non-SSL. To use SSL secured connection with CA verification sslmode
needs to be set to either 'verify-ca' or 'verify-full'. Underlying libpg
uses DB root cert stored at '~/.postgresql/root.crt' which is in case of
Foreman at '/usr/share/foreman/.postgresql/root.crt'. There is no way to
setup different path (besides using env vars). System CA trust is not
supported. The cert needs to be real file as links are not allowed too.

For more details see SSL support in libpg:
  https://www.postgresql.org/docs/9.2/static/libpq-ssl.html

### DIFF
--- a/manifests/database/postgresql.pp
+++ b/manifests/database/postgresql.pp
@@ -10,6 +10,22 @@ class foreman::database::postgresql {
     default => postgresql_password($::foreman::db_username, $::foreman::db_password),
   }
 
+  if $::foreman::db_root_cert {
+    $pg_cert_dir = "${::foreman::app_root}/.postgresql/"
+
+    file { $pg_cert_dir:
+      ensure => 'directory',
+    }
+
+    file { "${pg_cert_dir}/root.crt":
+      ensure => file,
+      source => $::foreman::db_root_cert,
+      owner  => 'root',
+      group  => $::foreman::group,
+      mode   => '0640',
+    }
+  }
+
   # Prevents errors if run from /root etc.
   Postgresql_psql {
     cwd => '/',
--- a/manifests/database/postgresql.pp
+++ b/manifests/database/postgresql.pp
@@ -15,6 +15,9 @@ class foreman::database::postgresql {
 
     file { $pg_cert_dir:
       ensure => 'directory',
+      owner  => 'root',
+      group  => $::foreman::group,
+      mode   => '0640',
     }
 
     file { "${pg_cert_dir}/root.crt":
--- a/manifests/database/postgresql.pp
+++ b/manifests/database/postgresql.pp
@@ -11,7 +11,7 @@ class foreman::database::postgresql {
   }
 
   if $::foreman::db_root_cert {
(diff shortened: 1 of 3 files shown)


mbacovsky:
Redmine ticket updated accordingly.

Question for reviewer: Is the `prefer` secure default sslmode? It implicitly uses non-verified SSL connection without warning to the user. Wouldn't be 'disabled` better default forcing the user to choose the sslmode consciously?

manifests/database/postgresql.pp
```diff
@@ -10,6 +10,22 @@
     default => postgresql_password($::foreman::db_username, $::foreman::db_password),
   }
 
+  if $::foreman::db_root_cert {
+    $pg_cert_dir = "${::foreman::app_root}/.postgresql/"
+
+    file { $pg_cert_dir:
+      ensure => 'directory',
```
ekohl: You should manage owner and permission here as well. I've learned that it's better to be explicit about that than to rely on defaults. There could be a weird umask for example.

manifests/init.pp
```diff
@@ -225,6 +227,7 @@
   Optional[String] $db_username = $::foreman::params::db_username,
   Optional[String] $db_password = $::foreman::params::db_password,
   Optional[String] $db_sslmode = 'UNSET',
+  Optional[String] $db_root_cert = 'UNSET',
```
ekohl: I dislike the ```UNSET```. It used to be needed in puppet versions but now I think we can just use ```undef```.

manifests/database/postgresql.pp
```diff
@@ -10,6 +10,25 @@
     default => postgresql_password($::foreman::db_username, $::foreman::db_password),
   }
 
+  if $::foreman::db_root_cert {
+    $pg_cert_dir = "${::foreman::app_root}/.postgresql/"
```
ekohl: Nit: this trailing slash isn't needed.

mbacovsky:
@ekohl, updated. I gave the sslmode conditioned by presence of the root cert a try. Does it make sense and is it worth it?

mmoll:
This is looking good, @mbacovsky I presume you tested this already with verification using the Katello certificate factory?

mbacovsky:
@mmoll I've created standalone CA cert chain to sign the pg server certificate. My guess was this will better simulate the real deployment where the Foreman and remote Pg server will have certs signed by common corporate CA or the Pg server CA will be totally independent. 

Anyway the patch just installs the pg server CA cert so it shouldn't depend on where the cert comes from.

mmoll:
fair enough, merged, díky @mbacovsky!

PR #577, opened 2017-07-24
Refs #20378 - update node.rb for Puppet5 env fact

mmoll:


GregSutcliffe:
Ty @mmoll!

alexjfisher:
FWIW (and who doesn't like commenting on a 3 year old PR!?), I think the original implementation suffers from a race condition when using the `--watch-facts` option.

The code watches vardir/yaml/facts/some_node.yaml but fills in the environment from vardir/yaml/node/some_node.yaml.
On my puppetserver installation, the node file is written ~200ms after the facts file and I suspect this is why the environment reported in Foreman (which isn't being used as the ENC in my case), lags behind the actual value for several puppet runs (load balanced puppetservers make the experience worse).

So I'm hoping this fixes my issue, (as I think might upgrading my puppetserver!).  Thanks @mmoll ! :)

PR #578, opened 2017-08-02
Include a resource type group to ensure foreman user groups.

viniciusxavierbove:


manifests/config.pp
```diff
@@ -51,6 +51,10 @@
   }
 
   if $::foreman::manage_user {
+    group { $::foreman::user_groups:
+      ensure => 'present',
+    }
+
```
ekohl: Here's trailing whitespace which doesn't agree with our linting.

sean797:
@vinixavier whats the status here? :-)

PR #579, opened 2017-08-10
LDAP Authentication Type/Provider

matonb:
RFC

Clunky with a bunch of repeated code in updates, but it appears to be working....

### DIFF
--- a/lib/puppet/provider/foreman_auth/rest_v3.rb
+++ b/lib/puppet/provider/foreman_auth/rest_v3.rb
@@ -1,0 +1,282 @@
+Puppet::Type.type(:foreman_auth).provide(
+  :rest_v3,
+  parent: Puppet::Type.type(:foreman_resource).provider(:rest_v3)
+) do
+  confine :feature => [:json, :oauth]
+
+  mk_resource_methods
+
+  def b2i(value)
+    value ? 1 : 0
+  end
+
+  def source
+    @source ||= begin
+      r = request(:get,
+                  'api/v2/auth_source_ldaps',
+                  search: %(name="#{resource[:name]}"))
+      unless success?(r)
+        raise Puppet::Error,
+              "LDAP #{resource[:name]} cannot be retrieved: #{error_message(r)}"
+      end
+      JSON.parse(r.body)['results'][0]
+    end
+  end
+
+  def id
+    source ? source['id'] : nil
+  end
+
+  def create
+    server_type = servertype_to_s(resource[:server_type])
+    case server_type
+    when 'active_directory'
+      attr_login = 'userPrincipalName'
+    else
+      attr_login = 'uid'
+    end
+    port = resource[:tls] ? '636' : '389'
+    post_data = {
+      auth_source_ldap: {
+        account:           resource[:account],
+        account_password:  resource[:account_password],
+        attr_firstname:    resource[:attr_firstname] || 'givenname',
+        attr_lastname:     resource[:attr_lastname] || 'sn',
+        attr_login:        resource[:attr_login] || attr_login,
+        attr_mail:         resource[:attr_mail] || 'mail',
+        attr_photo:        resource[:attr_photo],
+        base_dn:           resource[:base_dn],
+        host:              resource[:host],
+        name:              resource[:name],
+        onthefly_register: b2i(resource[:onthefly_register]) || false,
+        port:              resource[:port] || port,
+        server_type:       server_type,
+        tls:               b2i(resource[:tls]) || true
+       }
+    }.to_json
+    r = request(:post,
+                'api/v2/auth_source_ldaps',
+                {},
+                post_data)
+    unless success?(r)
+      raise Puppet::Error,
+            "LDAP #{resource[:name]} cannot be registered: #{error_message(r)}"
+    end
+  end
+
+  def destroy
+    r = request(:delete,
+                "api/v2/auth_source_ldaps/#{id}")
+    unless success?(r)
+      raise Puppet::Error,
+            "LDAP #{resource[:name]} cannot be removed: #{error_message(r)}"
+    end
+    @source = nil
+  end
+
+  def exists?
+    !id.nil?
+  end
+
+  def account=(value)
+    put_data = {
+      auth_source_ldap: {
+        account: value
+      }
+    }.to_json
+    r = request(:put,
+                "api/v2/auth_source_ldaps/#{id}",
+                {},
+                put_data)
+    unless success?(r)
+      raise Puppet::Error,
+            "LDAP #{resource[:account]} cannot be updated: #{error_message(r)}"
+    end
+  end
+
+  def account_password=(value)
+    put_data = {
+      auth_source_ldap: {
+        account_password: value
+      }
+    }.to_json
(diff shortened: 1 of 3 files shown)


lib/puppet/type/foreman_auth.rb
```diff
@@ -0,0 +1,121 @@
...
+  newproperty(:groups_base) do
+    desc 'LDAP server hostname'
+    defaultto ''
+  end
+
+  newparam(:name, namevar: :true) do
+    desc 'Name of the LDAP authentication source.'
+  end
+
+#  newproperty(:onthefly_register, :boolean => true, :parent => Puppet::Parameter::Boolean) do
```
alexjfisher: Instead of `Puppet::Parameter::Boolean`, can you use `Puppet::Property::Boolean` here?
eg https://github.com/Katello/puppet-pulp/blob/479f65c1e7531e2ab4f174725d2c58d842cc1339/lib/puppet/type/pulp_rpmrepo.rb#L68

lib/puppet/type/foreman_auth.rb
```diff
@@ -0,0 +1,121 @@
...
+  newproperty(:onthefly_register) do
+    desc 'Automatically create accounts in Foreman.'
+    defaultto false 
+    newvalues(true, false, 'true', 'false')
+  end
+
+  newproperty(:port) do
```
ekohl: You can use ```newvalues(/\d+/)``` to only accept (positive) numbers.

alexjfisher:
Is it desirable for the properties to have defaults?

The 'Puppet Types and Providers' book has
> The defaultto method is used to specify default values for any attribute. Defaults are typically only specified for parameters. They are less common for properties because a default value will result in that property always being managed (there is no way to unset a value for the property).

Ideally, the following would just enforce tls on.
```puppet
foreman_auth { 'AD':
  tls => true,
}
```
With the defaults present, I think it would wipe the existing configuration. You'd probably need to have defaults in the provider create method though. I'll try to hunt out an example.

lib/puppet/provider/foreman_auth/rest_v3.rb
```diff
@@ -0,0 +1,274 @@
...
+  end
+
+  def id
+    source ? source['id'] : nil
+  end
+
+  def create
+    post_data = {
+      auth_source_ldap: {
+        account:           resource[:account],
```
alexjfisher: You could put any defaults here. See https://github.com/voxpupuli/puppet-proxysql/blob/00eb03323c52fe0f2665fdaf3e60937eb7e22867/lib/puppet/provider/proxy_mysql_server/proxysql.rb#L59 for example

lib/puppet/type/foreman_auth.rb
matonb: Probably, I'll give it a go and see if it works, I expect the 'parameter' bit is left over from a previous attempt.

lib/puppet/provider/foreman_auth/rest_v3.rb
matonb: Yup that'll do the trick

matonb:
On subsequent puppet runs I'm seeing the following notices being generated, not necessarily a bad thing but I would like at the very least to obfuscate the password entry, maybe base_dn as well:
```
Notice: /Stage[main]/Profiles::Foreman::Ui/Foreman_auth[test.co.uk]/account: defined 'account' as 'svc_domainjoin'
Notice: /Stage[main]/Profiles::Foreman::Ui/Foreman_auth[test.co.uk]/account_password: defined 'account_password' as 'displayed in clear text'
Notice: /Stage[main]/Profiles::Foreman::Ui/Foreman_auth[test.co.uk]/base_dn: defined 'base_dn' as 'a valid dn'
Notice: /Stage[main]/Profiles::Foreman::Ui/Foreman_auth[test.co.uk]/host: defined 'host' as 'test.co.uk'
Notice: /Stage[main]/Profiles::Foreman::Ui/Foreman_auth[test.co.uk]/onthefly_register: defined 'onthefly_register' as 'true'
Notice: /Stage[main]/Profiles::Foreman::Ui/Foreman_auth[test.co.uk]/port: defined 'port' as '3268'
Notice: /Stage[main]/Profiles::Foreman::Ui/Foreman_auth[test.co.uk]/server_type: defined 'server_type' as 'Active Directory'
```

Any ideas ?

lib/puppet/type/foreman_auth.rb
matonb: Thanks for the comment, I've replaced the vanilla property with `Puppet::Property::Boolean`

alexjfisher:
@matonb IIRC, you can change the output by overriding `change_to_s`.  See `Customizing Event Output` in chapter 4 of `Puppet Types and Providers`

alexjfisher:
Normally looking for examples in the core types isn't helpful, (they were written a long time ago), but in the case of `User`, I think the way it redacts passwords is just what you need.
https://github.com/puppetlabs/puppet/blob/6c3a895fe5d796b861e3a8a44984ca343ca0c9c4/lib/puppet/type/user.rb#L227-L240

matonb:
@alexjfisher Thanks for the link, overriding those functions did the trick.

lib/puppet/type/foreman_auth.rb
```diff
@@ -0,0 +1,116 @@
+require 'puppet/parameter/boolean'
```
alexjfisher: Should be `require 'puppet/property/boolean'`?

matonb: Oh, FFS :)

ekohl:
What is the current status of this? Is it still a RFC/WIP?

matonb:
@ekhol Removed WIP from the title.

It's clunky but works...

matonb:
@mdwheele 

I've been rather busy with other 'stuff' of late and this branch has been left waiting for any comments.
As I said in my previous post, it's a bit clunky (not refined or elegant code) but it does work

mdwheele:
Nono, it's all good. Was just curious if there was an estimate from maintainer on pulling this in. I've gone ahead and based my work on these commits so I could move forward. Works well enough for what I need. Thanks!

mdwheele:
@matonb 

I had to add `:usergroup_sync` to the type to control. Also, the reason we're getting Notices every run is that we're using `mk_resource_methods` to add basic setters/getters, but that only works if `@property_hash` is populated by instances / prefetch. I think the simplest solution would be to implement getters that use a memoized resource to get attributes. Trying to have these resources implement `instances` is problematic because it can't receive parameters via resource such as the base URL and CA file. For us, we could derive most of this stuff from configs on disk / database, but the CA file can't be easily retrieved (and the default should remain the Puppet generated CA for compatibility). We use a certificate issued by InCommon, so we have to use the machine cert bundle instead.

matonb:
@mdwheele Please fill your boots, my effort is a hackathon with the blunt side of an axe based on some bits I found lurking elsewhere.
I'm glad that someone else is finding this to be a useful proposed addition.  I'm happy to help where I can but this area is by no means my forte!

mmoll:
closing due to inactivity

Issue #580, opened 2017-08-25
curl of fdi image fails due to missing boot directory


tuxmea:
When activating discorvery extension during initial installation, the curl of the fdi image fails:

`
 /Stage[main]/Foreman::Plugin::Discovery/Foreman::Remote_file[/var/lib/tftpboot/boot/fdi-image-latest.tar]/Exec[retrieve_/var/lib/tftpboot/boot/fdi-image-latest.tar]/returns: change from notrun to 0 failed: '/usr/bin/curl -s http://downloads.theforeman.org/discovery/releases/latest/fdi-image-latest.tar -o /var/lib/tftpboot/boot/fdi-image-latest.tar' returned 23 instead of one of [0]
`
curl can not save the file due to missing boot directory.

PR #581, opened 2017-08-25
Create boot directory if not already declared.

tuxmea:
Fixes GH-580

### DIFF
--- a/manifests/plugin/discovery.pp
+++ b/manifests/plugin/discovery.pp
@@ -25,9 +25,12 @@ class foreman::plugin::discovery (
   if $install_images {
     $tftp_root_clean = regsubst($tftp_root, '/$', '')
 
+    ensure_resource('file', "${tftp_root_clean}/boot", {ensure => directory})
+
     foreman::remote_file {"${tftp_root_clean}/boot/${image_name}":
       remote_location => "${source_url}${image_name}",
       mode            => '0644',
+      require         => File["${tftp_root_clean}/boot"],
     } ~> exec { "untar ${image_name}":
       command => "tar xf ${image_name}",
       path    => '/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin',


manifests/plugin/discovery.pp
```diff
@@ -25,6 +25,12 @@
   if $install_images {
     $tftp_root_clean = regsubst($tftp_root, '/$', '')
 
+    if ! defined(File["${tftp_root_clean}/boot"]) {
```
ekohl: I prefer ```ensure_resource('file', "${tftp_root_clean}/boot", {ensure => directory})```

tuxmea: Done. I only wonder why this is now three commits...

mmoll: I think this is still not correct. ;)
Please squash all your commits together and force-push, if possible - however, we can also fix this up on merge...

manifests/plugin/discovery.pp
```diff
@@ -25,6 +25,8 @@
   if $install_images {
     $tftp_root_clean = regsubst($tftp_root, '/$', '')
 
+    ensure_resource('file', "${tftp_root_clean}/boot", {ensure => directory})
+
     foreman::remote_file {"${tftp_root_clean}/boot/${image_name}":
       remote_location => "${source_url}${image_name}",
       mode            => '0644',
```
mmoll: I think this should require that file resource then.

tuxmea: Which is the oldest Puppet version supported by Foreman? Puppet 3? In Puppet 4 we have ordering = manifest as default option set.

mmoll: Hrm, that would be 4.6.2.

ekohl: Personally I prefer not to rely on manifest ordering

tuxmea: OK. Done.

PR #582, opened 2017-08-31
Fixes #20819 - Allow turning task cleanup cron on and off

adamruzicka:
to allow easy toggle of the cleanup, so far _opt-in_. For discussion see https://github.com/theforeman/foreman-tasks/pull/247


manifests/plugin/tasks.pp
```diff
@@ -20,4 +23,22 @@
     enable => true,
     name   => $service,
   }
+  $cron_content = @(END)
```
ekohl: I think it's cleaner to move this to ```files/tasks.cron``` and use ```content => file('foreman/tasks.cron')```.

manifests/plugin/tasks.pp
```diff
@@ -20,4 +23,22 @@
...
+             SHELL=/bin/sh
+             RAILS_ENV=production
+             FOREMAN_HOME=/usr/share/foreman
+
+             # Clean up expired tasks from the database
+             45 19 * * *     foreman    /usr/sbin/foreman-rake foreman_tasks:cleanup >>/var/log/foreman/cron.log 2>&1
+             | END
+  file { 'foreman-tasks-cleanup-cron':
+    ensure  => if $automatic_cleanup {
```
ekohl: The inline conditional is not that readable so I'd prefer a separate variable.

manifests/plugin/tasks.pp
```diff
@@ -8,9 +8,12 @@
 #
 # $service:: Service name
 #
+# $automatic_cleanup:: Enable automatic task cleanup using a cron job
+#
 class foreman::plugin::tasks(
-  String $package = $::foreman::plugin::tasks::params::package,
-  String $service = $::foreman::plugin::tasks::params::service,
+  String  $package = $::foreman::plugin::tasks::params::package,
```
ekohl: We don't align parameters anymore. Too much changes, huge diffs for little benefit.

adamruzicka:
I'm not against revisiting the other foreman-tasks PR, but we would probably still keep this on/off option for the cleaner, because of changes in the cleaner itself.

manifests/plugin/tasks.pp
```diff
@@ -20,4 +23,15 @@
     enable => true,
     name   => $service,
   }
+  $cron_state = if $automatic_cleanup {
```
ekohl: The following syntax is a bit more common:
```puppet
$cron_state = $automatic_cleanup ? {
  true    => 'present',
  default => 'absent',
}
```

manifests/plugin/tasks.pp
```diff
@@ -20,4 +23,15 @@
...
+    present
+  } else {
+    absent
+  }
+  file { 'foreman-tasks-cleanup-cron':
+    ensure  => $cron_state,
+    owner   => 'root',
+    path    => '/etc/cron.d/foreman-tasks-cleanup',
```
ekohl: Would ```foreman-tasks``` be more future proof? There could be another cronjob. We also often put the path in the file resource: ```file { '/etc/cron.d/foreman-tasks': }```

manifests/plugin/tasks.pp
```diff
@@ -20,4 +23,15 @@
...
     name   => $service,
   }
+  $cron_state = if $automatic_cleanup {
+    present
+  } else {
+    absent
+  }
+  file { 'foreman-tasks-cleanup-cron':
+    ensure  => $cron_state,
+    owner   => 'root',
```
ekohl: Could you add the ```group => 'root'``` as well? Otherwise Puppet will try to derive it based on the file owner of the manifest which could be different. An explicit mode might be a good idea as well.

manifests/plugin/tasks.pp
```diff
@@ -20,4 +23,16 @@
...
   }
+  $cron_state = $automatic_cleanup ? {
+    true    => 'present',
+    default => 'absent',
+  }
+  file { '/etc/cron.d/foreman-tasks':
+    ensure  => $cron_state,
+    owner   => 'root',
+    group   => 'root',
+    mode    => 'u=rw,go=r',
```
ekohl: Usually we write 0644 but this works too I believe.

manifests/plugin/tasks.pp
```diff
@@ -20,4 +23,16 @@
...
+    path    => '/etc/cron.d/foreman-tasks',
```
ekohl: This is now redundant since you have it in the title.

files/tasks.cron
```diff
@@ -0,0 +1,6 @@
+SHELL=/bin/sh
+RAILS_ENV=production
+FOREMAN_HOME=/usr/share/foreman
+
+# Clean up expired tasks from the database
+45 19 * * *     foreman    /usr/sbin/foreman-rake foreman_tasks:cleanup >>/var/log/foreman/cron.log 2>&1
```
ekohl: Usually we have line endings at the end of the file, this results in smaller diffs if you add another line.

ekohl: (done)

adamruzicka:
Still pending the tasks PR

ekohl:
Ok, added the label Waiting on contributor for now.

ehelms:
I anticipate we are going to get requests to be able to customize the cron timestamp. While users could hand edit it, the moment they run the installer again this would get wiped out. I'd recommend solving this now rather than later by introducing a variable and templatizing the cron.

adamruzicka:
Converted the cron file to a template, hope having the entire cronline as a single string is enough.

templates/tasks.cron.erb
```diff
@@ -0,0 +1,7 @@
+SHELL=/bin/sh
+RAILS_ENV=production
+FOREMAN_HOME=/usr/share/foreman
+
+# Clean up expired tasks from the database
+<% cronline = scope.lookupvar("::foreman::plugin::tasks::cron_line") %>
```
ekohl: cronline is in the local scope so ```@cronline``` also works.

adamruzicka: Chmm, the tests don't seem to like this change

manifests/plugin/tasks/params.pp
```diff
@@ -1,5 +1,7 @@
 # Data for the foreman-tasks plugin
 class foreman::plugin::tasks::params {
+  $automatic_cleanup = false
+  $cron_line = "45 19 * * *"
```
mmoll: manifests/plugin/tasks/params.pp:4:WARNING: double quoted string containing no variables

iNecas: Updated

templates/tasks.cron.erb
```diff
@@ -0,0 +1,6 @@
+SHELL=/bin/sh
+RAILS_ENV=production
+FOREMAN_HOME=/usr/share/foreman
+
+# Clean up expired tasks from the database
+<%= @cron_line %>    foreman    /usr/sbin/foreman-rake foreman_tasks:cleanup >>/var/log/foreman/cron.log 2>&1
```
ekohl: The foreman username is also available as the variable ```foreman::user```.

adamruzicka: fixed

ehelms:
Templating and configuration aspect look good, thanks @adamruzicka for making that change

mmoll:
merged, díky @adamruzicka!

adamruzicka:
Thanks everybody who helped me pull this one through

PR #583, opened 2017-08-31
Remove discovery image downloading

ekohl:
The image downloading is tied to tftp and handled by the proxy. Since puppet-foreman_proxy is already able to download images there's no sense in duplicating the code here.

Based on https://github.com/theforeman/puppet-foreman/pull/486

I think this also needs installer migrations to remove the answers.

Closes GH-580
Closes GH-581

### DIFF
--- a/.fixtures.yml
+++ b/.fixtures.yml
@@ -8,7 +8,6 @@ fixtures:
     postgresql:       'git://github.com/puppetlabs/puppetlabs-postgresql'
     puppet:           'git://github.com/theforeman/puppet-puppet'
     stdlib:           'git://github.com/puppetlabs/puppetlabs-stdlib'
-    tftp:             'git://github.com/theforeman/puppet-tftp'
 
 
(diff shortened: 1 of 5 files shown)


ekohl:
cc @tuxmea

mmoll:
not 100% sure, but I think when removing parameters we just left these in the answers file

ekohl:
Leaving them in doesn't hurt (I think) and could be useful when downgrading.

ekohl:
Anything holding this merge back?

mmoll:
no. bedankt @ekohl! :)

ares:
@ekohl doesn't it fail if you instantiate puppet class with extra parameters? I'm afraid that's what happens if you have some extra things in answer file

ekohl:
@ares I think that might have been the case in the past but with hiera it should ignore additional values. It is a good thing to test though.

PR #587, opened 2017-09-15
hardwire Rails to 4.2

mmoll:


mmoll:
This is quite a kludge, but should get the BATS tests suceeding on Ubuntu/xenial again. While at package build time, Rails is pinned to 4.2, the installer run is overwriting the settings file and then foreman-rake fails.

mmoll:
This will be removed, once plugins got compatible with Rails 5.0 and the pinning to 4.2 is removed from DEB package builds.

ekohl:
IIRC we only have a settings.plugins.d and not a settings.d where we could drop it in via packaging, correct? Or can we abuse the settings.plugins.d directory for this?

mmoll:
In `config/boot_settings.rb` the settings.yml file is read directly, so I think settings.plugins.d won't work.

ekohl:
Looking at https://github.com/theforeman/foreman/blob/develop/config/boot_settings.rb I'd propose to ship the rails settings in ```settings.yaml.dist```. I submitted https://github.com/theforeman/foreman/pull/4849 which should allow us to fix this in packaging instead.

PR #591, opened 2017-09-19
Fixes #21023 - Update httpd timeout to 600

chris1984:


manifests/config.pp
```diff
@@ -25,6 +25,13 @@
     content => template('foreman/database.yml.erb'),
   }
 
+  file { '/etc/systemd/system/httpd.service/10-longer_timeout.conf':
```
chris1984: I need to add an ensure => file here

templates/longer_startup.erb
```diff
@@ -0,0 +1,2 @@
+[Service]
+TimeoutStartSec=$::foreman::params::apache_start_timeout
```
ekohl: These are ERB templates so ERB syntax should be used. That means ```<%= scope['foreman::apache_start_timeout'] %>```.

manifests/init.pp
```diff
@@ -289,6 +291,7 @@
   Enum['none', 'plain', 'login', 'cram-md5'] $email_smtp_authentication = $::foreman::params::email_smtp_authentication,
   Optional[String] $email_smtp_user_name = $::foreman::params::email_smtp_user_name,
   Optional[String] $email_smtp_password = $::foreman::params::email_smtp_password,
+  Integer[600] $apache_start_timeout = $::foreman::params::apache_start_timeout
```
ekohl: Is there a reason this should be at least 600? You're also missing a trailing comma.

manifests/config.pp
ekohl: For systemd to pick this up you need to do a ```systemctl daemon-reload``` but in other another module we've started to use ```systemd::dropin```.

ekohl: ```ensure => file``` is indeed cleaner but is implied when you supply ```content```.

ehelms:
@ekohl I disagree with this should live in katello for the sole fact that puppet-foreman is what configures the passenger start timeout [1] and thus should be responsible for ensuring that whatever service manager is managing it matches that value for the best user experience. Ideally, puppetlabs-apache would be doing this properly, but the next best choice is the module configuring the passenger service. A user cannot predicate that their system might be slow, causing Apache to startup faster than passenger, thus giving the impression that the application or Puppet is up and running only to find out thats not quite true given Passenger is still starting.


[1] https://github.com/theforeman/puppet-foreman/pull/591/files#diff-2f383a197b1a428561b364755701d60eR35

ekohl:
@ehelms I think we both agree the best fix is in puppetlabs-apache but you do have a point that here the service is configured. In that case the workaround should live in ```foreman::config::passenger``` so that users who deploy Foreman without passenger won't attempt to deploy the config.

manifests/config.pp
ekohl: I now also see that it should be ```httpd.service.d```, at least on RH systems. For better compatibility I'd use ```${::apache::service_name}.service.d``` since on Debian it's ```apache2.service``` and RH users can set up their system to use the SCL ```httpd24-httpd``` service name. The ```systemd::dropin``` will also take care of ensuring the parent directory exists.

ehelms:
:+1: to limiting the scope as much as possible to passenger configurations only. And if possible, do the right thing and work on a proper fix into puppetlabs-apache.

ehelms:
@chris1984 Can you please revisit this?

templates/longer_startup.conf.erb
```diff
@@ -0,0 +1,2 @@
+[Service]
+TimeoutStartSec <%= scope['foreman::config::passenger::apache_start_timeout'] %>
```
ekohl: Since ```foreman::config::passenger``` has all parameters in the local scope you can use ```<%= @apache_start_timeout %>```

manifests/config/passenger.pp
```diff
@@ -195,7 +205,8 @@
         port                    => $server_ssl_port,
         priority                => $priority,
         servername              => $servername,
-        serveraliases           => $serveraliases,
+        serveraliases
```
ekohl: This looks unintended.

manifests/config/passenger.pp
```diff
@@ -181,6 +184,13 @@
         recurse => true,
       }
 
+      if $::service_provider == 'systemd' {
+        include ::systemd
+        systemd::dropin_file { '10-larger_timeout.conf':
+          unit    => '${::apache::service_name}.service.d',
+          content => template('foreman/longer_startup.conf.erb'),
```
ekohl: I'd just name this ```startup.conf.erb```.

manifests/config/passenger.pp
```diff
@@ -59,6 +59,8 @@
 #
 # $ipa_authentication::       Whether to install support for IPA authentication
 #
+# $apache_start_timeout::       Startup time for Apache
```
ekohl: The alignment is off. It'd also mention this is only for systemd and replace time with timeout.

chris1984:
@ekohl I added the tests I was working on to make sure we have coverage, but I see its not happy, can you do another review, please.

evgeni:
I think this is obsoleted by the actual Foreman fix in https://github.com/theforeman/foreman/pull/4908, no?

chris1984:
After talking to @ekohl we are closing this with https://github.com/theforeman/foreman/pull/4908

ehelms:
I disagree that #4908 obsoletes this. The point is that passenger *can* spend 600 seconds if it needs to before starting for whatever reason and httpd is configured to timeout before that 600 seconds. The application *should* start sooner than that, but the point of this issue is to protect against httpd timing out before passenger does. Be it slow hardware, bad code, etc.

Issue #593, opened 2017-09-25
Change key for apt repo not to explicitly reference pgp.mit.edu


tarioch:
Currently it uses 

`source => 'https://pgp.mit.edu/pks/lookup?op=get&search=0x6F8600B9563278F6',`

This is causing lots of issues as lately pgp.mit.edu gives me lots of errors and issues.

I would suggest to change it to not specify the source and depend on the ability of the apt module to lookup the key by the keyid itself.

PR #594, opened 2017-09-25
Drop explicit source for pgp key

tarioch:
Fixes GH-593


ekohl:
We've seen similar intermittent failures on Travis because of this so you're not the only one with issues. @mmoll do you think this is good to go in?

mmoll:
In general I'd say yes, but the tests need updating.

ekohl:
Thanks @tarioch! For future reference: we also accept PRs without a separate issue if they properly describe the problem.

PR #595, opened 2017-09-29
get data by ssl authentication

pronix:
using theforeman as data source(without puppetdb). using ssl keys of foreman proxy for authentication.
require also change in foreman file ./app/controllers/api/v2/hosts_controller.rb
```
add_smart_proxy_filters :facts, :features => Proc.new { FactImporter.fact_features }
```
to
```
add_smart_proxy_filters [:facts,:index], :features => Proc.new { FactImporter.fact_features }
```

### DIFF
--- a/lib/puppet/parser/functions/foreman.rb
+++ b/lib/puppet/parser/functions/foreman.rb
@@ -38,6 +38,7 @@
 #
 # Happy Foreman API-ing!
 
+require "yaml"
 require "net/http"
 require "net/https"
 require "uri"
@@ -51,6 +52,7 @@ module Puppet::Parser::Functions
     item          = args_hash["item"]
     search        = args_hash["search"]
     per_page      = args_hash["per_page"]     || "20"
+    use_tfmproxy  = args_hash["use_tfmproxy"] || false
     foreman_url   = args_hash["foreman_url"]  || "https://localhost" # defaults: all-in-one
     foreman_user  = args_hash["foreman_user"] || "admin"             # has foreman/puppet
     foreman_pass  = args_hash["foreman_pass"] || "changeme"          # on the same box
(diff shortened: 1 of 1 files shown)


lib/puppet/parser/functions/foreman.rb
```diff
@@ -64,14 +66,24 @@ module Puppet::Parser::Functions
...
 
-      http = Net::HTTP.new(uri.host, uri.port)
+      if use_tfmproxy
+        tfmproxy = YAML.load(File.read('/etc/foreman-proxy/settings.yml'))
+        uri = URI.parse(tfmproxy[:foreman_url])
+        http = Net::HTTP.new(uri.host, uri.port)
+        http.ca_file = tfmproxy[:foreman_ssl_ca]
+        http.cert = OpenSSL::X509::Certificate.new(File.read(tfmproxy[:foreman_ssl_cert]))
+        http.key = OpenSSL::PKey::RSA.new(File.read(tfmproxy[:foreman_ssl_key]), nil)
+        http.verify_mode = OpenSSL::SSL::VERIFY_NONE
```
ekohl: Why do you configure a CA but don't verify?

pronix:
hello @ekohl 
thanks for review.
peer verification is changed and use_tfmproxy can be a string for custom file.
but i still don't want to change current behaviour.

lib/puppet/parser/functions/foreman.rb
```diff
@@ -64,14 +66,29 @@ module Puppet::Parser::Functions
...
-      uri = URI.parse(foreman_url)
 
       req = Net::HTTP::Get.new(path)
-      req.basic_auth(foreman_user, foreman_pass)
       req['Content-Type'] = 'application/json'
       req['Accept'] = 'application/json'
 
-      http = Net::HTTP.new(uri.host, uri.port)
+      if use_tfmproxy
+        configfile = if use_tfmproxy.is_a? String and File.exist?(use_tfmproxy)
```
ekohl: I'd move the ```File.exist?``` to just before reading it. If a user specifies a non-existing file it shouldn't silently ignore it. The same is true for ```true``` but ```/etc/foreman-proxy/settings.yml``` doesn't exist.

pronix: you mean exception or just puts ?

ekohl: I think a hard exception is in order.

pronix:
hello @ekohl 
added files checking before read.

lib/puppet/parser/functions/foreman.rb
```diff
@@ -64,14 +66,36 @@ module Puppet::Parser::Functions
...
 
       req = Net::HTTP::Get.new(path)
-      req.basic_auth(foreman_user, foreman_pass)
       req['Content-Type'] = 'application/json'
       req['Accept'] = 'application/json'
 
-      http = Net::HTTP.new(uri.host, uri.port)
+      if use_tfmproxy
+        if use_tfmproxy.is_a? String
+          if File.exist?(use_tfmproxy)
```
ekohl: This block can be simplified I think:
```ruby
configfile = '/etc/foreman-proxy/settings.yaml'
configfile = use_tfmproxy if use_tfmproxy.is_a? String
raise Puppet::ParseError, "File #{configfile} not found while use_tfmproxy is enabled" unless File.exists?(configfile)
```

pronix:
thanks @ekohl . you are absolutely right.

lib/puppet/parser/functions/foreman.rb
```diff
@@ -64,14 +66,27 @@ module Puppet::Parser::Functions
...
+        http.cert = OpenSSL::X509::Certificate.new(File.read(tfmproxy[:foreman_ssl_cert]))
+        http.key = OpenSSL::PKey::RSA.new(File.read(tfmproxy[:foreman_ssl_key]), nil)
+        http.verify_mode = OpenSSL::SSL::VERIFY_PEER
+      else
+        uri = URI.parse(foreman_url)
+        http = Net::HTTP.new(uri.host, uri.port)
+        req.basic_auth(foreman_user, foreman_pass)
+      end
       http.use_ssl = true if uri.scheme == 'https'
       http.verify_mode = OpenSSL::SSL::VERIFY_NONE if http.use_ssl?
```
ekohl: I think this should move to the above ```else``` section.

pronix:
thanks @ekohl . fixed

mmoll:
I can't really this this, but it looks like if all fallback mechanisms are there, so merging...

mmoll:
merged, спасибо @pronix!

PR #596, opened 2017-10-09
Refs #18422 - Allow configuring Dynflow pool size

adamruzicka:


ekohl:
I assume this only makes sense when dynflow is in core which would require https://github.com/theforeman/puppet-foreman/pull/530, correct?

adamruzicka:
Not really, foreman-tasks inherits this value from foreman so even if there's no dynflow in core this can still be useful.

mmoll:
merged, díky @adamruzicka!

PR #597, opened 2017-10-12
metadata.json: bump allowed version of puppet-extlib to 3.0.0

mateusz-gozdek-sociomantic:
Recently, version 2.0.0 was released and it does not contain any
breaking changes. See https://forge.puppet.com/puppet/extlib/changelog#v200-2017-10-11
for details.

Currently puppet module list --tree throws following warnings because of
this:

Warning: Module 'puppet-extlib' (v2.0.0) fails to meet some dependencies:
  'theforeman-foreman' (v8.1.0) requires 'puppet-extlib' (>= 0.10.4 < 2.0.0)

### DIFF
--- a/metadata.json
+++ b/metadata.json
@@ -34,7 +34,7 @@
     },
     {
       "name": "puppet/extlib",
-      "version_requirement": ">= 0.10.4 < 2.0.0"
+      "version_requirement": ">= 0.10.4 < 3.0.0"
     }
   ],
   "requirements": [


mmoll:
This is looking fine, @ekohl I think we need a 8.1-stable branch, as in current master the discovery image downloading got removed, what do you think?

ekohl:
@mateusz-gozdek-sociomantic that was on my todo list but you beat me to it :)
@mmoll I do agree we made some larger changes. Would removal of the image downloading technically be a 9.0? In that case we can also fix the parameter renaming of admin_{username,password} to be prefixed with seed_

mmoll:
@ekohl one could see that removal as a breaking change, so 9.0 would be in order and give us freedom to break other things, too. 👍

mmoll:
merged, dzięki @mateusz-gozdek-sociomantic!

PR #598, opened 2017-10-13
Changes for 8.1-stable

mmoll:


mmoll:
While there are some enhancements in here that would require a minor version, I think these are all non-breaking for existing 1.16 (and older) installs.

ekohl:
Thanks!

PR #600, opened 2017-10-16
Add ability to set SSLProtocol for Apache vhost

ehelms:


ekohl:
The actual class parameter is missing. I also wonder if we should default to undef so you can set the default on ```apache::mod::ssl```. Another downside of your setting in the tests is that it means TLSv1.3 (currrently a draft) won't be enabled if available.

ehelms:
I meant to add the undef and missed the class parameters. Those are updated now. As for the test example, I am open to any string for the test itself or leaving it as undef. My eventual goal will be to document how you could install and lock to TLSv1.2.

ekohl:
Never mind, I assumed that your example was in a test for equality, not a parameter.

I do expect the tests to fail in the config_passenger_spec.rb file because we have to set every parameter there. We test without the foreman class and that breaks strict variables so we override them.

ehelms:
Tests are passing now

ekohl:
Thanks!

PR #601, opened 2017-10-18
add plugin foreman_column_view

dgoetz:
This PR adds support for the plugin foreman_column_view.

In my tests it worked fine, but there are some points I am not sure about:
* I used Puppet 4+ Datatypes in an excessive way for validation as some other code is already doing this but only in a much more simple way
* I fear using a hash of hashes will not work with kafo, only with Puppet
* I only added the default tests and one positive test for the configuration file, no negative one to fail if configuration is wrong
* Changes of the plugin's configuration require a restart of Foreman which I did not address

Please let me know if something requires a change in the PR.

### DIFF
--- a/manifests/plugin/column_view.pp
+++ b/manifests/plugin/column_view.pp
@@ -1,0 +1,15 @@
+# @summary Install the column view plugin and optionally manage the configuration file
+#
+# @param columns
+#   an hash of columns to add to the configuration
+#
+class foreman::plugin::column_view (
+  Hash[String, Hash] $columns = {},
+){
+  # https://projects.theforeman.org/issues/21398
+  assert_type(Hash[String, Foreman::Column_view_column], $columns)
+
+  foreman::plugin { 'column_view':
+    config => template('foreman/foreman_column_view.yaml.erb'),
+  }
+}
--- a/spec/classes/plugin/column_view_spec.rb
+++ b/spec/classes/plugin/column_view_spec.rb
@@ -1,0 +1,63 @@
+require 'spec_helper'
+
+describe 'foreman::plugin::column_view' do
+  include_examples 'basic foreman plugin tests', 'column_view'
+
+  context 'with columns hash architecture' do
+    let(:params) do
+      {
+        'columns' => {
+          'architecture' => {
+            'title' => 'Architecture',
+            'after' => 'last_report',
+            'content' => 'facts_hash["architecture"]',
+          }
+        }
+      }
+    end
+
+    it { is_expected.to contain_file('/etc/foreman/plugins/foreman_column_view.yaml').with_content(/.*:title: Architecture.*/) }
+
+    it { is_expected.to compile.with_all_deps }
+  end
+
+  context 'with columns hash architecture and console' do
+    let(:params) do
+      {
+        'columns' => {
+          'architecture' => {
+            'title' => 'Architecture',
+            'after' => 'last_report',
+            'content' => 'facts_hash["architecture"]',
+          },
+          'console' => {
+            'title' => 'Console',
+            'after' => '0',
+            'content' => 'link_to(_("Console"), "https://#{host.interfaces.first.name}.domainname", { :class => "btn btn-info" } )',
+	    'conditional' => ':bmc_available?',
+	    'eval_content' => 'true',
+	    'view' => ':hosts_properties',
+          }
+        }
+      }
+    end
+
+    it { is_expected.to contain_file('/etc/foreman/plugins/foreman_column_view.yaml').with_content(/.*:title: Console.*/) }
+
+    it { is_expected.to compile.with_all_deps }
+  end
+
+  context 'with columns hash broken' do
+    let(:params) do
+      {
+        'columns' => {
+          'broken' => {
+            'title' => 'Broken',
+          }
+        }
+      }
+    end
+
+    it { is_expected.to compile.and_raise_error(%r{broken}) }
+  end
+end
--- a/templates/foreman_column_view.yaml.erb
+++ b/templates/foreman_column_view.yaml.erb
@@ -1,0 +1,12 @@
+<%= ERB.new(File.read(File.expand_path("_header.erb",File.dirname(file)))).result(binding) -%>
+#
+# See tfm-rubygem-foreman_column_view-doc and /opt/theforeman/tfm/root/usr/share/gems/gems/foreman_column_view-0.3.0/README.md for more information
+:column_view:
+<% if @columns -%>
+<%   @columns.each do |column, hash| -%>
+  :<%= column %>:
+<%     hash.each do |key,value| -%>
+    :<%= key %>: <%= value %>
+<%     end -%>
+<%   end -%>
+<% end -%>
--- a/types/column_view_column.pp
+++ b/types/column_view_column.pp
@@ -1,0 +1,12 @@
+type Foreman::Column_view_column = Struct[
+  {
+    title => String[1],
+    after => String[1],
+    content => String[1],
+    Optional['conditional'] => String[1],
+    Optional['eval_content'] => String[1],
+    Optional['view'] => String[1],
+    Optional['width'] => String[1],
+    Optional['custom_method'] => String[1],
+  }
+]


manifests/plugin/column_view.pp
```diff
@@ -0,0 +1,26 @@
+# = Foreman Column View
+#
+# This class installs the column view plugin and optionally manages the configuration file
+#
+# === Parameters:
+#
+# $columns:: an hash of columns to add to the configuration
+#
+class foreman::plugin::column_view (
+  Optional[Hash[String,Hash[Enum['title','after','content','title','conditional','eval_content','view','width','custom_method'],String,3,9]]] $columns = undef
```
ekohl: I'd suggest looking at https://github.com/camptocamp/puppet-systemd/blob/master/types/servicelimits.pp since that results in an easier way. Another thing to consider is handling an empty hash instead of undef as default.

templates/foreman_column_view.yaml.erb
```diff
@@ -0,0 +1,10 @@
+<%= ERB.new(File.read(File.expand_path("_header.erb",File.dirname(file)))).result(binding) -%>
+#
+# See tfm-rubygem-foreman_column_view-doc and /opt/theforeman/tfm/root/usr/share/gems/gems/foreman_column_view-0.3.0/README.md for more information
+:column_view:
+<% @columns.each do |column, hash| -%>
```
ekohl: Won't this result in an error if its' empty?

manifests/plugin/column_view.pp
```diff
@@ -0,0 +1,26 @@
...
+){
+  $config_file = "${foreman::plugin_config_dir}/foreman_column_view.yaml"
+
+  foreman::plugin { 'column_view':
+  }
+
+  if $columns {
+    file { $config_file:
```
ekohl: You can also use the ```$config``` parameter of ```foreman::plugin``` to achieve the same to manage this file. It'll even match the filename in by default so no need to override ```$config_file```.

spec/classes/plugin/column_view_spec.rb
```diff
@@ -0,0 +1,21 @@
...
+        'columns' => {
+          'architecture' => {
+            'title' => 'Architecture',
+            'after' => 'last_report',
+            'content' => 'facts_hash["architecture"]',
+          }
+        }
+      }
+    end
+
```
ekohl: Mind adding a ```it { is_expected.to compile.with_all_deps }```?

dgoetz:
Thanks for the review!
I changed the template to work with columns undefined so I do no longer require the if clause and can use the parameter config. 
I moved the data type to a custom data type and added the check. 
I also addressed the lint error from the checks, but I have no idea why the test failed now with a timeout.

manifests/plugin/column_view.pp
```diff
@@ -0,0 +1,17 @@
...
+# This class installs the column view plugin and optionally manages the configuration file
+#
+# === Parameters:
+#
+# $columns:: an hash of columns to add to the configuration
+#
+class foreman::plugin::column_view (
+  Optional[Foreman::Column_view] $columns = undef
+){
+  $config_file = "${foreman::plugin_config_dir}/foreman_column_view.yaml"
```
ekohl: This variable is now unused.

types/column_view.pp
```diff
@@ -0,0 +1 @@
+type Foreman::Column_View = Hash[String,Hash[Enum['title','after','content','title','conditional','eval_content','view','width','custom_method'],String,3,9]]
```
ekohl: What's the 3, 9?

Another way of doing this is just having a ```Foreman::Column_View_Column``` and have the ```Hash[String, Foreman::Column_View_Column]``` type inside the manifest.

spec/classes/plugin/column_view_spec.rb
ekohl: (done)

manifests/plugin/column_view.pp
```diff
@@ -0,0 +1,17 @@
+# = Foreman Column View
+#
+# This class installs the column view plugin and optionally manages the configuration file
+#
+# === Parameters:
+#
+# $columns:: an hash of columns to add to the configuration
+#
+class foreman::plugin::column_view (
+  Optional[Foreman::Column_view] $columns = undef
```
ekohl: The current code would handle ```undef``` and ```{}``` in the same way. This means you always manage the file now. Is that intended? If that's the case then I'd prefer to default to ```{}``` and remove ```Optional``` but otherwise you can store the config in a variable:
```puppet
if $columns {
  $config = template('foreman/foreman_column_view.yaml.erb')
} else {
  $config = undef
}
```

Would you mind adding a trailing comma on this line as well?

dgoetz:
Thanks again!
Removed the undef and optional as it worked fine without it. Managing the file is intended as I checked that only the header is still a valid configuration for the plugin.
Removed the unused variable.
And moved the datatype from a hash with a minimum and maximum numbers of key value pairs (the 3,9) to a defined Struct. Added additional checks while doing so as it took me a while to understand how this should work. But now the checks all fail, while local check run successful (using the puppet 5.3.2 gem). Any idea?

ekohl:
It does show up with ```rake lint``` but I haven't been able to pin it down where exactly. I'll need to dive into it.

ekohl:
Looking into it comes from https://github.com/theforeman/kafo/blob/master/lib/kafo/data_type.rb but there is a ```Struct``` class. Somehow it's parsing ```Struct[``` instead so there might be an issue with the regex.

ekohl:
I opened https://projects.theforeman.org/issues/21398. Not sure how to easily proceed with this but maybe we can ignore the puppet class that includes it from linting or that particular test.

dgoetz:
Thanks for all the effort.

ekohl:
I wonder if we should accept the PR now with a simple type check while we work on the kafo issue.

dgoetz:
I do not see a need to rush. Column_view is easily installed and configured manually, I just recognized it is the only one missing in the module and installer of those I commonly use. Also it will not make it into 1.16 if merged now, or will it? Furthermore kafo can not handle arrays of hashes, so it will only handle installation and not configuration.

ekohl:
So the actual problem is https://github.com/theforeman/kafo/blob/93aaa70eff832fc6ccfbcbbf672f5cd775007420/lib/kafo/data_type_parser.rb#L5 which can't parse multi line types. A multi-line regex sounds like a bad idea so we probably need to find a better way. We can probably extract them using puppet-strings and cache them like we do with others.

manifests/plugin/column_view.pp
```diff
@@ -0,0 +1,15 @@
...
+#
+# This class installs the column view plugin and optionally manages the configuration file
+#
+# === Parameters:
+#
+# $columns:: an hash of columns to add to the configuration
+#
+class foreman::plugin::column_view (
+  Hash[String, Foreman::Column_view_column] $columns = {},
+){
```
ekohl, line 11: How about a workaround:

```suggestion
  Hash[String, Hash] $columns = {},
){
  # https://projects.theforeman.org/issues/21398
  assert_type(Hash[String, Foreman::Column_view_column], $columns)
```

spec/classes/plugin/column_view_spec.rb
```diff
@@ -0,0 +1,63 @@
...
+      {
+        'columns' => {
+          'broken' => {
+            'title' => 'Broken',
+          }
+        }
+      }
+    end
+
+    it { is_expected.to_not compile.with_all_deps }
```
ekohl, line 61: You can also test a bit more specific on the error:
```suggestion
    it { is_expected.to compile.and_raise_error(%r{title}) }
```

You can probably be more specific than just title.

manifests/plugin/column_view.pp
```diff
@@ -0,0 +1,18 @@
+# = Foreman Column View
+#
+# This class installs the column view plugin and optionally manages the configuration file
+#
+# === Parameters:
+#
+# $columns:: an hash of columns to add to the configuration
+#
```
ekohl, line 8: For simple parameter docs I've started to prefer the puppet-strings documentation style
```suggestion
# @summary Install the column view plugin and optionally manage the configuration file
#
# @param columns
#   an hash of columns to add to the configuration
#
```

dgoetz:
I had to adjust the checks because it needed a different string to match but Travis is green now! :smiley:

PR #602, opened 2017-10-24
Fixes #18757 - Handle dynflow service in foreman core

ekohl:
The dynflow service was moved into core with 1.16. That renamed it to dynflowd and means the service should be started by default.

For compatibility parameters are provided. This means we must move the service name detection to the jobs.pp file because that's the only place we know the actual value.

This replaces https://github.com/theforeman/puppet-foreman/pull/530. It's mostly the same, but split into two commits to make reviews easier. The first is a compatible refactor where the second is just the introduction of dynflow in core.

### DIFF
--- a/README.md
+++ b/README.md
@@ -51,6 +51,7 @@ previous stable release.
 
 ### Foreman version compatibility notes
 
+For Foreman 1.16 or older, please set the `dynflow_in_core` parameter to false.
 For Foreman 1.11 or older, please use the 5.x release series of this module.
 
 ## Types and providers
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -203,6 +203,9 @@
 #
 # $dynflow_pool_size::          How many workers should Dynflow use
 #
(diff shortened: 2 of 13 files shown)


ekohl:
@mmoll I'm looking at merging the RPM packaging PR. Is there anything that should be done for the deb side?

mmoll:
Something like theforeman/foreman-packaging#1549 would need to get added to the DEB package also...

mmoll:
If the service is also called "dynflowd" on Debian, this should work OOTB, AFAICT 👍

ekohl:
I'll have a look at adding dynflowd to Debian. It was my intention to unify them and simplify our instructions/packaging/configuration.

dLobatog:
https://github.com/theforeman/foreman-packaging/pull/1954

ekohl:
Rebased and updated to state it's 1.17+. Now that it's in both debs and rpms this can be merged once tests pass.

PR #604, opened 2017-10-25
Manage puppetdb_api_version config entry

treydock:


mmoll:
merged, thanks @treydock!

PR #606, opened 2017-10-26
Add unattended_url parameter

matonb:
Set `unattended_url` parameter in `foreman/settings.yaml`

Enables setting of unattended_url during install with foreman-installer

### DIFF
--- a/manifests/params.pp
+++ b/manifests/params.pp
@@ -15,6 +15,7 @@ class foreman::params {
   $receive_facts  = true
   # should foreman manage host provisioning as well
   $unattended     = true
+  $unattended_url = undef
   # Enable users authentication (default user:admin pw:changeme)
   $authentication = true
   # configure foreman via apache and passenger
--- a/spec/classes/foreman_config_spec.rb
+++ b/spec/classes/foreman_config_spec.rb
@@ -14,6 +14,7 @@ describe 'foreman::config' do
         it 'should set up the config' do
           should contain_concat__fragment('foreman_settings+01-header.yaml').
             with_content(/^:unattended:\s*true$/).
+            without_content(/^:unattended_url:/).
             with_content(/^:login:\s*true$/).
             with_content(/^:require_ssl:\s*true$/).
             with_content(/^:locations_enabled:\s*false$/).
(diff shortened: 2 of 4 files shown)


manifests/params.pp
```diff
@@ -15,6 +15,7 @@
   $receive_facts  = true
   # should foreman manage host provisioning as well
   $unattended     = true
+  $unattended_url = "http://${lower_fqdn}:8000"
```
ekohl: This is a non-default location I believe. When is it needed to override this and would it make sense to default to ```undef``` and only place it in the settings when the user overrides it?

matonb: I can't check right now, but I'm pretty sure that's the current default location.

But sure, I was in two minds of setting it to a default location, or not at all...

matonb: When would I use it?
Automated deployments of foreman puppet, my current need is to stand everything up from scratch with no user intervention

ekohl: I don't recall having to ever override this and I did do bare metal deployments. Now that might just be my memory but I think by default the unattended url is based on the regular foreman url on port 80. A plain foreman doesn't even listen on port 8000 I think.

matonb:
@ekohl Changed default to undef :)

matonb:
@ekohl Want me to squash the commits?

ekohl:
We can also do that on merge so you don't have to.

matonb:
Cool

mmoll:
a test would be nice.

mmoll:
please rebase.

spec/classes/foreman_config_spec.rb
```diff
@@ -161,6 +162,17 @@ class {'foreman':
...
 
+      describe 'with unattended_url' do
+        let :pre_condition do
+          "class {'foreman':
+            unattended_url => 'http://example.com',
+          }"
+        end
+
+        it { should contain_concat__fragment('foreman_settings+01-header.yaml').
+            with_content(/^:unattended_url:\s*http:\/\/example.com$/) }
```
ekohl: You can use ```with_content(%r{^:unattended_url:\s*http://example.com$})```

matonb: @ekohl 
  Thanks, I couldn't remember the regex pattern syntax and forgot to go back to it :)

mmoll:
merged, thanks @matonb!

PR #613, opened 2018-01-03
Added ability to search against hostenc 

ccheline:
We had a need to query hostenc and found piggybacking on this to be the best method for us.

Because the hostenc data comes back in a different payload element, I had to also add logic for specifying the element to read the data from.

(note that I am not a developer, nor experienced in puppet or foreman so there likely is a much better way to complete this and I am glad to be educated if I am approaching this wrong)

### DIFF
--- a/lib/puppet/parser/functions/foreman.rb
+++ b/lib/puppet/parser/functions/foreman.rb
@@ -60,12 +60,19 @@ module Puppet::Parser::Functions
     timeout       = (args_hash['timeout']      || 5).to_i
 
     # extend this as required
-    searchable_items = %w{ environments fact_values hosts hostgroups puppetclasses smart_proxies subnets }
+    searchable_items = %w{ environments fact_values hosts hostenc hostgroups puppetclasses smart_proxies subnets }
+
     raise Puppet::ParseError, "Foreman: Invalid item to search on: #{item}, must be one of #{searchable_items.join(", ")}." unless searchable_items.include?(item)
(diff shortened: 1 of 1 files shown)


ekohl:
In which case would you need this? Aren't the various values already available as facts?

ccheline:
We use it to query the ENC data from a different host (such as other hosts in the same hostgroup).

For example, in our dynamic HAProxy config:
```
 $hosts_search = {
         item         => 'hosts',
         search       => "location=${location} and params.vip != null",
         per_page     => '1000',
         foreman_url  => $foreman_api_url,
         foreman_user => $foreman_api_user,
         foreman_pass => $foreman_api_password,
         timeout => 60
  }
  $hosts = foreman($hosts_search)
....
  $maps = $hosts['results'].map |$value| {
    $membername = $value['name']
    $enc_search = {
           item         => 'hostenc',
           search       => $membername,
           per_page     => '100',
           foreman_url  => $foreman_api_url,
           foreman_user => $foreman_api_user,
           foreman_pass => $foreman_api_password,
           timeout => 60
    }
    $enc = foreman($enc_search)
    ...
    $vip = split($enc['data']['parameters']['vip'], ',')
```

sean797:
@ccheline interesting use-case... I like it.

I wonder if this should be its own function? One of the reasons is I don't really feel like `search` is a fitting parameter for hostname.

ccheline:
That makes sense because it is a different API URL and different result set.
Adding this as something more like a host detail search, other host details
- other than enc - could be easily added (like status and
vm_compute_attributes).

I can mock one up if you like.


-----
colin

ekohl:
I'm going to close this per my previous comment and the lack of activity.

PR #614, opened 2018-01-09
Refs #22165 - Add installer support for disabling hsts

tbrisker:


templates/settings.yaml.erb
```diff
@@ -30,6 +30,9 @@
 :ssl_ca_file: <%= scope.lookupvar("foreman::client_ssl_ca") %>
 :ssl_priv_key: <%= scope.lookupvar("foreman::client_ssl_key") %>
 
+# HSTS setting
+:disable_hsts: <%= scope.lookupvar("foreman::disable_hsts") %>
```
ekohl: IMHO this is a badly named variable: it should always be positive (so ```enable_hsts`` defaulting to ```true```)

tbrisker: Still hasn't been merged in core - see related PR (and request change there so noone merges it)

tbrisker: I thought about this and I don't think we should change it. If we default to true here and make the name e.g. `enable_hsts` that will cause any users who haven't updated their settings.yaml with the new option to default to false, thus disabling HSTS for them.

tbrisker:
@ekohl updated to match the change in core

tbrisker:
@ekohl core has been merged, rebased this on master.

manifests/init.pp
```diff
@@ -206,6 +206,9 @@
 # $jobs_service::               Name of the service for running the background Dynflow executor
 #
 # $dynflow_in_core::            Whether the Dynflow executor service is provided by Foreman or tasks
+#
+# $hsts_enabled::               Disable HSTS enforcement in https requests
```
ekohl: Disable -> Enable in the doc?

tbrisker: done, thanks.

mmoll:
merged, thanks @tbrisker!

PR #616, opened 2018-01-12
Use puppet4 functions-api

juliantodt:
The legacy puppet3 functions-api should be avoided.
This migrates the existing functions to the new api and uses it in the module. It also adds the tests for the new functions based on those for the olds.
We should keep the legacy functions for now to allow compatibility for those using the functions outside out this module.

### DIFF
--- a/lib/puppet/functions/foreman/ensure_value_in_string.rb
+++ b/lib/puppet/functions/foreman/ensure_value_in_string.rb
@@ -1,0 +1,28 @@
+# Returns an string with appended values to the end if they were not present in original string.
+#     Prototype:
+#       ensure_value_in_string(string, array, separator = ',')
+# Where string is the original string, array is array of additional values to append,
+# separator is optional specifying delimiter and defaults to a comma.
+#     For example:
+# Given the following statements:
+#       ensure_value_in_string('one,two', ['two', 'three'])
+# The result will be as follows:
+#       'one,two,three'
+# You can specify you own separator as a third argument
+#       ensure_value_in_string('one,two', ['two', 'three'], ', ')
+# results in
+#       'one,two, three'
+Puppet::Functions.create_function(:'foreman::ensure_value_in_string') do
+  dispatch :ensure_value_in_string do
+    required_param 'String', :string
+    required_param 'Array', :adding
+    optional_param 'String', :separator
+  end
+
+  def ensure_value_in_string(string, adding, separator = ',')
+    existing = string.split(separator.strip).map(&:strip)
+    to_add = adding - existing
+
+    ([ string.empty? ? nil : string ] + to_add).compact.join(separator)
+  end
+end
--- a/lib/puppet/functions/foreman/ensure_value_in_string.rb
+++ b/lib/puppet/functions/foreman/ensure_value_in_string.rb
@@ -15,7 +15,7 @@
 Puppet::Functions.create_function(:'foreman::ensure_value_in_string') do
   dispatch :ensure_value_in_string do
     required_param 'String', :string
-    required_param 'Array', :adding
+    required_param 'Array[String]', :adding
     optional_param 'String', :separator
   end
 
--- a/lib/puppet/functions/foreman/foreman.rb
+++ b/lib/puppet/functions/foreman/foreman.rb
@@ -30,8 +30,7 @@
 #           This defaults to five seconds.
 #
 # Then, use a variable to capture its output:
-# $f = ['hosts', 'hostgroup=Grid', '20', 'https://foreman', 'my_api_foreman_user', 'my_api_foreman_pass']
-# $hosts = foreman(*$f)
+# $hosts = foreman('hosts', 'hostgroup=Grid', '20', 'https://foreman', 'my_api_foreman_user', 'my_api_foreman_pass')
 #
 # Note: If you're using this in a template, you may be receiving an array of
 # hashes. So you might need to use two loops to get the values you need.
@@ -46,22 +45,20 @@ require "timeout"
 
 Puppet::Functions.create_function(:'foreman::foreman') do
   dispatch :foreman do
-    required_param 'String', :item
+    required_param 'Enum["environments", "fact_values", "hosts", "hostgroups", "puppetclasses", "smart_proxies", "subnets"]', :item
     required_param 'String', :search
-    optional_param 'Variant[Integer, String]', :per_page
-    optional_param 'String', :foreman_url
+    optional_param 'Variant[Integer[0], Pattern[/\d+/]]', :per_page
+    optional_param 'Stdlib::Httpurl', :foreman_url
     optional_param 'String', :foreman_user
     optional_param 'String', :foreman_pass
-    optional_param 'Integer', :timeout
+    optional_param 'Integer[0]', :timeout
     optional_param 'Variant[String, Array, Hash, Boolean]', :filter_result
     optional_param 'Variant[Boolean, String]', :use_tfmproxy
   end
 
   def foreman(item, search, per_page = "20", foreman_url = "https://localhost", foreman_user = "admin", foreman_pass = "changeme", timeout = 5, filter_result = false, use_tfmproxy = false)
     # extend this as required
-    searchable_items = %w{ environments fact_values hosts hostgroups puppetclasses smart_proxies subnets }
-    raise Puppet::ParseError, "Foreman: Invalid item to search on: #{item}, must be one of #{searchable_items.join(", ")}." unless searchable_items.include?(item)
-    raise Puppet::ParseError, "Foreman: Invalid filter_result: #{filter_result}, must be a String or an Array or false" unless filter_result.is_a? String or filter_result.is_a? Array or filter_result.is_a? Hash or filter_result == false
+    raise Puppet::ParseError, "Foreman: Invalid filter_result: #{filter_result}, must not be boolean true" if filter_result == true
 
     begin
       path = URI.escape("/api/#{item}?search=#{search}&per_page=#{per_page}")
--- a/lib/puppet/functions/foreman/foreman.rb
+++ b/lib/puppet/functions/foreman/foreman.rb
@@ -1,16 +1,15 @@
 # Query Foreman
 #
-# The foreman() parser takes a hash value with parameters to execute the query.
+# The foreman() parser takes various parameters to execute a query against the foreman-api.
 #
-# To use foreman(), first create a hash. This sample hash will contain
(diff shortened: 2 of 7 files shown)


lib/puppet/functions/foreman/ensure_value_in_string.rb
```diff
@@ -0,0 +1,28 @@
...
+# The result will be as follows:
+#       'one,two,three'
+# You can specify you own separator as a third argument
+#       ensure_value_in_string('one,two', ['two', 'three'], ', ')
+# results in
+#       'one,two, three'
+Puppet::Functions.create_function(:'foreman::ensure_value_in_string') do
+  dispatch :ensure_value_in_string do
+    required_param 'String', :string
+    required_param 'Array', :adding
```
ekohl: Could be ```Array[String]```

lib/puppet/functions/foreman/ensure_value_in_string.rb
```diff
@@ -0,0 +1,28 @@
+# Returns an string with appended values to the end if they were not present in original string.
```
ekohl: Now that we have structured facts I'm wondering if we should make https://github.com/theforeman/puppet-foreman/blob/3abb8b1ddb083c289e682a040f6e46e745f2cf4a/lib/facter/sssd.rb return one ```sssd``` fact (and one ```ipa``` fact) with all the properties. That would allow us to return a proper array and do this in pure puppet code without a function.

Doesn't have to be part of this PR but if we're cleaning up we might as well do it properly.

lib/puppet/functions/foreman/foreman.rb
```diff
@@ -0,0 +1,144 @@
...
+require "uri"
+require "timeout"
+
+Puppet::Functions.create_function(:'foreman::foreman') do
+  dispatch :foreman do
+    required_param 'Hash', :args_hash
+  end
+
+  def foreman(args_hash)
+    # parse an args hash
```
ekohl: I'm wondering if we should move these to proper top level parameters rather than a hash. Especially now that the Puppet 4 syntax allows passing in a hash with extra args (```foreman(* => data)```).

lib/puppet/functions/foreman/smartvar.rb
```diff
@@ -0,0 +1,38 @@
...
+
+require "net/http"
+require "net/https"
+require "uri"
+require "timeout"
+
+Puppet::Functions.create_function(:'foreman::smartvar') do
+  def smartvar(var)
+    #URL to query
+    foreman_url  = "http://foreman"
```
ekohl: Could you make these proper parameters while you're at it?

lib/puppet/functions/foreman/smartvar.rb
```diff
@@ -0,0 +1,38 @@
...
+# Foreman holds all the value names and their possible values,
+# this function simply ask foreman for the right value for this host.
+
+require "net/http"
+require "net/https"
+require "uri"
+require "timeout"
+
+Puppet::Functions.create_function(:'foreman::smartvar') do
+  def smartvar(var)
```
ekohl: Var could be checked for the type ```String```

lib/puppet/functions/foreman/ensure_value_in_string.rb
juliantodt: I would prefer it if we could stick to migrating the functions in this PR as I'm not too familiar with these parts.

ekohl: https://github.com/theforeman/puppet-foreman/pull/618 should remove the need for this function altogether.

lib/puppet/functions/foreman/foreman.rb
juliantodt: Only found `foreman(*$data)` for arrays in the puppet documentary, is that what you mean?

ekohl: Quite possible. I'm still somewhat new to those aspects of Puppet 4.

juliantodt:
Let me know if you think another order of the parameters of `foreman::foreman()` would be better.

lib/puppet/functions/foreman/foreman.rb
```diff
@@ -0,0 +1,141 @@
...
+require "net/http"
+require "net/https"
+require "uri"
+require "timeout"
+
+Puppet::Functions.create_function(:'foreman::foreman') do
+  dispatch :foreman do
+    required_param 'String', :item
+    required_param 'String', :search
+    optional_param 'Variant[Integer, String]', :per_page
```
ekohl: Would this stricter type work as well? ```Variant[Integer[0], Pattern[/\d+/]]```

lib/puppet/functions/foreman/foreman.rb
```diff
@@ -0,0 +1,141 @@
...
+    optional_param 'Integer', :timeout
+    optional_param 'Variant[String, Array, Hash, Boolean]', :filter_result
+    optional_param 'Variant[Boolean, String]', :use_tfmproxy
+  end
+
+  def foreman(item, search, per_page = "20", foreman_url = "https://localhost", foreman_user = "admin", foreman_pass = "changeme", timeout = 5, filter_result = false, use_tfmproxy = false)
+    # extend this as required
+    searchable_items = %w{ environments fact_values hosts hostgroups puppetclasses smart_proxies subnets }
+    raise Puppet::ParseError, "Foreman: Invalid item to search on: #{item}, must be one of #{searchable_items.join(", ")}." unless searchable_items.include?(item)
+    raise Puppet::ParseError, "Foreman: Invalid filter_result: #{filter_result}, must be a String or an Array or false" unless filter_result.is_a? String or filter_result.is_a? Array or filter_result.is_a? Hash or filter_result == false
```
ekohl: Isn't this redundant now?

lib/puppet/functions/foreman/foreman.rb
```diff
@@ -0,0 +1,141 @@
...
+    optional_param 'String', :foreman_user
+    optional_param 'String', :foreman_pass
+    optional_param 'Integer', :timeout
+    optional_param 'Variant[String, Array, Hash, Boolean]', :filter_result
+    optional_param 'Variant[Boolean, String]', :use_tfmproxy
+  end
+
+  def foreman(item, search, per_page = "20", foreman_url = "https://localhost", foreman_user = "admin", foreman_pass = "changeme", timeout = 5, filter_result = false, use_tfmproxy = false)
+    # extend this as required
+    searchable_items = %w{ environments fact_values hosts hostgroups puppetclasses smart_proxies subnets }
```
ekohl: You can make ```item``` and ```Enum[...]``` with these and remove this + the next line

lib/puppet/functions/foreman/smartvar.rb
```diff
@@ -0,0 +1,38 @@
...
+
+require "net/http"
+require "net/https"
+require "uri"
+require "timeout"
+
+Puppet::Functions.create_function(:'foreman::smartvar') do
+  dispatch :smartvar do
+    required_param 'String[1]', :var
+    optional_param 'String', :foreman_url
```
ekohl: There is a [Stdlib::Httpurl](https://github.com/puppetlabs/puppetlabs-stdlib#stdlibhttpurl)

lib/puppet/functions/foreman/foreman.rb
```diff
@@ -0,0 +1,141 @@
...
+#                 attributes given in array
+#                 in case of an given hash foreman() returns an array of hashes selecting only
+#                 attribute keys given in hash renamed to values of given keys. This can be used
+#                 to rename keys in result
+# 'timeout' is the Foreman request timeout in seconds as an integer.
+#           This defaults to five seconds.
+#
+# Then, use a variable to capture its output:
+# $f = ['hosts', 'hostgroup=Grid', '20', 'https://foreman', 'my_api_foreman_user', 'my_api_foreman_pass']
+# $hosts = foreman(*$f)
```
ekohl: I think we should document this syntax only as an alternative (or not at all).

lib/puppet/functions/foreman/foreman.rb
```diff
@@ -0,0 +1,141 @@
...
+
+Puppet::Functions.create_function(:'foreman::foreman') do
+  dispatch :foreman do
+    required_param 'String', :item
+    required_param 'String', :search
+    optional_param 'Variant[Integer, String]', :per_page
+    optional_param 'String', :foreman_url
+    optional_param 'String', :foreman_user
+    optional_param 'String', :foreman_pass
+    optional_param 'Integer', :timeout
```
ekohl: Could also specify 0 as a lower bound

lib/puppet/functions/foreman/foreman.rb
```diff
@@ -0,0 +1,141 @@
...
+require "net/https"
+require "uri"
+require "timeout"
+
+Puppet::Functions.create_function(:'foreman::foreman') do
+  dispatch :foreman do
+    required_param 'String', :item
+    required_param 'String', :search
+    optional_param 'Variant[Integer, String]', :per_page
+    optional_param 'String', :foreman_url
```
ekohl: There is a [Stdlib::Httpurl](https://github.com/puppetlabs/puppetlabs-stdlib#stdlibhttpurl)

lib/puppet/functions/foreman/foreman.rb
juliantodt: It would fail if you pass `filter_result = true`. Puppet would accept it because its a boolean but we can't use it.

ekohl: Since it's optional already, would it make sense not to accept a boolean at all then or simplify the error here to only reject true?

juliantodt: `false` is the default value for this parameters, so it should accept Booleans, but we can obviously simplify the error.

ekohl:
Due to https://github.com/theforeman/puppet-foreman/pull/618 part of this is no longer needed. Due to the time I'm going to resolve the conflicts and merge it.

ekohl:
I just noticed this doesn't remove the old functions. Do we want to do that?

mmoll:
Let's do a relase with the old stuff included and then remove it. 👍

ekohl:
> Let's do a relase with the old stuff included and then remove it. 👍

I can create a PR with the conflicts resolved and include it for 9.0.0. Is that what you mean?

mmoll:
> I can create a PR with the conflicts resolved and include it for 9.0.0. Is that what you mean?

exactly.

ekohl:
https://github.com/theforeman/puppet-foreman/pull/623

PR #623, opened 2018-01-29
Use puppet4 functions-api

ekohl:
Closes GH-616


mmoll:
merged, danke @juliantodt!

PR #624, opened 2018-02-05
Remove test and development database declarations

ehelms:
These database declarations don't make sense in a production
environment which the module is focused on. If these environment
databases are desired then they can be added back properly. This
also serves to fix an odd issue in production with Dynflow.


ekohl:
We have a ```rails_env``` variable. Should we actually reuse that in this file instead of the hardcoded ```production```?

ehelms:
@ekohl Thats a great idea, might open it up for re-use as well in development module

ehelms:
@ekohl One issue with this being the solution for nightly... we'd need this in 1.17 for Katello 3.6 to pass testing.

PR #630, opened 2018-03-25
Reduce PARALLEL_TEST_PROCESSORS to 8

ekohl:
Tests are starting to fail in Travis without a clear error. We've seen
this before and reducing the concurrency has helped.

### DIFF
--- a/.sync.yml
+++ b/.sync.yml
@@ -5,7 +5,7 @@
     - docker/debian-8
   env:
     global:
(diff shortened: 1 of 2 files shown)


mmoll:
merged, bedankt @ekohl! 
Maybe we should set this to 8 globally per modulesync?

ekohl:
@mmoll I think that could make sense given all the modules we do set it, it's lowered to 8 where the other modules don't have enough spec files to be parallel.

PR #631, opened 2018-03-27
Refs #22559 - Add parameters for structured logging

ekohl:


ekohl:
This will also need code to install packages if any logger other than `file` is used.

manifests/params.pp
```diff
@@ -253,6 +253,8 @@
 
   # Application logging
   $logging_level = 'info'
+  $logging_type = 'file'
+  $logging_layout = 'pattern'
```
lzap: Default was `multiline_pattern`, but I am fine with switching to `pattern`. The difference is that the former adds ` |` characters at the beginning of each line, which is useful for development but in syslog/journald this might not be the best thing to do. So +1.

manifests/params.pp
```diff
@@ -253,6 +253,8 @@
 
   # Application logging
   $logging_level = 'info'
+  $logging_type = 'file'
```
lzap: This is indeed the most sane default - everything stays the same for production instance.

But logging to files is considered bad practice, except some corner cases (e.g. apache httpd access logs) using standard logging mechanisms should be preferred. I was hoping to default to journald (if present), syslog (otherwise) and just dropping `/etc/rsyslog.d/foreman.conf` configuration file and alerting daemon to reload (similarly for syslog-ng). If that's too complex in Puppet, we can just leave file logging for now.

ekohl: > This is indeed the most sane default - everything stays the same for production instance.

That was indeed exactly what I thought, also because I can't quite oversee what else is affected. Start with configuration options and safe defaults.

If journald is just drop in, we could use facts to determine if systemd is used. I'm hesitant to start managing syslog since that can be fairly complex and more up to the admin to set up. That's why I'd fall back to `file` if systemd is not used. My plan was to revisit the defaults later.

manifests/params.pp
ekohl: Was `multiline_pattern` also used before structured logging was introduced? I thought it was `pattern` so stayed with it here.

manifests/params.pp
ares: The question is how or when do we find out it's now safe to change the default, too bad didn't have this in the beginning of 1.18 dev window. Would it make sense to change the default just after we branch out?

I agree that we shouldn't manage syslog but dropping a oneliner to `/etc/rsyslog.d/foreman.conf` would be acceptable I think?

ekohl: > The question is how or when do we find out it's now safe to change the default, too bad didn't have this in the beginning of 1.18 dev window. Would it make sense to change the default just after we branch out?

For me it depends on whether journald is a drop in replacement:
* Does journald also log to a file?
* Can foreman-debug extract logs from journald?

> I agree that we shouldn't manage syslog but dropping a oneliner to /etc/rsyslog.d/foreman.conf would be acceptable I think?

The difficulty is that we can't rely on any syslog implementation. I can imagine users are using `syslog-ng` for example. Then we also need to support that. Suddenly we have a few extra parameters for things that IMHO belong in a profile module. Will we manage the syslog daemon as well for example. We've worked hard to keep that out of puppet-foreman. It just doesn't fit the design. I really don't want to touch syslog at all.

lzap: Unfortunately, journald main idea is all around binary journal that can be sealed - there is no file logging in journald. It can however forward to syslog and that's what it does by default on most systems these days.

I will create patch for foreman-debug to collect the logs from there for sure.

If we don't want to touch syslog, then another option is to ditch `/var/log/foreman/production.log` completely. An installer could archive/rename all files in pattern `production*.log*` and drop a dummy file with short explanation that this is not used. I don't think this is a bad experience for users after upgrade at all. Things change.

Structured logging patch is not only about structured logging, but also about collecting things from other parties like backend systems or smart proxy. With files, we can't achieve that, but having things in journald allows us to start working towards this goal.

ekohl:
I've expanded the PR to include acceptance tests. This told me we don't have the `foreman-journald` package in RPM nightly yet so I'm going to focus on that. Locally the Debian 9 tests looked good. I'm still running a last test run because I had an incorrect text in my test, but expecting it to pass.

manifests/params.pp
ekohl: To me it sounds like we should talk to our users about their expectations because nowadays I'm somewhat removed from actual deployments. Should we open a topic on community.theforeman.org?

ekohl:
Tests now pass locally:

```
debian-9-x64.example.com 17:54:52$ /bin/sh -c journalctl\ -u\ apache2
  -- Logs begin at Tue 2018-03-27 15:47:49 UTC, end at Tue 2018-03-27 15:54:51 UTC. --
  Mar 27 15:49:53 debian-9-x64.example.com systemd[1]: Starting The Apache HTTP Server...
  Mar 27 15:49:53 debian-9-x64.example.com systemd[1]: Started The Apache HTTP Server.
  Mar 27 15:53:48 debian-9-x64.example.com systemd[1]: Stopping The Apache HTTP Server...
  Mar 27 15:53:48 debian-9-x64.example.com systemd[1]: Stopped The Apache HTTP Server.
  Mar 27 15:53:48 debian-9-x64.example.com systemd[1]: Starting The Apache HTTP Server...
  Mar 27 15:53:48 debian-9-x64.example.com systemd[1]: Started The Apache HTTP Server.
  Mar 27 15:54:04 debian-9-x64.example.com foreman[18917]: [I|dyn|] start terminating throttle_limiter...
  Mar 27 15:54:04 debian-9-x64.example.com foreman[18917]: [I|dyn|] start terminating client dispatcher...
  Mar 27 15:54:04 debian-9-x64.example.com foreman[18917]: [I|dyn|] stop listening for new events...
  Mar 27 15:54:04 debian-9-x64.example.com foreman[18917]: [I|dyn|] start terminating clock...
  Mar 27 15:54:14 debian-9-x64.example.com foreman[18971]: [I|app|] Started HEAD "/" for 127.0.0.1 at 2018-03-27 15:54:14 +0000
  Mar 27 15:54:14 debian-9-x64.example.com foreman[18971]: [I|app|cc02b] Processing by DashboardController#index as HTML
  Mar 27 15:54:14 debian-9-x64.example.com foreman[18971]: [I|app|cc02b] Redirected to https://debian-9-x64.example.com/
  Mar 27 15:54:14 debian-9-x64.example.com foreman[18971]: [I|app|cc02b] Filter chain halted as #<Proc:0x00564bec4d09e8@/usr/share/foreman/vendor/ruby/2.3.0/gems/actionpack-5.1.4/lib/action_controller/metal/force_ssl.rb:65> rendered or redirected
  Mar 27 15:54:14 debian-9-x64.example.com foreman[18971]: [I|app|cc02b] Completed 301 Moved Permanently in 31ms (ActiveRecord: 8.1ms)
  Mar 27 15:54:14 debian-9-x64.example.com foreman[18971]: [I|app|] Started HEAD "/" for 127.0.0.1 at 2018-03-27 15:54:14 +0000
  Mar 27 15:54:14 debian-9-x64.example.com foreman[18971]: [I|app|54c81] Processing by DashboardController#index as HTML
  Mar 27 15:54:14 debian-9-x64.example.com foreman[18971]: [I|app|54c81] Redirected to https://debian-9-x64.example.com/users/login
  Mar 27 15:54:14 debian-9-x64.example.com foreman[18971]: [I|app|54c81] Filter chain halted as :require_login rendered or redirected
  Mar 27 15:54:14 debian-9-x64.example.com foreman[18971]: [I|app|54c81] Completed 302 Found in 42ms (ActiveRecord: 16.3ms)

debian-9-x64.example.com executed in 0.05 seconds
      should match /Redirected to https:\/\/debian-9-x64.example.com\/users\/login/
```

manifests/params.pp
lzap: Yep, started it: https://community.theforeman.org/t/default-logging-in-foreman-file-syslog-or-journald/8675

lzap:
Shall I re-review now? 2/3 merged atm.

ekohl:
@lzap previously the tests on Debian failed because it wasn't logging in a situation where it first used file logging and then moved to journald. I still want to reproduce this.

ekohl:
@lzap looks like we have the same issue on CentOS in containers. Going to test it on a VM on my desktop now to see if I can reproduce it.

lzap:
Oh interesting, no journald in container? It was designed to be container friendly, let's see.

ekohl:
@lzap when i manually ran only the journald test it did work for me locally. On Travis it first installs a basic Foreman with file logging. It could be that the upgrade doesn't work.

ekohl:
@lzap I know what the problem is. For the redirect log line I relied on the apache preloader to do a request but that isn't sent when you reload the app by touching `restart.txt`. I'll modify the tests to explicitly do a curl request to trigger a log line.

ekohl:
Ready for re-review now.

ekohl:
(Debian tests are now failing because nightly doesn't install).

mmoll:
this needs a rebase now

pondrejk:
Tested various combinations of --foreman-logging-layout --foreman-logging-level and --foreman-logging-type on rhel 7.4 based on docs in https://github.com/theforeman/theforeman.org/pull/1053/files
Some claims form that document that document I didn't verify, but that may be my misunderstanding of what parts are in progress. In short:

-- journald part works all right with all three patterns, foreman fields are logged as expected, filtering and persistent storage works
-- systemd setting for me just keeps logging to journald, logs land in /var/log/messages but that is I suppose due to journald default. So so far, these two options have the same effect, I also don't see the default facility setting mentioned in the doc
-- I din't find a way to set json_items in installer, is it something only configurable in the config file?

mmoll:
please rebase :)

pondrejk:
@ekohl 
> Do you mean syslog here?

Yes syslog, sorry

> We don't have that setting, but it should be easy to add. Shall I modify the PR?

To me it makes sense to have yaml-configurable options also as installer parameters, so I'd say yes.

mmoll:
Let's bring this in as-is for now :+1:

mmoll:
merged, bedankt @ekohl!

lzap:
Thanks a bunch @ekohl !

I realized one thing, when journald or syslog is set, we should configure logrotate NOT to poke httpd because it's not necessary anymore.

PR #632, opened 2018-03-27
Run acceptance tests on Debian 9 instead of Debian 8

ekohl:


mmoll:
merged, bedankt @ekohl!

PR #634, opened 2018-04-09
Handle releases/ properly for yum plugins repo

ekohl:
This also tightens the defined type by using data types.


mmoll:
merged, bedankt @ekohl!

PR #635, opened 2018-04-09
Add a basic acceptance test

ekohl:


ekohl:
Currently fails on Debian nightly installation. This is something we can see more often, but I guess that's just more reason to keep nightlies working :)

mmoll:
merged, bedankt @ekohl!

PR #636, opened 2018-04-09
Add classes for hammer cli commands

ekohl:
This is not complete and there are more plugins, but this is a good start.

### DIFF
--- a/manifests/cli/discovery.pp
+++ b/manifests/cli/discovery.pp
@@ -1,0 +1,10 @@
+# = Hammer Discovery plugin
+#
+# This installs the Discovery plugin for Hammer CLI
+#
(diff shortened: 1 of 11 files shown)


mmoll:
I didn't test this, but think this doesn't work on Debian, where no underscores are allowed in package names.

ekohl:
I pushed an updated version with acceptance tests - haven't ran them myself yet, will check when I get back home.

mmoll:
:broken_heart: but almost there :)

ekohl:
I do believe it should be fixed now. Let's wait to see if Travis agrees :)

mmoll:
merged, bedankt @ekohl!

PR #637, opened 2018-04-10
Fixes #23101 - add telemetry options

ares:
@ekohl this would be really good to have in 1.18, otherwise every installer run disables telemetry configuration that users configured, cc @lzap 

note that the biggest past is indentation of comments

### DIFF
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -2,212 +2,226 @@
 #
 # === Parameters:
 #
-# $admin_username::             Username for the initial admin user
+# $admin_username::               Username for the initial admin user
 #
-# $admin_password::             Password of the initial admin user, default is randomly generated
+# $admin_password::               Password of the initial admin user, default is randomly generated
 #
-# $admin_first_name::           First name of the initial admin user
+# $admin_first_name::             First name of the initial admin user
 #
-# $admin_last_name::            Last name of the initial admin user
+# $admin_last_name::              Last name of the initial admin user
 #
-# $admin_email::                E-mail address of the initial admin user
+# $admin_email::                  E-mail address of the initial admin user
 #
-# $db_manage::                  if enabled, will install and configure the database server on this host
+# $db_manage::                    if enabled, will install and configure the database server on this host
 #
-# $db_type::                    Database 'production' type
+# $db_type::                      Database 'production' type
 #
-# $email_delivery_method::      Email delivery method
+# $email_delivery_method::        Email delivery method
 #
-# $email_smtp_address::         SMTP server hostname, when delivery method is SMTP
+# $email_smtp_address::           SMTP server hostname, when delivery method is SMTP
 #
-# $email_smtp_port::            SMTP port
+# $email_smtp_port::              SMTP port
 #
-# $email_smtp_domain::          SMTP HELO domain
+# $email_smtp_domain::            SMTP HELO domain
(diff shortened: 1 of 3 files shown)


manifests/init.pp
```diff
@@ -303,6 +317,13 @@
   Enum['none', 'plain', 'login', 'cram-md5'] $email_smtp_authentication = $::foreman::params::email_smtp_authentication,
   Optional[String] $email_smtp_user_name = $::foreman::params::email_smtp_user_name,
   Optional[String] $email_smtp_password = $::foreman::params::email_smtp_password,
+  Optional[String] $telemetry_prefix = $::foreman::params::telemetry_prefix,
+  Boolean $telemetry_prometheus_enabled = $::foreman::params::telemetry_prometheus_enabled,
+  Boolean $telemetry_statsd_enabled = $::foreman::params::telemetry_statsd_enabled,
+  Optional[String] $telemetry_statsd_host = $::foreman::params::telemetry_statsd_host
+  Optional[Enum['statsd', 'statsite', 'datadog']] $telemetry_statsd_protocol = $::foreman::params::telemetry_statsd_protocol
```
lzap: I'd actually rather not add this, it can overwhelm users with options and we actually do not use any extensions, I think it might be better idea simply to remove this setting. Or at least write "keep statsd" in documentation comment.

manifests/init.pp
```diff
@@ -303,6 +317,13 @@
...
+  Boolean $telemetry_logger_enabled = $::foreman::params::telemetry_logger_enabled
+  Optional[Enum['DEBUG', 'INFO', 'WARN', 'ERROR', 'FATAL']] $telemetry_logger_level = $::foreman::params::telemetry_logger_level
```
lzap: This was meant to be for development, maybe at least mention this in docs comment. No practical use.

manifests/init.pp
ares: It's already hidden under advanced params. If we do not expose it and user changes protocol, on next upgrade, it would be overriden back. Not sure how you mean "keep statsd"

manifests/init.pp
ares: the same as above, if we don't expose and people override, it would get reset every time installer runs

ares:
Updated parameters docs according to suggestions

manifests/init.pp
```diff
@@ -302,7 +316,14 @@
   Optional[String] $email_smtp_domain = $::foreman::params::email_smtp_domain,
   Enum['none', 'plain', 'login', 'cram-md5'] $email_smtp_authentication = $::foreman::params::email_smtp_authentication,
   Optional[String] $email_smtp_user_name = $::foreman::params::email_smtp_user_name,
-  Optional[String] $email_smtp_password = $::foreman::params::email_smtp_password,
+  String $email_smtp_password = $::foreman::params::email_smtp_password,
```
ekohl: I think you didn't intend to change this line.

ares: no, certainly not, sorry :-)

mmoll:
Doesn't this mean per default telemetry data will get send to localhost, no matter if anything is running there?

mmoll:
ah, nope... it's only the configuration, it's still disabled :+1:

mmoll:
meregd, díky @ares!

PR #645, opened 2018-06-19
Adding spacewalk plugin

cocker-cc:
Adding spacewalk plugin.  RPM available [here](https://yum.theforeman.org/plugins/nightly/el7/x86_64/)

### DIFF
--- a/manifests/plugin/spacewalk.pp
+++ b/manifests/plugin/spacewalk.pp
@@ -1,0 +1,5 @@
+# Installs foreman_spacewalk plugin
+class foreman::plugin::spacewalk {
+  foreman::plugin {'spacewalk':
+  }
+}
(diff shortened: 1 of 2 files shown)


manifests/plugin/foreman_wreckingball.pp
```diff
@@ -0,0 +1,5 @@
+# Installs foreman_wreckingball plugin
```
ekohl: Looks like you included https://github.com/theforeman/puppet-foreman/pull/647 here. For the changelog it'd be nice to separate the two.

cocker-cc:
I cleaned it up.

cocker-cc:
I add a trivial Test, and renamed the File.

ekohl:
Thanks!

PR #646, opened 2018-06-19
Adding dlm plugin

cocker-cc:
Adding dlm plugin.  RPM available [here](https://yum.theforeman.org/plugins/nightly/el7/x86_64/)


ekohl:
Thanks!

PR #647, opened 2018-06-19
Adding wreckingball Plugin

cocker-cc:
Adding wreckingball plugin.  RPM available [here](https://yum.theforeman.org/plugins/nightly/el7/x86_64/)


manifests/plugin/foreman_wreckingball.pp
```diff
@@ -0,0 +1,5 @@
+# Installs foreman_wreckingball plugin
```
ekohl: Looks like the file should be named `wreckingball.pp` instead of `foreman_wreckingball.pp`

cocker-cc:
I add a trivial Test, and renamend the File.

ekohl:
Thanks!

PR #650, opened 2018-07-03
Fixes #22940 - Ensure the PG root cert is installed

ekohl:
When a remote database is used the postgres server is not managed. This
meant that the certificate was not placed. It's moved to the config
since it's a client config that always needs to happen.


mmoll:
merged, bedankt @ekohl!

PR #660, opened 2018-08-13
Refactor repo handling

ekohl:
* Drop use of the 'stable' repository
* Disallow releases/ as part of the repository
* Add foreman-rails repository
* Stop including foreman::params
* Remove anchor pattern
* Move to https

Will need https://github.com/theforeman/puppet-foreman/pull/659 for the acceptance tests to pass.

Note that the foreman-rails repository assumes Foreman 1.17+ so that'll become the new minimum version. I'm going to continue with removing some compatibility with older versions in follow up PRs.

### DIFF
--- a/manifests/params.pp
+++ b/manifests/params.pp
@@ -47,8 +47,8 @@ class foreman::params {
   # if set to true, no repo will be added by this module, letting you
   # set it to some custom location.
   $custom_repo       = false
-  # this can be stable, or nightly
-  $repo              = 'stable'
+  # this can be a version or nightly
+  $repo              = '1.18'
   $app_root          = '/usr/share/foreman'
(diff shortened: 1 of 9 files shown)


ekohl:
I'll need to split out a few more things, work on tests and reorder a bit. Then I'll need to verify if there are more version specific workarounds we can get rid of. Also thinking about removing the explicit failures about an unsupported OS and setting values to undef so people can override where needed.

mmoll:
tests fail.

ekohl:
Updated to fix a bug in the yum handling. Debian nightlies are still broken so can't promise we don't have regressions there.

mmoll:
merged, bedankt @ekohl!

PR #661, opened 2018-08-15
Set release compatibility to 1.17+

ekohl:


ekohl:
Split off from https://github.com/theforeman/puppet-foreman/pull/660 so we get a nice changelog.

ehelms:
What is the issue that leads to dropping dynflow_in_core for my understanding?

ekohl:
It was a compatibility setting we had so you could use the module with 1.16 if you set it to false. Then it would revert back to the `foreman-tasks` service. This PR is raising the minimum version to 1.17 so we can assume it's `dynflowd`. In a multi-host environment you can still set the ensure to stopped to only run it on a single server.

mmoll:
merged, bedankt @ekohl!

PR #662, opened 2018-08-15
Refactor Puppet handling

ekohl:
This moves all puppetmaster defaults to its own parameters file.
Parameters that were used in both are duplicated. It also adds Puppet
types to all parameters.

This removes puppet settings from the top level. The puppet_ssldir is
now an internal detail to determine defaults. It is also no longer set
in Foreman since it's only used as a base for determining SSL paths
which we explicitly set. There is a PR open to remove this setting
altogether.

### DIFF
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -124,10 +124,6 @@
 #
 # $user_groups::                  Additional groups for the Foreman user
 #
-# $puppet_home::                  Puppet home directory
-#
-# $puppet_ssldir::                Puppet SSL directory
-#
 # $passenger_interface::          Defines which network interface passenger should listen on, undef means all interfaces
 #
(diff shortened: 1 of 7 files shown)


mmoll:
lint fails

ekohl:
Looks like I accidentally removed a bunch of code that would have broken otherwise. That was not intentional.

ekohl:
I'm not sure. While that variable is used by Foreman to determine defaults for other values, we don't really use the value because ssl certs are explicitly configured. That's why I'm slightly in doubt about it.

ekohl:
@mmoll do you want to wait for the core merge or can we merge this now? IMHO both are ready to be merged now.

mmoll:
merged, bedankt @ekohl!

PR #663, opened 2018-08-15
Clean up providers

ekohl:
* Removes JSON installation which is no longer needed
* Removes apipie-based providers
* Removes the foreman_hostgroup type since there's no more provider


mmoll:
tests fail

mmoll:
merged, bedankt @ekohl!

mdwheele:
Okay, I see. The `foreman_hostgroup` provider was using the apipie bindings and since those were removed, the `rest_v2` provider was removed. At any rate, we'd be happy to send what we have for:

- `foreman_auth`
- `foreman_domain`
- `foreman_filter`
- `foreman_media`
- `foreman_operatingsystem`
- `foreman_role`
- `foreman_subnet`
- `foreman_usergroup`

We're adding provisioning templates and partition tables. Also, we'll need some help with tests when the time comes :grin:

PR #664, opened 2018-08-15
Remove remote_file

ekohl:
This was used by the discovery image but that has all moved to the
proxy. See foreman_proxy::remote_file for a replacement.


mmoll:
merged, bedankt @ekohl!

PR #667, opened 2018-08-24
allow puppetlabs-stdlib 5.x

mmoll:


mateusz-gozdek-sociomantic:
puppetlabs-postgresql 5.9.0 is now released https://github.com/puppetlabs/puppetlabs-postgresql/releases/tag/5.9.0, to tests should pass, but they need to be re-triggered :crossed_fingers:

ekohl:
The acceptance tests are already broken because of beaker-puppet so I'm going to merge this. A proper dependency resolver (like librarian-puppet) should be able to deal with it anyway.

PR #670, opened 2018-09-13
add --push-only-current-client-facts option 

xorpaul:
We want to push Puppet Facter facts to our Foreman synchronously, but have no need for using Foreman as an ENC.

This PR adds `--push-only-current-client-facts` as a `node.rb` argument, which then pushes the Facter facts of the currently Puppet client to Foreman.

Our `puppet.conf` then looks like this:
```
external_nodes = /etc/puppetlabs/puppet/node.rb --push-only-current-client-facts
node_terminus  = exec  
```
The Puppetserver then triggers 
```
$ /etc/puppetlabs/puppet/node.rb --push-only-current-client-facts foobar.domain.tld      
---
classes: {}
```

and sends the up-to-date Fact facts of node `foobar.domain.tld` to Foreman.

### DIFF
--- a/files/external_node_v2.rb
+++ b/files/external_node_v2.rb
@@ -7,6 +7,10 @@
 # If --push-facts is given as the only arg, it uploads facts for all hosts and then exits.
 # Useful in scenarios where the ENC isn't used.
 
+# If --push-only-current-client-facts is given as the only arg, it uploads only the facts
+# of the current Puppet client that requests its Puppet catalog. Returns a YAML
+# with an empty `classes` hash.
+
 require 'rbconfig'
 require 'yaml'
 
@@ -356,6 +360,7 @@ if __FILE__ == $0 then
     watch = ARGV.delete("--watch-facts")
     push_facts_parallel = ARGV.delete("--push-facts-parallel")
     push_facts = ARGV.delete("--push-facts")
+    push_only_current_client_facts = ARGV.delete("--push-only-current-client-facts")
     if watch && ! ( push_facts || push_facts_parallel )
         raise "Cannot watch for facts without specifying --push-facts or --push-facts-parallel"
     end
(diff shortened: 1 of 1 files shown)


files/external_node_v2.rb
```diff
@@ -377,6 +382,12 @@ def watch_and_send_facts(parallel)
             upload_facts(certname, req)
           end
 
+          if push_only_current_client_facts
+            require 'yaml'
+            empty_env_result = { "classes" => {}}
+            puts empty_env_result.to_yaml
```
ekohl: This isn't quite correct. If the fact upload goes wrong for whatever reason, you'll be served the cache if available. That means switching from ENC to non-ENC could give some weird results. I think you should do:
```ruby
if push_only_current_client_facts
  require 'yaml'
  empty_env_result = { "classes" => {}}
  result = empty_env_result.to_yaml
 else
  result = enc(certname)
  cache(certname, result)
end
```

xorpaul: If I understand you correctly, if the fact upload goes wrong, we should end up in the `rescue` part (https://github.com/theforeman/puppet-foreman/pull/670/files#diff-56980159eb7873d008e28343dd19c1f3L383), which means that the `exit 0` gets skipped.

So I would need to fix the `rescue` part by adding the empty yaml puts _and_ `exit 0` to it if `push_only_current_client_facts` is set.

With your suggested snippet the `result = read_cache(certname)` and the puts of it will still happen, because the `exit 0` was skipped.

ekohl: You're right. I did mean that's the problem but my solution is incorrect. I would prefer to setting `result = ...` rather than exiting early. Looking at the code I think it could use some more helper functions but that doesn't have to be part of this PR.

xorpaul: Ah, I understand. I've tried to avoid the premature `exit` here: https://github.com/theforeman/puppet-foreman/pull/670/commits/aa7638cea2dd3980be542d310c57a3dd1a2a6265

I'll rebase later and I'm also okay with renaming the param to `--no-env` although I prefer more descriptive arguments in general :smirk:

files/external_node_v2.rb
```diff
@@ -377,15 +382,27 @@ def watch_and_send_facts(parallel)
             upload_facts(certname, req)
           end
 
-          result = enc(certname)
-          cache(certname, result)
+          if ! push_only_current_client_facts
```
ekohl: I'd prefer still setting the `result` variable. Makes it easier to reason about the program. Perhaps a `blank_enc` function that returns the empty content would be useful?

mmoll:
@xorpaul what's the status here?

xorpaul:
I'm satisfied with my version of the script.

If @ekohl would like to add his improvements as he described in his latest comment, then he can close this pull request and open a new one.

mmoll:
@xorpaul well, this PR needs at least a rebase onto the current master branch

mmoll:
@ekohl WDYT? :point_up:

ekohl:
We already have an implementation that watches the facts directory and uploads them. Then it runs more as a daemon. Wouldn't this be a better solution since then you don't block your puppet runs on the Foreman ENC processing all your facts.

ekohl:
Closing for inactivity and my last comment.

PR #672, opened 2018-10-08
Refactor extras repo handling

ekohl:
This drops support for EL6 by only having the URL for the EL6 GPG key.

It also allows setting the package state of foreman-release-scl.

### DIFF
--- a/spec/classes/foreman_repos_extra_spec.rb
+++ b/spec/classes/foreman_repos_extra_spec.rb
@@ -1,53 +1,49 @@
 require 'spec_helper'
 
 describe 'foreman::repos::extra' do
-
   on_os_under_test.each do |os, facts|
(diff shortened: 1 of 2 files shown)


spec/classes/foreman_repos_extra_spec.rb
```diff
@@ -1,53 +1,49 @@
 require 'spec_helper'
 
 describe 'foreman::repos::extra' do
-
   on_os_under_test.each do |os, facts|
-    context "on #{os}" do
+    context "on #{os}", if: facts[:osfamily] == 'RedHat' do
```
ekohl: I thought this was a nice optimization but I'd be happy to change this.

mmoll:
merged, bedankt @ekohl!

PR #674, opened 2018-10-10
Fixes #25170 - Update admin/pass param to initial_param.

### LINKED ISSUE #406: Wrong admin username and password for sqlite database
If sqlite database is in use, `admin_username` and `admin_password` are not used.

chris1984:
Fixes #406

### DIFF
--- a/spec/classes/foreman_database_spec.rb
+++ b/spec/classes/foreman_database_spec.rb
@@ -21,7 +21,8 @@ describe 'foreman' do
       describe 'with db_manage_rake set to false' do
         let(:params) { super().merge(db_manage_rake: false) }
 
-        it { should contain_class('foreman::database::postgresql').with_notify(nil) }
+        it { should compile.with_all_deps }
+        it { should contain_class('foreman::database::postgresql') }
 
         it { should_not contain_foreman_config_entry('db_pending_migration') }
         it { should_not contain_foreman__rake('db:migrate') }
--- a/spec/classes/foreman_spec.rb
+++ b/spec/classes/foreman_spec.rb
@@ -142,7 +142,7 @@ describe 'foreman' do
         it { should_not contain_class('foreman::database::mysql') }
         it {
           should contain_class('foreman::database::postgresql')
-            .with_notify('Foreman_config_entry[db_pending_migration]')
+            .that_notifies('Foreman_config_entry[db_pending_migration]')
         }
 
         it { should contain_foreman_config_entry('db_pending_migration') }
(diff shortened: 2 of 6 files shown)


ekohl:
Perhaps we should open a redmine issue for this as well for easier tracking in the foreman release log. This will impact installation manuals for example.

ekohl:
@ares IMHO the parameter name should be self explanatory. With the amount of parameters we have we simply can't expect people to read them all. We already do so for the organization and location which is also part of the seeding defaults. This makes it consistent. In RH we also had downstream complaints about it besides the already linked issue.

ares:
Ok, no problems with that if we add migration and improve the parameter description. It's just that everyone who has the.installation scripted will need to update the script as the initial password is likely being set.

chris1984:
Related installer prs:

Katello:
https://github.com/Katello/katello-installer/pull/699

Foreman:
https://github.com/theforeman/foreman-installer/pull/296

spec/classes/foreman_spec.rb
```diff
@@ -43,92 +43,92 @@ class { 'foreman::providers':
...
-          :logging_level             => 'info',
-          :loggers                   => {},
-          :email_delivery_method     => 'sendmail',
-          :email_smtp_address        => 'smtp.example.com',
-          :email_smtp_port           => 25,
-          :email_smtp_domain         => 'example.com',
-          :email_smtp_authentication => 'none',
-          :email_smtp_user_name      => 'root',
-          :email_smtp_password       => 'secret',
+          :foreman_url                       => 'http://localhost',
```
ekohl: This whitespace re-alignment isn't needed

spec/classes/foreman_database_spec.rb
```diff
@@ -63,8 +63,8 @@
       describe 'with seed parameters' do
```
ekohl: Maybe this should be exanded to include all `initial_*` parameters to verify they're set correctly.

ekohl:
This will also need a change in forklift. Grep for foreman-admin-password and anything that applies nightly needs an update. It is the last change I want to get in before 10.0 and before branching. Usually I prefer to avoid such a breaking change so late, but I really want to avoid a major version bump for a while.

cc @tbrisker

tbrisker:
@ekohl how close is this to being merged? and does this mean we're blocked on branching the installer?  How important is getting this change in for 1.20?
seeing that this has impacts on installer, forklift and possibly breaking user's scripting, i'm inclined to suggesting we push this off to the next version, but i'm willing to be persuaded otherwise.

chris1984:
@tbrisker I can get all the prs updated tonight, first thing tomorrow. It would be nice to get this into 6.5

ekohl:
@chris1984 in the release meeting we decided it was too late in the branching cycle to pull this in. I'm going to do a 10.0.0 release of this module and soon after we can merge this starting the 11.0.0 cycle. For downstream we can find a solution, but don't need to discuss this in upstream.

ekohl:
@chris1984 could you rebase this? Now that the installer merge is complete, I'd like to get this in.

PR #675, opened 2018-10-11
namespace extlib functions

mmoll:


ekohl:
I've been holding off on this to avoid increasing the lowest extlib version so users could upgrade, but maybe the 10.0 release is a good moment.

ekohl:
Thanks!

PR #677, opened 2018-10-15
Rewrite to support reverse proxy

ekohl:
This doesn't implement the actual reverse proxy part, but it does make passenger optional and prepares the code for it.

### DIFF
--- a/README.md
+++ b/README.md
@@ -52,7 +52,7 @@ previous stable release.
 ### Foreman version compatibility notes
 
 This module targets Foreman 1.21+. Running without passenger is only supported
-on Foreman 1.22+.
+on Foreman 1.23+.
 
 The Foreman userdata plugin has been merged into Foreman 1.23 and removed from
 this module.
@@ -64,7 +64,7 @@ To use this module without passenger, the `passenger` parameter must be set to
 is running.
 
 This introduces a soft dependency on `camptocamp-systemd`. This feature is only
-available on Foreman 1.22+.
+available on Foreman 1.23+.
 
 ## Types and providers
 
--- a/manifests/config/apache.pp
+++ b/manifests/config/apache.pp
@@ -7,7 +7,7 @@
 # $listen_on_interface::      Specify which interface to bind passenger to.
 #                             Defaults to all interfaces.
 #
-# $ruby::                     Path to Ruby interpreter
+# $passenger_ruby::           Path to Ruby interpreter
 #
 # $priority::                 Apache vhost priority
 #
@@ -37,11 +37,11 @@
 #
 # $user::                     The user under which the application runs.
 #
-# $prestart::                 Pre-start the first passenger worker instance process during httpd start.
+# $passenger_prestart::       Pre-start the first passenger worker instance process during httpd start.
 #
-# $min_instances::            Minimum passenger worker instances to keep when application is idle.
+# $passenger_min_instances::  Minimum passenger worker instances to keep when application is idle.
 #
-# $start_timeout::            Amount of seconds to wait for Ruby application boot.
+# $passenger_start_timeout::  Amount of seconds to wait for Ruby application boot.
 #
 # $foreman_url::              The URL Foreman should be reachable under. Used for loading the application
 #                             on startup rather than on demand.
@@ -65,14 +65,18 @@
 # $https_vhost_options::      Direct options to apache::vhost for the https vhost
 #
 class foreman::config::apache(
+  Boolean $passenger = $::foreman::passenger,
   Stdlib::Absolutepath $app_root = $::foreman::app_root,
   Optional[String] $listen_on_interface = $::foreman::passenger_interface,
-  Optional[String] $ruby = $::foreman::passenger_ruby,
+  Optional[String] $passenger_ruby = $::foreman::passenger_ruby,
   String $priority = $::foreman::vhost_priority,
   Stdlib::Fqdn $servername = $::foreman::servername,
   Array[Stdlib::Fqdn] $serveraliases = $::foreman::serveraliases,
   Stdlib::Port $server_port = $::foreman::server_port,
   Stdlib::Port $server_ssl_port = $::foreman::server_ssl_port,
+  Stdlib::Httpurl $proxy_backend = "http://${::foreman::foreman_service_bind}:${::foreman::foreman_service_port}/",
+  Hash $proxy_params = {'retry' => '0'},
+  Array[String] $proxy_no_proxy_uris = ['/pulp', '/streamer', '/pub'],
   Boolean $ssl = $::foreman::ssl,
   Stdlib::Absolutepath $ssl_ca = $::foreman::server_ssl_ca,
   Stdlib::Absolutepath $ssl_chain = $::foreman::server_ssl_chain,
@@ -82,9 +86,9 @@ class foreman::config::apache(
   Variant[Enum[''], Stdlib::Absolutepath] $ssl_crl = $::foreman::server_ssl_crl,
   Optional[String] $ssl_protocol = $::foreman::server_ssl_protocol,
   String $user = $::foreman::user,
-  Boolean $prestart = $::foreman::passenger_prestart,
-  Integer[0] $min_instances = $::foreman::passenger_min_instances,
-  Integer[0] $start_timeout = $::foreman::passenger_start_timeout,
+  Boolean $passenger_prestart = $::foreman::passenger_prestart,
+  Integer[0] $passenger_min_instances = $::foreman::passenger_min_instances,
+  Integer[0] $passenger_start_timeout = $::foreman::passenger_start_timeout,
   Stdlib::HTTPUrl $foreman_url = $::foreman::foreman_url,
   Boolean $keepalive = $::foreman::keepalive,
   Integer[0] $max_keepalive_requests = $::foreman::max_keepalive_requests,
@@ -93,6 +97,7 @@ class foreman::config::apache(
   Boolean $ipa_authentication = $::foreman::ipa_authentication,
   Hash[String, Any] $http_vhost_options = {},
   Hash[String, Any] $https_vhost_options = {},
+  Optional[Boolean] $selinux = $::foreman::selinux,
 ) {
   $docroot = "${app_root}/public"
   $suburi_parts = split($foreman_url, '/')
@@ -101,12 +106,94 @@ class foreman::config::apache(
     $suburi_without_slash = join(values_at($suburi_parts, ["3-${suburi_parts_count}"]), '/')
     if $suburi_without_slash {
       $suburi = "/${suburi_without_slash}"
+    } else {
+      $suburi = undef
+    }
+  } else {
+    $suburi = undef
+  }
+
+  if $passenger {
+    if $suburi {
+      $custom_fragment = template('foreman/_suburi.conf.erb')
+    } else {
+      $custom_fragment = template('foreman/_assets.conf.erb')
+    }
+
+    $passenger_options = {
+      'passenger_app_root' => $app_root,
+      'passenger_min_instances' => $passenger_min_instances,
+      'passenger_start_timeout' => $passenger_start_timeout,
+      'passenger_ruby' => $passenger_ruby,
+    }
+
+    if $passenger_prestart {
+      $vhost_http_internal_options = $passenger_options + {'passenger_pre_start' => "http://${servername}:${server_port}"}
+      $vhost_https_internal_options = $passenger_options + {'passenger_pre_start' => "https://${servername}:${server_ssl_port}"}
+    } else {
+      $vhost_http_internal_options = $passenger_options
+      $vhost_https_internal_options = $passenger_options
+    }
+
+    if $app_root {
+      file { ["${app_root}/config.ru", "${app_root}/config/environment.rb"]:
+        owner => $user,
+      }
+    }
+  } else {
+    if $suburi {
+      $custom_fragment = undef
+    } else {
+      $custom_fragment = template('foreman/_assets.conf.erb')
+    }
+
+    include ::apache::mod::proxy_wstunnel
+    $websockets_backend = regsubst($proxy_backend, 'http://', 'ws://')
+
+    $vhost_http_internal_options = {
+      'proxy_preserve_host' => true,
+      'proxy_add_headers'   => true,
+      'request_headers'     => [
+        'set X_FORWARDED_PROTO "http"',
+        'set SSL_CLIENT_S_DN ""',
+        'set SSL_CLIENT_CERT ""',
+        'set SSL_CLIENT_VERIFY ""',
+      ],
+      'proxy_pass'          => {
+        'no_proxy_uris' => $proxy_no_proxy_uris,
+        'path'          => pick($suburi, '/'),
+        'url'           => $proxy_backend,
+        'params'        => $proxy_params,
+      },
+      'rewrites'            => [
+        {
+          'comment'      => 'Upgrade Websocket connections',
+          'rewrite_cond' => '%{HTTP:Upgrade} =websocket [NC]',
+          'rewrite_rule' => "/(.*) ${websockets_backend}\$1 [P,L]",
+        },
+      ],
+    }
+
+    $vhost_https_internal_options = $vhost_http_internal_options + {
+      'ssl_proxyengine' => true,
+      'request_headers' => [
+        'set X_FORWARDED_PROTO "https"',
+        'set SSL_CLIENT_S_DN "%{SSL_CLIENT_S_DN}s"',
+        'set SSL_CLIENT_CERT "%{SSL_CLIENT_CERT}s"',
+        'set SSL_CLIENT_VERIFY "%{SSL_CLIENT_VERIFY}s"',
+      ],
+    }
+
+    if $selinux or ($facts['selinux'] and $selinux != false) {
+      selboolean { 'httpd_can_network_connect':
+        persistent => true,
+        value      => 'on',
+      }
     }
   }
 
   include ::apache
   include ::apache::mod::headers
-  include ::apache::mod::passenger
 
   if $ipa_authentication {
     include ::apache::mod::authnz_pam
@@ -122,11 +209,6 @@ class foreman::config::apache(
     $listen_interface = undef
   }
 
-  $http_prestart = $prestart ? {
-    true  => "http://${servername}:${server_port}",
-    false => undef,
-  }
-
   file { "${apache::confd_dir}/${priority}-foreman.d":
     ensure  => 'directory',
     owner   => 'root',
@@ -137,35 +219,26 @@ class foreman::config::apache(
   }
 
   apache::vhost { 'foreman':
-    add_default_charset     => 'UTF-8',
-    docroot                 => $docroot,
-    manage_docroot          => false,
-    ip                      => $listen_interface,
-    options                 => ['SymLinksIfOwnerMatch'],
-    passenger_app_root      => $app_root,
-    passenger_min_instances => $min_instances,
-    passenger_pre_start     => $http_prestart,
-    passenger_start_timeout => $start_timeout,
-    passenger_ruby          => $ruby,
-    port                    => $server_port,
-    priority                => $priority,
-    servername              => $servername,
-    serveraliases           => $serveraliases,
-    keepalive               => bool2str($keepalive, 'on', 'off'),
-    max_keepalive_requests  => $max_keepalive_requests,
-    keepalive_timeout       => $keepalive_timeout,
-    access_log_format       => $access_log_format,
-    additional_includes     => ["${::apache::confd_dir}/${priority}-foreman.d/*.conf"],
-    use_optional_includes   => true,
-    custom_fragment         => template('foreman/_assets.conf.erb', 'foreman/_suburi.conf.erb'),
-    *                       => $http_vhost_options,
+    add_default_charset    => 'UTF-8',
+    docroot                => $docroot,
+    manage_docroot         => false,
+    ip                     => $listen_interface,
+    options                => ['SymLinksIfOwnerMatch'],
+    port                   => $server_port,
+    priority               => $priority,
+    servername             => $servername,
+    serveraliases          => $serveraliases,
+    keepalive              => bool2str($keepalive, 'on', 'off'),
+    max_keepalive_requests => $max_keepalive_requests,
+    keepalive_timeout      => $keepalive_timeout,
+    access_log_format      => $access_log_format,
+    additional_includes    => ["${::apache::confd_dir}/${priority}-foreman.d/*.conf"],
+    use_optional_includes  => true,
+    custom_fragment        => $custom_fragment,
+    *                      => $vhost_http_internal_options + $http_vhost_options,
   }
 
   if $ssl {
-    $https_prestart = $prestart ? {
-      true  => "https://${servername}:${server_ssl_port}",
-      false => undef,
-    }
     if $ssl_crl and $ssl_crl != '' {
       $ssl_crl_real = $ssl_crl
       $ssl_crl_check = 'chain'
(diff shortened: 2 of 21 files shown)


manifests/init.pp
```diff
@@ -53,7 +53,9 @@
 # $authentication::               Enable user authentication. Initial credentials are set using admin_username
 #                                 and admin_password.
 #
-# $passenger::                    Configure foreman via apache and passenger
+# $apache::                       Configure Foreman via Apache. By default via passenger but otherwise as a reverse proxy.
+#
+# $passenger::                    Whether to configure Apache with passenger or as a reverse proxy.
```
ehelms: Should these two conflict? or is the idea you need both now to get passenger?

ehelms:
Just to show you how easy "enabling" puma is (And confirm my theory works about installing the RPM):

```
[root@centos7-katello-nightly vagrant]# journalctl -xn
-- Logs begin at Mon 2018-10-15 12:34:30 UTC, end at Mon 2018-10-15 12:54:54 UTC. --
Oct 15 12:54:38 centos7-katello-nightly.war.example.com scl[15030]: /opt/theforeman/tfm/root/usr/share/gems/gems/katello-3.9.0/app/models/katello/concerns/subscription_facet_host_extensions.rb:36: warning: previ
Oct 15 12:54:38 centos7-katello-nightly.war.example.com scl[15030]: /opt/theforeman/tfm/root/usr/share/gems/gems/katello-3.9.0/app/models/katello/concerns/subscription_facet_host_extensions.rb:41: warning: alrea
Oct 15 12:54:38 centos7-katello-nightly.war.example.com scl[15030]: /opt/theforeman/tfm/root/usr/share/gems/gems/katello-3.9.0/app/models/katello/concerns/subscription_facet_host_extensions.rb:41: warning: previ
Oct 15 12:54:40 centos7-katello-nightly.war.example.com scl[15030]: /opt/theforeman/tfm/root/usr/share/gems/gems/foreman_docker-4.1.0/app/controllers/api/v2/containers_controller.rb:107: warning: constant ::Fixn
Oct 15 12:54:41 centos7-katello-nightly.war.example.com scl[15030]: => Booting WEBrick
Oct 15 12:54:41 centos7-katello-nightly.war.example.com scl[15030]: => Rails 5.2.1 application starting in production on http://0.0.0.0:3000
Oct 15 12:54:41 centos7-katello-nightly.war.example.com scl[15030]: => Run `rails server -h` for more startup options
Oct 15 12:54:54 centos7-katello-nightly.war.example.com scl[15030]: [2018-10-15 12:54:54] INFO  WEBrick 1.4.2
Oct 15 12:54:54 centos7-katello-nightly.war.example.com scl[15030]: [2018-10-15 12:54:54] INFO  ruby 2.5.0 (2017-12-25) [x86_64-linux]
Oct 15 12:54:54 centos7-katello-nightly.war.example.com scl[15030]: [2018-10-15 12:54:54] INFO  WEBrick::HTTPServer#start: pid=15062 port=3000
[root@centos7-katello-nightly vagrant]# yum -y install tfm-rubygem-puma
Loaded plugins: fastestmirror
Loading mirror speeds from cached hostfile
 * base: packages.oit.ncsu.edu
 * epel: download-ib01.fedoraproject.org
 * extras: packages.oit.ncsu.edu
 * updates: packages.oit.ncsu.edu
Resolving Dependencies
--> Running transaction check
---> Package tfm-rubygem-puma.x86_64 0:3.11.4-3.el7 will be installed
--> Finished Dependency Resolution

Dependencies Resolved

===================================================================================================================================================================================================================
 Package                                                 Arch                                          Version                                                Repository                                      Size
===================================================================================================================================================================================================================
Installing:
 tfm-rubygem-puma                                        x86_64                                        3.11.4-3.el7                                           foreman                                         79 k

Transaction Summary
===================================================================================================================================================================================================================
Install  1 Package

Total download size: 79 k
Installed size: 231 k
Downloading packages:
tfm-rubygem-puma-3.11.4-3.el7.x86_64.rpm                                                                                                                                                    |  79 kB  00:00:00     
Running transaction check
Running transaction test
Transaction test succeeded
Running transaction
  Installing : tfm-rubygem-puma-3.11.4-3.el7.x86_64                                                                                                                                                            1/1 
  Verifying  : tfm-rubygem-puma-3.11.4-3.el7.x86_64                                                                                                                                                            1/1 

Installed:
  tfm-rubygem-puma.x86_64 0:3.11.4-3.el7                                                                                                                                                                           

Complete!
[root@centos7-katello-nightly vagrant]# service foreman restart
Redirecting to /bin/systemctl restart foreman.service
[root@centos7-katello-nightly vagrant]# journalctl -xn
-- Logs begin at Mon 2018-10-15 12:34:30 UTC, end at Mon 2018-10-15 12:55:59 UTC. --
Oct 15 12:55:56 centos7-katello-nightly.war.example.com scl[15030]: /opt/rh/rh-ruby25/root/usr/share/rubygems/rubygems/core_ext/kernel_require.rb:39:in `require'
Oct 15 12:55:56 centos7-katello-nightly.war.example.com scl[15030]: bin/rails:9:in `<main>'
Oct 15 12:55:56 centos7-katello-nightly.war.example.com scl[15030]: [2018-10-15 12:55:56] INFO  going to shutdown ...
Oct 15 12:55:56 centos7-katello-nightly.war.example.com scl[15030]: [2018-10-15 12:55:56] INFO  WEBrick::HTTPServer#start done.
Oct 15 12:55:56 centos7-katello-nightly.war.example.com scl[15030]: Exiting
Oct 15 12:55:56 centos7-katello-nightly.war.example.com systemd[1]: Started Foreman.
-- Subject: Unit foreman.service has finished start-up
-- Defined-By: systemd
-- Support: http://lists.freedesktop.org/mailman/listinfo/systemd-devel
-- 
-- Unit foreman.service has finished starting up.
-- 
-- The start-up result is done.
Oct 15 12:55:56 centos7-katello-nightly.war.example.com systemd[1]: Starting Foreman...
-- Subject: Unit foreman.service has begun start-up
-- Defined-By: systemd
-- Support: http://lists.freedesktop.org/mailman/listinfo/systemd-devel
-- 
-- Unit foreman.service has begun starting up.
Oct 15 12:55:56 centos7-katello-nightly.war.example.com polkitd[319]: Unregistered Authentication Agent for unix-process:15274:128656 (system bus name :1.67, object path /org/freedesktop/PolicyKit1/Authenticatio
Oct 15 12:55:59 centos7-katello-nightly.war.example.com scl[15295]: /usr/share/foreman/lib/foreman.rb:8: warning: already initialized constant Foreman::UUID_REGEXP
Oct 15 12:55:59 centos7-katello-nightly.war.example.com scl[15295]: /usr/share/foreman/lib/foreman.rb:8: warning: previous definition of UUID_REGEXP was here
[root@centos7-katello-nightly vagrant]# journalctl -xn
-- Logs begin at Mon 2018-10-15 12:34:30 UTC, end at Mon 2018-10-15 12:56:03 UTC. --
Oct 15 12:56:01 centos7-katello-nightly.war.example.com scl[15295]: /opt/theforeman/tfm/root/usr/share/gems/gems/katello-3.9.0/app/models/katello/concerns/subscription_facet_host_extensions.rb:31: warning: alrea
Oct 15 12:56:01 centos7-katello-nightly.war.example.com scl[15295]: /opt/theforeman/tfm/root/usr/share/gems/gems/katello-3.9.0/app/models/katello/concerns/subscription_facet_host_extensions.rb:31: warning: previ
Oct 15 12:56:01 centos7-katello-nightly.war.example.com scl[15295]: /opt/theforeman/tfm/root/usr/share/gems/gems/katello-3.9.0/app/models/katello/concerns/subscription_facet_host_extensions.rb:36: warning: alrea
Oct 15 12:56:01 centos7-katello-nightly.war.example.com scl[15295]: /opt/theforeman/tfm/root/usr/share/gems/gems/katello-3.9.0/app/models/katello/concerns/subscription_facet_host_extensions.rb:36: warning: previ
Oct 15 12:56:01 centos7-katello-nightly.war.example.com scl[15295]: /opt/theforeman/tfm/root/usr/share/gems/gems/katello-3.9.0/app/models/katello/concerns/subscription_facet_host_extensions.rb:41: warning: alrea
Oct 15 12:56:01 centos7-katello-nightly.war.example.com scl[15295]: /opt/theforeman/tfm/root/usr/share/gems/gems/katello-3.9.0/app/models/katello/concerns/subscription_facet_host_extensions.rb:41: warning: previ
Oct 15 12:56:03 centos7-katello-nightly.war.example.com scl[15295]: /opt/theforeman/tfm/root/usr/share/gems/gems/foreman_docker-4.1.0/app/controllers/api/v2/containers_controller.rb:107: warning: constant ::Fixn
Oct 15 12:56:03 centos7-katello-nightly.war.example.com scl[15295]: => Booting Puma
Oct 15 12:56:03 centos7-katello-nightly.war.example.com scl[15295]: => Rails 5.2.1 application starting in production
Oct 15 12:56:03 centos7-katello-nightly.war.example.com scl[15295]: => Run `rails server -h` for more startup options
```

ehelms:
Not finding a traceback, but I did get an error trying to switch from passenger to service:

```
[ERROR 2018-10-15T12:54:31 main]  /Stage[main]/Foreman::Service/Service[foreman]/ensure: change from 'stopped' to 'running' failed: Could not enable foreman:
```

manifests/init.pp
ekohl: This changes it so that the parameter is ignored if `apache` is `false`

ekohl:
I got that too in the acceptance tests and wondered if it was something about the docker container.

ekohl:
> Googling around seems to have something to do with if the symlink already exists. Perhaps something else is creating it before sytsemctl?

Perhaps we do it wrong in our systemd unit setup in packaging?

manifests/config/apache.pp
```diff
@@ -0,0 +1,273 @@
...
+      ssl_verify_depth       => '3',
+      keepalive              => $keepalive_onoff,
+      max_keepalive_requests => $max_keepalive_requests,
+      keepalive_timeout      => $keepalive_timeout,
+      access_log_format      => $access_log_format,
+      additional_includes    => ["${::apache::confd_dir}/${priority}-foreman-ssl.d/*.conf"],
+      use_optional_includes  => true,
+      custom_fragment        => $custom_fragment,
+      *                      => $passenger_options + $proxy_https_options,
+      passenger_pre_start    => $passenger_https_prestart,
```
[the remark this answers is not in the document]
ekohl: It shouldn't. If passenger is false it should be undef. The reason I don't set it in `passenger_options` is that the pre-start differs for HTTP and HTTPS and didn't want to duplicate more code.

ehelms:
Just keeping track of the issues I run into while testing. Trying to switch from passenger to non-passenger, the Apache proxy seems to be failing:

```
[root@centos7-foreman-nightly vagrant]# curl https://`hostname` -k
<!DOCTYPE HTML PUBLIC "-//IETF//DTD HTML 2.0//EN">
<html><head>
<title>503 Service Unavailable</title>
</head><body>
<h1>Service Unavailable</h1>
<p>The server is temporarily unable to service your
request due to maintenance downtime or capacity
problems. Please try again later.</p>
</body></html>
```

ekohl:
What needs to be done is to unset the request headers so they can't be spoofed. Then the application needs to be configured to use HTTP_SSL_* instead of SSL_.

```
RequestHeader set SSL_CLIENT_S_DN    ""
RequestHeader set SSL_CLIENT_CERT ""
RequestHeader set SSL_CLIENT_VERIFY  ""
<Location />
  RequestHeader set SSL_CLIENT_S_DN "%{SSL_CLIENT_S_DN}s"
  # Broken - needs encoding/decoding of newlines
  #RequestHeader set SSL_CLIENT_CERT "%{SSL_CLIENT_CERT}s"
  RequestHeader set SSL_CLIENT_VERIFY "%{SSL_CLIENT_VERIFY}s"
</Location>
```

ekohl:
Did a trivial rebase. Not sure when I'll get back to it, but setting the right headers and configuring Foreman shouldn't be hard to add and then it'll work mostly of out the box.

ehelms:
Adding some additional information here based on conversation. Katello takes in certificates from Apache and processes them here:

https://github.com/Katello/katello/blob/master/app/services/cert/rhsm_client.rb

That code is called from the following and derived from this:

https://github.com/Katello/katello/blob/abbff485b07e94f8dd1dc1caeff4fc23f4e1b356/app/services/katello/authentication/client_authentication.rb#L32

ehelms:
Two additional thoughts from testing:

```
RequestHeader set SSL_CLIENT_S_DN    ""
RequestHeader set SSL_CLIENT_CERT ""
RequestHeader set SSL_CLIENT_VERIFY  "
```

I believe this would break one expectation Katello has which is a certificate passing through a reverse proxy on a smart proxy and then to the main server. See the comment from here where we do part of this today -- https://github.com/theforeman/puppet-katello/blob/76496ef444c1b6051b1e41a2596428fc32103f1f/files/katello-apache-ssl.conf#L2-L3

The second part is around how these are controlled. The code seems to make use of a Foreman configuration setting to determine the header which seems odd to me that an admin in the UI can control the infrastructure configuration of the setup, i.e.

```
app/models/setting/auth.rb:      self.set('ssl_client_cert_env', N_("Environment variable containing a client's SSL certificate"), 'SSL_CLIENT_CERT', N_('SSL client cert env')),
``

I'll do some digging to see if I can understand why this would ever need to be set this way vs. hard-coded or in `settings.yaml`.

**Comment by ekohl** [comment]
This is a rather thorough rebase where I've tried to split things up into smaller pieces. Still not entirely content but I want to start testing this now and see if it works.

File: templates/settings.yaml.erb [hunk 3]
```diff
@@ -70,3 +70,10 @@
 
 :dynflow:
   :pool_size: <%= scope.lookupvar("foreman::dynflow_pool_size") %>
+<% if scope.lookupvar("foreman::apache") && !scope.lookupvar("foreman::passenger") -%>
+
+# Configure reverse proxy headers
+:ssl_client_dn_env: HTTP_SSL_CLIENT_S_DN
+:ssl_client_verify_env: HTTP_SSL_CLIENT_VERIFY
+:ssl_client_cert_env: HTTP_SSL_CLIENT_CERT
+<% end -%>
```
ehelms: Based on our discussions, I thought you were gonna keep this still being set in the Settings (database) vs moving to SETTINGS (app restart). Don't get me wrong I like this change given what these represent, but I'd do this stand alone so we can make the code change for moving them alongside it easier.

ekohl: At this point I first want to get this working, then I'll see whether we use the config file or database. First I want to get the Apache config right.

ekohl:
In its basic form I think this works now.

templates/settings.yaml.erb
ehelms: Right... my point is the application won't respect these because they are SETTINGS and it currently is built to read Settings (database).

I was wondering if for now we'd be best doing:

```
    foreman_config_entry { 'ssl_client_cert_env':
      value => 'HTTP_SSL_CLIENT_CERt',
    }
```

ekohl: Settings that are in settings.yaml are visible, but can't be overridden:

screen shot 2018-11-14 at 19 33 11

Note it's visible, but there's no edit button. The's also a tooltip:
> This setting is defined in the configuration file 'settings.yaml' and is read-only.

ehelms: Huh, never noticed that before.

manifests/config/apache.pp
```diff
@@ -0,0 +1,284 @@
...
+    }
+
+    $proxy_https_options = {
+      'ssl_proxyengine' => true,
+      'request_headers' => [
+        'set X_FORWARDED_PROTO "https"',
+        'set SSL_CLIENT_S_DN "%{SSL_CLIENT_S_DN}s"',
+        # Broken - needs https://github.com/theforeman/foreman/pull/6239
+        'set SSL_CLIENT_CERT "%{SSL_CLIENT_CERT}s"',
+        'set SSL_CLIENT_VERIFY "%{SSL_CLIENT_VERIFY}s"',
```
ehelms: I thought you wanted to unset these and then set them for a given location?

ehelms:
From my testing with Katello, this PR works and all bats tested passed.

manifests/config/apache.pp
ekohl: I need to verify that

manifests/config/apache.pp
```diff
@@ -0,0 +1,284 @@
...
+    $backend = 'http://localhost:3000/'
+
+    $proxy_http_options = {
+      'proxy_preserve_host' => true,
+      'proxy_add_headers'   => true,
+      'request_headers' => [
+        'set X_FORWARDED_PROTO "http"',
+      ],
+      'proxy_pass'          => {
+        'no_proxy_uris' => ['/pulp', '/streamer', '/pub'],
```
ekohl: I want to verify these are actually needed

ehelms: I can't see why they wouldn't be needed since Foreman owns the root URL.

ehelms:
What are you thinking needs to be done to get this out of WIP to a full PR?

ehelms:
Looks like this needs a rebase

ekohl:
Rebased to solve conflicts. To get this merged we need support to tune the puma process and let it listen on localhost or a socket for security reasons. This will be a packaging effort. Ideally it'd unify Debian and RPM distros which until now had slightly different init scripts.

manifests/install.pp
```diff
@@ -29,11 +29,22 @@
...
+    if $::foreman::passenger {
+      if $::foreman::passenger_ruby_package {
+        package { $::foreman::passenger_ruby_package:
+          ensure  => installed,
+          require => Class['apache'],
+          before  => Class['apache::service'],
+        }
+      }
+    } else {
+      # TODO: foreman-puma subpackage?
```
ekohl: https://github.com/theforeman/foreman-packaging/pull/3315 for RPMs, need one for Debs

ehelms: DEB attempt --https://github.com/theforeman/foreman-packaging/pull/3574

ehelms:
Linting error

ekohl:
Refactored to include https://github.com/theforeman/puppet-foreman/pull/722 and https://github.com/theforeman/puppet-foreman/pull/723. I applied https://github.com/theforeman/foreman/pull/6647 as well.

The benefit is that this now only listens on 127.0.0.1:3000 which prevents external attackers spoofing headers. Local users can still imitate other servers by doing direct requests with spoofed headers, but that's already a much smaller attack vector. Using a socket is probably an even better way. We could even use systemd socket activation in the future.

ekohl:
Rebased. Now only includes https://github.com/theforeman/puppet-foreman/pull/723 as an additional commit. All other dependencies have been merged.

README.md
```diff
@@ -52,7 +52,7 @@ previous stable release.
 ### Foreman version compatibility notes
 
 This module targets Foreman 1.21+. Running without passenger is only supported
-on Foreman 1.22+.
+on Foreman 1.23+.
```
ekohl: We didn't cherry pick https://github.com/theforeman/foreman-packaging/commit/ab3a75c954f616944199a68a2a76b86acd15aa6f into 1.22. So for RPMs it's 1.22 but debs 1.23. Rather than cherry picking it for a less used function I'm choosing to just mark it as 1.23.

manifests/config/apache.pp
```diff
@@ -65,14 +65,18 @@
...
-  Optional[String] $ruby = $::foreman::passenger_ruby,
+  Optional[String] $passenger_ruby = $::foreman::passenger_ruby,
   String $priority = $::foreman::vhost_priority,
   Stdlib::Fqdn $servername = $::foreman::servername,
   Array[Stdlib::Fqdn] $serveraliases = $::foreman::serveraliases,
   Stdlib::Port $server_port = $::foreman::server_port,
   Stdlib::Port $server_ssl_port = $::foreman::server_ssl_port,
+  Stdlib::Httpurl $proxy_backend = "http://${::foreman::foreman_service_bind}:${::foreman::foreman_service_port}/",
+  Hash $proxy_params = {'retry' => '0'},
+  Array[String] $proxy_no_proxy_uris = ['/pulp', '/streamer', '/pub'],
```
ekohl: These are the ones Katello has. I'm wondering if this is needed or if there's a cleaner solution.

ekohl:
I want to get this merged soon now that we're fresh in the 1.24 development cycle.

@ehelms @timogoebel @evgeni mind having a look?

manifests/install.pp
```diff
@@ -29,11 +29,22 @@
...
+      if $::osfamily == 'RedHat' {
```
evgeni: should this also gain maj==7 → tfm, else just rubygem, while we're at it?

ekohl: I should look closer at my own PRs. We now have a `foreman-service` package which should pull in all the dependencies. Will update.

manifests/install.pp
evgeni: So while https://github.com/theforeman/foreman-packaging/pull/3643 was merged, I don't see anything that actually installs Puma in this scenario?

ekohl: https://github.com/theforeman/foreman/blob/develop/bundler.d/service.rb should do this.

manifests/config/apache.pp
```diff
@@ -101,12 +106,88 @@
...
+      'ssl_proxyengine' => true,
+      'request_headers' => [
+        'set X_FORWARDED_PROTO "https"',
+        'set SSL_CLIENT_S_DN "%{SSL_CLIENT_S_DN}s"',
+        'set SSL_CLIENT_CERT "%{SSL_CLIENT_CERT}s"',
+        'set SSL_CLIENT_VERIFY "%{SSL_CLIENT_VERIFY}s"',
+      ],
+    }
+
+    if $selinux or ($facts['selinux'] and $selinux != false) {
```
evgeni: Why this second boolean instead of just relying on `$facts['selinux']`?

ekohl: We have a parameter that allows you to turn off selinux and this copies the pattern in a slightly cleaner way:
https://github.com/theforeman/puppet-foreman/blob/458f93ad21937a7583cad57927e42917cddd8447/manifests/config.pp#L91-L96

We should clean that up as well to a more modern coding standard.

evgeni: oh, ok :)

manifests/config/apache.pp
timogoebel: `/pub` is probably required, `/pulp` is a wsgi mount, isn't it?

manifests/service.pp
```diff
@@ -16,12 +17,14 @@
...
 
-  if $passenger {
-    exec {'restart_foreman':
-      command     => "/bin/touch ${app_root}/tmp/restart.txt",
-      refreshonly => true,
-      cwd         => $app_root,
-      path        => '/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin',
+  if $apache {
+    if $passenger {
+      exec {'restart_foreman':
```
timogoebel: Where is this used?

manifests/config/apache.pp
```diff
@@ -101,12 +106,88 @@
...
+      $custom_fragment = undef
+    } else {
+      $custom_fragment = template('foreman/_assets.conf.erb')
+    }
+
+    $proxy_http_options = {
+      'proxy_preserve_host' => true,
+      'proxy_add_headers'   => true,
+      'request_headers' => [
+        'set X_FORWARDED_PROTO "http"',
```
timogoebel: can we also set `SSL_CLIENT_S_DN`, `SSL_CLIENT_CERT` and `SSL_CLIENT_VERIFY` to an empty string here? Just to make sure nobody injects something malicious here.

manifests/service.pp
ekohl: It's used implicitly. It's a `refreshonly` exec, but when `Class['foreman::service']` is refreshed, this `exec` is refreshed. Note it's been there for ages.

manifests/config/apache.pp
ekohl: Honestly, I don't know exactly. Some locations are mapped to files. It's a very complicated deployment. This was my "rather safe than sorry" solution.

ekohl:
One thing to consider is that if you can make HTTP requests to the Puma backend, you can imitate anyone. This is a security concern because most things can connect to http://localhost:$port. It could be mitigated by running Puma on a unix socket and limiting access. https://github.com/puma/puma/blob/master/docs/systemd.md#socket-activation describes how we could use this.

Is this something we want to do?

manifests/config/apache.pp
ekohl: Done

manifests/init.pp
neomilium: A future-proof approach, if we consider multiple kind of setups, could be an `Enum['apache+passenger', 'apache+puma', 'puma', 'nginx+passenger', 'nginx+puma'], what do you think about?

ekohl: While I like the idea of supporting nginx, I think it would be premature. Longer term I'd like to deprecate passenger. It'd be better to decide then how to best support it.

neomilium: I agreed about nginx, I noted it as example. The idea I want to point is to have a single class variable to select the targeted mode. Internally, we could keep `$passenger`, `$apache` and `$use_foreman_service`.

timogoebel:
> Is this something we want to do?

I would make it the default but allow to change it to some localhost port. The app (e.g. foreman) would then need to verify if the client is actually a trusted host (e.g. that's why we have https://github.com/theforeman/foreman/blob/d74dc4251f8397e31fcb9214d83f929c8992c9c3/app/controllers/concerns/foreman/controller/ip_from_request_env.rb#L12)

ekohl:
I'm looking into the socket and while I think it should be possible, I'm going to prefer to harden the setup later and initially launch it as an experimental feature.

I've adapted @neomilium's suggestion and added websockets support now.

ekohl:
Current issues:
* On CentOS 7 the statsd test is failing. This should be unmodified but does run after the service version. I suspect it may still run something in the background and need to investigate
* On Debian 9 it's failing because it can't find puma. This indicates a packaging issue.

ekohl:
> On Debian 9 it's failing because it can't find puma. This indicates a packaging issue.

I think https://github.com/theforeman/foreman-packaging/pull/4044 should fix this.

ekohl:
Still the issue:

```
  Notice: /Stage[main]/Foreman::Database/Foreman::Rake[db:migrate]/Exec[foreman-rake-db:migrate]/returns: /usr/lib/ruby/vendor_ruby/bundler/resolver.rb:366:in `block in verify_gemfile_dependencies_are_found!': Could not find gem 'puma' in any of the gem sources listed in your Gemfile or available on this machine. (Bundler::GemNotFound)
```

Not sure if my change actually made it into nightlies.

ekohl:
The issue with statsd is that the Apache config without passenger cleans `/etc/httpd/conf.d/passenger.conf` which holds `PasssengerRoot`. This file is not managed on EL7 but is relied on from the `mod_passenger` package. Removing `mod_passenger` before the test will work around this but is something users can hit if they switch back and forth.

ekohl:
I've cancelled the Debian acceptance tests because they need https://ci.theforeman.org/job/foreman-nightly-deb-pipeline/143/ to complete before they can pass. Other than that I think this PR should turn green.

ekohl:
:green_heart: now

mmoll:
merged, dank je wel @ekohl!

PR #678, opened 2018-10-17
Test with Puppet 6

ekohl:


ekohl:
Setting back to WIP. I've added a very much WIP commit to refactor the certificate handling. Not sure I'm entirely happy with it, but I'm wondering what you think. I'm thinking of providing the basics more via https://github.com/theforeman/puppet-certs and give each module with certificates their own certs class to chain to.

I'd be interested what @mmoll and @ehelms think about the general design. @stbenjam might have an idea about the realm aspect of certmonger. Ideally we could use certmonger as a nice abstraction layer.

mmoll:
@ekohl Centralizing cert handling looks very promising, as that could (in the longer run) provide an easier solution for people who want to use their own certs and also potentially unifies foreman- and katello-installer :+1:.

However, I'd like to see this green with Puppet 6 without the cert change (if that's not too tedious), so we can release a new set of modules which are Puppet 6 compatible.

ekohl:
> However, I'd like to see this green with Puppet 6 without the cert change (if that's not too tedious), so we can release a new set of modules which are Puppet 6 compatible.

You're right. This was a Friday experiment and when I get back to this I'll split it off into a separate PR with this one just using an openssl self signed certificate.

ehelms:
What is the background of the required certificate changes?

mmoll:
@ekohl what's the status here? :point_up:

mmoll:
@ekohl fixed the Debian test failures, this should be ready then?

PR #679, opened 2018-10-17
Added optional organization_fact and location_fact

eringet:
We needed to additional parameters in our foreman settings.yaml file, so I have added them as optional params to your module.  Thanks,
Erin

### DIFF
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -32,9 +32,13 @@
 #
 # $locations_enabled::            Enable locations?
 #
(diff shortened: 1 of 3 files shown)


ekohl:
I do wonder if the [foreman_config_entry](https://github.com/theforeman/puppet-foreman#types-and-providers) type would be better suited for this.

mmoll:
IMHO this is a good case for using the type to set the respective setting, closing for now.

PR #685, opened 2018-11-06
Trying to fix rescue syntax in ENC script

qingbo:
I'm not very familiar with Ruby, but I think there's some problem with the syntax.

Today I saw errors from the ENC script when the foreman server was under high load. I manually ran the script and it only gave me the following output (not sure why there weren't any stacktraces):

```
class or module required for rescue clause
```

After removing the square brackets it was able to output like this:

```
Serving cached ENC: execution expired
---
classes: {}
parameters:
```


ekohl:
The error is unrelated

PR #690, opened 2018-11-18
Fixes #25496 - Remove :login setting

tbrisker:


ekohl:
I'm marking this as a breaking change since it removes a setting. I think this also could use an installer migration to delete it from the answers file. It shouldn't have an impact if it's there, but it's nice to clean up after yourself. Per https://github.com/theforeman/puppet-foreman#support-policy I'm hesitant to merge this now. Would prefer to wait till we drop support for 1.20.

ehelms:
Should we add a version check then? if the code repository has dropped this then I wouldn't want the settings file generated for 1.21 to have a false configuration options in it.

ekohl:
We could make it an `Optional[Boolean]` and describe it in the compatibility notes. When it's `undef`, we leave it out of the config.

ehelms:
Works for me if we can't do a version check.

mmoll:
@tbrisker What's the status here?

ekohl:
> do we use the same module versions for 1.20 and 1.21?

Yes, generally we keep 2 or 3 versions supported at the same time.

Issue #691, opened 2018-11-19
External node: Allow single node push facts


ahmet2mir:
Hello,

Sometime we just need specific host fact upload on master side.

So just add certname in args for --push-facts

PR #692, opened 2018-11-19
Allow single node fact upload. Fixes GH-691

ahmet2mir:


ahmet2mir:
See #691

mmoll:
mergd thanks @ahmet2mir!

PR #697, opened 2018-11-28
Declare Foreman group explicitly

ehelms:


manifests/config.pp
```diff
@@ -66,6 +66,9 @@
       home    => $::foreman::app_root,
       gid     => $::foreman::group,
       groups  => $::foreman::user_groups,
+    } ->
+    group { $::foreman::group:
```
ekohl: The user has it as the primary group so it needs to be done before the user. IIRC the dependency also automatically declared because of it.

ehelms:
Updated

PR #699, opened 2018-12-16
Use beaker-hiera to reduce duplication

ekohl:


mmoll:
@ekohl needs a rebase

mmoll:
@ekohl now it needs fixing ;)

ekohl:
I'll look into using hiera for some variables. This also has the issue that a modulesync would overwrite it, but it's cleaner in the mean time.

ekohl:
I've gone a bit further to clean it up. It does divert from modulesync and I've used my own fork of beaker-hiera (which should actually be part of beaker-puppet, but that code was lost in the beaker 4 migration). I'm going to work in integrating the right pieces in the right projects, but it removes all workarounds from the examples and really to the point.

ekohl:
:green_heart: now.

Gemfile
```diff
@@ -29,6 +29,7 @@ gem 'puppet-blacksmith', '>= 4.1.0', {"groups"=>["development"]}
 gem 'beaker', '>= 4.2.0', {"groups"=>["system_tests"]}
 gem 'beaker-docker', {"groups"=>["system_tests"]}
 gem 'beaker-hostgenerator', '>= 1.1.10', {"groups"=>["system_tests"]}
+gem 'beaker-hiera', {"git"=>"https://github.com/ekohl/beaker-hiera", "branch"=>"fix", "groups"=>["system_tests"]}
```
ekohl: I'm not sure how to deal with this. I'll start nagging to get the upstream repo un-achived and the fix merged

spec/acceptance/hieradata/os/RedHat/7.yaml
```diff
@@ -0,0 +1,2 @@
+---
+redis::globals::scl: rh-redis5
```
ekohl: Should we also switch to PG 12 while we're at it?

spec/spec_helper_acceptance.rb
```diff
@@ -33,6 +34,61 @@
...
+          file { [$key, $certificate]:
+            owner => 'root',
+            group => 'root',
+            mode  => '0640',
+          }
+
+          class { 'foreman::repo':
+            repo                => 'nightly',
+            gpgcheck            => false,
+            configure_epel_repo => false,
```
ekohl: Since we don't actually need EPEL anymore, should we just drop the code to manage it altogether in a separate PR?

Gemfile
mmoll: I'd say this is OK, as it's "only" about testing.

spec/acceptance/hieradata/os/RedHat/7.yaml
mmoll: I'm actually a bit surprised, this still works without it ;)

spec/spec_helper_acceptance.rb
mmoll: Yes, if it's expected to stay this way... :+1:

ekohl:
Include https://github.com/theforeman/puppet-foreman/pull/827 now. Keeping it in draft due to beaker-hiera. It's also hard to combine with https://github.com/theforeman/puppet-foreman/pull/823 now.

ekohl:
https://github.com/theforeman/puppet-foreman/pull/829 takes everything except the hiera bits since I'm still unsure about the gem. It's also incompatible with modulesync.

ekohl:
There's talk of beaker moving to Voxpupuli and that would allow me to merge it and deal with it properly. I think this is a good thing in the meanwhile to merge since it cleans up things.

ekohl:
Rebased and updated to use PG 12 from SCL on EL7.

Gemfile
ekohl: Ok, let's accept this for now. There is talk about moving beaker from puppetlabs to Voxpupuli at which point I can solve it cleanly.

spec/acceptance/hieradata/os/RedHat/7.yaml
ekohl: Updated to include this now.

mmoll:
merged, bedankt @ekohl!

PR #702, opened 2018-12-19
Handle websockets_encrypt as a boolean

ekohl:
Since Foreman 97c34475c9e674db6443b28e03a5e4d94d585f2e (1.10+) this is actually a boolean. The string 'off' would be interpreted as the value true.


ekohl:
Updated

mmoll:
merged, bedankt @ekohl!

PR #708, opened 2019-01-10
Remove default repo management

ekohl:
Because we're trying to move people away from the latest repository, we set it to configure an explicit version with a default. Because this default would need to be updated all the time, it's a bad idea.

This patch removes the $custom_repo and makes $repo undef by default. This means users get no repository by default and need to explicitly choose which repository to use.


ekohl:
This mirrors https://github.com/theforeman/puppet-foreman_proxy/commit/b526c91a9c8a39abf68e56bc9d226c5d46b56bda

mmoll:
merged, bedankt @ekohl!

PR #712, opened 2019-01-18
Clarify the defaults switched

ekohl:


alexjfisher:
Needs more red flashing 10 foot letters... ;)

ekohl:
How about this?

README.md
```diff
@@ -51,11 +51,22 @@ previous stable release.
...
 will only have any affect on Foreman 1.20 or older, in newer versions these
 settings have been removed.
 
+**Warning** Users configuring Foreman 1.20 and earlier will need to pay
+particular attention. Some defaults have been flipped, including all user
+authentication.
+
+| Setting                    | module 11.x with 1.20 | module 10.x with 1.20 |
+|----------------------------|-----------------------|-----------------------|
+| `authentication` (`login`) | false                 | true                  |
```
tbrisker: in 1.20 and earlier `login` still defaults to true if missing (https://github.com/theforeman/foreman/blob/1.20-stable/config/settings.rb#L11 and https://github.com/theforeman/foreman/blob/1.19-stable/config/settings.rb#L10)

alexjfisher: But not in 1.18 which I was still using...

https://github.com/theforeman/foreman/blob/1.18-stable/config/settings.rb#L11

I was prepping for upgrading to 1.19 by first updating Puppet modules and refactoring my profiles etc.

PR #715, opened 2019-02-23
add cors domains parameter

timogoebel:
This is required to support: https://github.com/theforeman/foreman/pull/6467

### DIFF
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -211,6 +211,8 @@
 #
 # $hsts_enabled::                 Should HSTS enforcement in https requests be enabled
 #
+# $cors_domains::                 List of domains that show be allowed for Cross-Origin Resource Sharing. This requires Foreman 1.22+
+#
 class foreman (
   Stdlib::HTTPUrl $foreman_url = $::foreman::params::foreman_url,
   Boolean $puppetrun = $::foreman::params::puppetrun,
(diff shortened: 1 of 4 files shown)


manifests/init.pp
```diff
@@ -311,6 +313,7 @@
   Stdlib::Port $dynflow_pool_size = $::foreman::params::dynflow_pool_size,
   String $jobs_service = $::foreman::params::jobs_service,
   Boolean $hsts_enabled = $::foreman::params::hsts_enabled,
+  Array[Stdlib::HTTPUrl] $cors_domains = $::foreman::params::cors_domains,
```
ekohl: HTTPUrl -> Httpurl

manifests/init.pp
```diff
@@ -211,6 +211,8 @@
 #
 # $hsts_enabled::                 Should HSTS enforcement in https requests be enabled
 #
+# $cors_domains::                 List of domains that show be allowed for Cross-Origin Resource Sharing
```
ekohl: Perhaps this needs a note it's only supported with 1.22+?

templates/settings.yaml.erb
```diff
@@ -46,6 +46,14 @@
     :type: <%= scope.lookupvar("foreman::logging_type") %>
     :layout: <%= scope.lookupvar("foreman::logging_layout") %>
 
+<% if Array(scope.lookupvar('foreman::cors_domains')).size >= 1 -%>
```
ekohl: It's already guaranteed to be an array. Should it be `<% unless scope.lookupvar('foreman::cors_domains').empty? -%>`

timogoebel: Ah, true. Still living in a Puppet 3 world. I'm getting old. 👴

timogoebel:
@ekohl: Updated.

manifests/init.pp
timogoebel: Tests fail now with `unknown data type Stdlib::Httpurl`.

ekohl: I went with https://github.com/puppetlabs/puppetlabs-stdlib#stdlibhttpurl but that's not what's actually in the code :( https://github.com/puppetlabs/puppetlabs-stdlib/pull/1001 should fix that.

timogoebel:
Reverted the `HTTPUrl` change.

PR #716, opened 2019-03-07
Expose options to the http and https vhosts

ekohl:
The intention with this is allow advanced users to pass in these options via hiera. Longer term I also want to remove our use of the `custom_fragment` so that power users can use that key as well.


mmoll:
merged, bedankt @ekohl!

PR #717, opened 2019-03-12
Avoid processing fact yaml files with empty 'values' hash

antaflos:
Under certain conditions it can happen that Puppet Server creates a fact
yaml file for a node that contains an empty `values` hash, for example
when multiple Puppet Servers are accessed through a load balancer like
HAProxy where Puppet agent requests are distributed equally and randomly.

When trying to process such a facts file with an empty `values` hash
Foreman will reject the HTTP request with an error message like this:

    During the fact upload the server responded with: 422 Unprocessable Entity. Error is ignored and the execution continues.
    {"message":"ERF42-3829 [Foreman::Exception]: invalid facts, missing operating system value"}

This change adds a simple method that loads the fact yaml file and
returns true if the `values` hash is empty. Such fact yaml files are
then skipped when processing host facts and a warning is written to
STDOUT. The Puppet administrator can then examine and delete any invalid
fact yaml files.

### DIFF
--- a/files/external_node_v2.rb
+++ b/files/external_node_v2.rb
@@ -93,9 +93,21 @@ rescue LoadError
   end
 end
 
+def empty_values_hash?(facts_file)
+  facts = File.read(facts_file)
+  puppet_facts = YAML::load(facts.gsub(/\!ruby\/object.*$/,''))
+
+  puppet_facts['values'].empty?
+end
+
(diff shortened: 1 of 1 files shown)


antaflos:
The test failures seem unrelated to the changes in this PR.

mmoll:
merged, thanks @antaflos!

PR #720, opened 2019-03-18
Simplify the memcache plugin

ekohl:
Using params.pp for static values needlessly complicates the class.


mmoll:
merged, bedankt @ekohl!

PR #722, opened 2019-04-03
Refactor f::config::passenger to f::config::apache

ekohl:
Also uses some functions from stdlib to clean up code.

### DIFF
--- a/manifests/config.pp
+++ b/manifests/config.pp
@@ -79,7 +79,7 @@ class foreman::config {
   }
 
   if $::foreman::passenger  {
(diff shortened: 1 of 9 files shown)


mmoll:
@ekohl tests need updating?

ehelms:
Will this break modules we manage ?

ekohl:
This is designed as non-breaking by keeping the deprecated defines in place. We should be updating them so we can actually drop them if we prepare PRs.

mmoll:
merged bedankt @ekohl!

PR #723, opened 2019-04-03
Refactor running with a service to Foreman 1.22

ekohl:
This uses the new foreman-service package. It slightly depends on the outcome of https://github.com/theforeman/foreman-packaging/pull/3594.

It needs some parameters to select the port and bind address as well as tests.


ekohl:
Note I'm dropping the use of environment files and only use systemd drop ins to be distro agnostic, as is the current best practice. This is breaking and we may need to drop the sysvinit file in Debian packaging. @ehelms @mmoll I'd like some feedback on this.

manifests/init.pp
```diff
@@ -339,6 +339,8 @@
     timeout => 0,
   }
 
+  $use_foreman_service = ! $passenger
```
ehelms: I get why we need to set passenger to false explicitly in the current setup, but I do find not setting `use_foreman_service` unobvious when looking at the parameters someone would be specifying. If we will reverse this very soon after the 1.22 release then I can live with it.

manifests/config.pp
```diff
@@ -31,8 +31,14 @@
   }
 
   file { $::foreman::init_config:
-    ensure  => file,
-    content => template("foreman/${foreman::init_config_tmpl}.erb"),
+    ensure  => absent,
+  }
+
+  if $::foreman::use_foreman_service {
+    systemd::dropin_file { 'installer.conf':
```
ehelms: Why is it called `installer.conf` ?

manifests/init.pp
ekohl: Note this is currently not a parameter but rather a variable inside the body so the user can't override this. In the reverse proxy scenario I'm expanding this logic. Because I couldn't properly figure out an API, I opted not to expose this to the user directly.

manifests/config.pp
ekohl: I thought it'd make it easy to track where it came from. It ends up as `/etc/systemd/system/foreman.service.d/installer.conf`

manifests/config.pp
```diff
@@ -31,8 +31,14 @@
   }
 
   file { $::foreman::init_config:
```
ekohl: I was wondering if we should keep this in or just leave the file. If we modify the loading in RPM packaging to stop reading this file, we can drop this.

ekohl:
This is now :green_heart: and even though the Debian side isn't merged (https://github.com/theforeman/foreman-packaging/pull/3643), it will be needed on EL for 1.22 because the service is no longer present unless installed.

mmoll:
merged, bedankt @ekohl!

PR #728, opened 2019-04-27
allow puppetlabs-apt 7.x and puppetlabs-postgresql 7.x

mmoll:


mmoll:
should be fine now :eyes:

PR #732, opened 2019-05-08
Allow `puppetlabs/stdlib` 6.x

alexjfisher:
stdlib 6.0.0 is due to be released soon.
This module has already dropped support for puppet 4.

See
https://github.com/puppetlabs/puppetlabs-stdlib/blob/217068f97edd2396ce37ac8c1bd7e23da621d6be/CHANGELOG.md#supported-release-600

### DIFF
--- a/metadata.json
+++ b/metadata.json
@@ -30,7 +30,7 @@
     },
     {
       "name": "puppetlabs/stdlib",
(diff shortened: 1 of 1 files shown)


alexjfisher:
stdlib 6.0.0 was released earlier today.

PR #734, opened 2019-05-20
Use system packages on EL8

ekohl:
This is untested, but here so it can be tested.

cc @ehelms

### DIFF
--- a/manifests/cli/params.pp
+++ b/manifests/cli/params.pp
@@ -12,13 +12,11 @@ class foreman::cli::params {
   # OS specific paths
   case $::osfamily {
     'RedHat': {
-      case $::operatingsystem {
-        'fedora': {
-          $hammer_plugin_prefix = 'rubygem-hammer_cli_'
-        }
-        default: {
-          $hammer_plugin_prefix = 'tfm-rubygem-hammer_cli_'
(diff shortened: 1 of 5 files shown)


manifests/cli/params.pp
```diff
@@ -12,13 +12,11 @@
...
     'RedHat': {
-      case $::operatingsystem {
-        'fedora': {
-          $hammer_plugin_prefix = 'rubygem-hammer_cli_'
-        }
-        default: {
-          $hammer_plugin_prefix = 'tfm-rubygem-hammer_cli_'
-        }
+      # We use system packages except on EL7
+      if versioncmp($facts['operatingsystemmajrelease'], '8') >= 0 {
```
evgeni: When you say "except on EL7", maybe we should check for EL7?
```suggestion
      if $facts['operatingsystemmajrelease'] == '7' {
```

evgeni: (and obviously flip that if/else around)

ekohl: I preferred this version because it's a smaller diff if you ignore whitespace, but did think about your suggestion before. I was also worried about strings vs integers in an exact comparison.

PR #736, opened 2019-05-23
Puma support

ehelms:


ehelms:
This includes the commit form https://github.com/theforeman/puppet-foreman/pull/677

manifests/service.pp
```diff
@@ -18,6 +18,8 @@
   }
 
   if $apache {
+    include ::apache
```
ekohl: This feels weird to me. I'd expect this to be set in config already.

spec/acceptance/foreman_journald_spec.rb
```diff
@@ -91,10 +91,10 @@ class { '::foreman':
   end
 
   describe command("journalctl -u #{service_name}") do
-    its(:stdout) { is_expected.to match(%r{Redirected to https://#{host_inventory['fqdn']}/users/login}) }
+    its(:stdout) { is_expected.to match(%r{Started The Apache HTTP Server}) }
```
ekohl: This test was intended to test that the application logs to the journal. That means what's normally in production.log is now in the journal. By changing this to only test the regular messages, it's missing the point. Why that's not actually happening is unknown to me and may be an indication of an actual bug.

ehelms: I revereted these and they passed for me locally. I don't know what I was seeing.

manifests/service.pp
ehelms: Trying to remember why this had to be there.

ehelms:
I don't undestand these tests @ekohl -- half the time they pass, half the time they fail. Is there maybe a start up delay I need to increase?

ekohl:
Incorporated in https://github.com/theforeman/puppet-foreman/pull/677

PR #740, opened 2019-07-12
add failed_login_attempts_limit attribute

lero:


manifests/init.pp
```diff
@@ -317,6 +319,7 @@
   Boolean $jobs_service_enable = $::foreman::params::jobs_service_enable,
   Boolean $hsts_enabled = $::foreman::params::hsts_enabled,
   Array[Stdlib::HTTPUrl] $cors_domains = $::foreman::params::cors_domains,
+  Stdlib::Port $failed_login_attempts_limit = $::foreman::params::failed_login_attempts_limit,
```
ekohl: This doesn't sound like a TCP port number to me but rather an integer.
```suggestion
  Integer[0] $failed_login_attempts_limit = $::foreman::params::failed_login_attempts_limit,
```

lero: True. bad copy/paste from Stdlib::Port $dynflow_pool_size. BTW, shouldn't $dynflow_pool_size also be Integer[0]?

I'll have a look at foreman_config_entry, didn't know it. I had to configure failed_login_attempts_limit and I saw that it was missing, that's why I added :)

lero:
Yeah, foreman_config_entry is better/easier. I'll close this pull request. Thank you!

PR #742, opened 2019-07-19
Implement a foreman::enc function

ekohl:
This makes it possible to use Foreman as an ENC to your environment via Hiera lookups.

### DIFF
--- a/README.md
+++ b/README.md
@@ -96,6 +96,34 @@ Foreman's database. Providers:
 * `rest_v3` provider uses API v2 with Ruby HTTP library, OAuth and JSON (default)
 * `rest_v2` provider uses API v2 with apipie-bindings and OAuth
 
+## Foreman ENC via hiera
+
+There is a function `foreman::enc` to retrieve the ENC data. This returns the
+data as a hash and can be used in Hiera. This requires the URL to use the
+Puppet CA infrastructure:
+
+```yaml
+---
+version: 5
+hierarchy:
+  - name: "Foreman ENC"
+    data_hash: foreman::enc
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -96,6 +96,34 @@ Foreman's database. Providers:
 * `rest_v3` provider uses API v2 with Ruby HTTP library, OAuth and JSON (default)
 * `rest_v2` provider uses API v2 with apipie-bindings and OAuth
 
+## Foreman ENC via hiera
+
+There is a function `foreman_enc` to retrieve the ENC data. This returns the
```
evgeni:
```suggestion
There is a function `foreman::enc` to retrieve the ENC data. This returns the
```

lib/puppet/functions/foreman/enc.rb
```diff
@@ -0,0 +1,108 @@
...
+      require 'json'
+      content = JSON.parse(body)
+    rescue
+      content = body
+    end
+
+    begin
+      content['error']['message']
+    rescue
+      "Check Foreman's /var/log/foreman/production.log for more information."
```
evgeni: I wonder if it's a good idea to hard code the log path here, as it might differ?

README.md
ekohl: Updated. Leftover from before I namespaced it.

lib/puppet/functions/foreman/enc.rb
ekohl: Updated

mmoll:
merged, bedankt @ekohl!

PR #743, opened 2019-07-24
Allow configuring SSLVerifyClient instead of hardcode

jhooyberghs:
Allow configuration of SSLVerifyClient with all possible options according the httpd documentation. The default will remain "optional" as it was before.


ekohl:
Apologies for the slow review. This now needs a rebase. Before you do, I wonder what the intended use case is. In various cases Foreman depends on it

jhooyberghs:
I will do the rebase shortly, first the use case explanation as requested :)

We only want to allow servers or "users" with a certificate signed by the same CA to have access to the foreman GUI/API. So we use the puppet CA/certificates to add an additional layer of security to foreman. Everyone that tries to access foreman, first has to offer a valid (signed by the same CA) certificate before they can attempt a login on the webpage.

More explanation about this directive you can find in the apache doc https://httpd.apache.org/docs/current/mod/mod_ssl.html#sslverifyclient

PR #744, opened 2019-07-26
Add --no-pager to journalctl in tests

ekohl:
This test is unstable. It does work when run as a single test, but fails if multiple tests are ran. This may be due to journalctl showing a pager.

Currently this is just a theory.


ekohl:
Doesn't appear to make a difference.

PR #745, opened 2019-08-06
Drop compatibility with Foreman 1.20 and older + puppetrun parameter

ekohl:
The puppetrun feature has become much less prominent and we shouldn't set this via the YAML file. This can be changed at runtime if desired or via the foreman_config_entry type.

### DIFF
--- a/templates/settings.yaml.erb
+++ b/templates/settings.yaml.erb
@@ -7,16 +7,7 @@
 <% unless [nil, :undefined, :undef].include?(scope.lookupvar("foreman::unattended_url")) -%>
 :unattended_url: <%= scope.lookupvar("foreman::unattended_url") %>
 <% end -%>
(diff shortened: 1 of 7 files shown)


templates/settings.yaml.erb
```diff
@@ -7,17 +7,7 @@
 <% unless [nil, :undefined, :undef].include?(scope.lookupvar("foreman::unattended_url")) -%>
 :unattended_url: <%= scope.lookupvar("foreman::unattended_url") %>
 <% end -%>
-<% unless [nil, :undefined, :undef].include?(scope.lookupvar("foreman::authentication")) -%>
-:login: <%= scope.lookupvar("foreman::authentication") %>
```
evgeni: this is still referenced in tests

ekohl: Fixed

PR #746, opened 2019-08-07
[WIP] Remove other acceptance tests

ekohl:
This is to verify if the Travis tests pass when this is the only test.


ekohl:
It looks like this is just broken when running in Travis on EL7 altogether. I'm sure this used to work and when I locally run it in a VM, it actually works. This leads me to believe it's poblably something in in how Docker and CentOS7 work together. Leaning to disabling the particular test for now.

@mmoll / @evgeni?

PR #748, opened 2019-08-18
Add rex cockpit integration support

iNecas:
This adds --foreman-plugin-remote-execution-cockpit boolean option to
add support for cockpit installation.

It enables foreman-cockpit.service (comming from
rubygem-foreman_remote_execution-cockpit package), that provides the
backend service for the cockpit tunel, and configures the settings
and authentication for the whole setup to work.

Requires:

- [ ]  https://github.com/theforeman/foreman-packaging/pull/4018.

When testing, SELinux needs to be disabled for now, as there are rules missing
for the foreman-cockpit service to work in enforcing mode.

### DIFF
--- a/manifests/plugin/remote_execution.pp
+++ b/manifests/plugin/remote_execution.pp
@@ -6,39 +6,26 @@
 # $cockpit:: Install cockpit support
 #
 class foreman::plugin::remote_execution (
-  Boolean $cockpit = $::foreman::plugin::remote_execution::params::cockpit,
-) inherits foreman::plugin::remote_execution::params {
+  Boolean $cockpit = false,
+) {
   include ::foreman::plugin::tasks
+  include apache::mod::proxy_wstunnel
+  include apache::mod::proxy_http
 
   foreman::plugin {'remote_execution':
   }
 
+  $cockpit_subpath = 'webcon'
+
   if $cockpit {
-    case $::osfamily {
-      'RedHat': {
-        case $::operatingsystem {
-          'fedora': {
-            $cockpit_package = 'rubygem-foreman_remote_execution-cockpit'
-          }
-          default: {
-            $cockpit_package = 'tfm-rubygem-foreman_remote_execution-cockpit'
-          }
-        }
-      }
-      default: {
-        fail("${::hostname}: foreman_remote_execution cockpit integration does not support osfamily ${::osfamily}")
-      }
+    if $::osfamily != 'RedHat' {
+      fail("${::hostname}: foreman_remote_execution cockpit integration does not support osfamily ${::osfamily}")
     }
 
-    package { $cockpit_package:
-      ensure => present,
-    }
-    ->
-    service { 'foreman-cockpit':
-      ensure     => running,
-      enable     => true,
-      hasstatus  => true,
-      hasrestart => true,
+    foreman::plugin { 'remote_execution-cockpit': }
+    -> service { 'foreman-cockpit':
+      ensure => running,
+      enable => true,
     }
 
     file { '/etc/foreman/cockpit/cockpit.conf':
@@ -64,7 +51,7 @@ class foreman::plugin::remote_execution (
     }
 
     foreman_config_entry { 'remote_execution_cockpit_url':
-      value => "/webcon/=%{host}",
+      value => "/${cockpit_subpath}/=%{host}",
     }
   }
 }
--- a/manifests/plugin/remote_execution.pp
+++ b/manifests/plugin/remote_execution.pp
@@ -1,7 +1,70 @@
+# = Foreman Remote Execution
 # Installs foreman_remote_execution plugin
-class foreman::plugin::remote_execution {
+#
+# === Parameters:
+#
+# $cockpit:: Install cockpit support
+#
+class foreman::plugin::remote_execution (
+  Boolean $cockpit = $::foreman::plugin::remote_execution::params::cockpit,
+) inherits foreman::plugin::remote_execution::params {
   include ::foreman::plugin::tasks
 
   foreman::plugin {'remote_execution':
+  }
+
+  if $cockpit {
+    case $::osfamily {
+      'RedHat': {
+        case $::operatingsystem {
+          'fedora': {
+            $cockpit_package = 'rubygem-foreman_remote_execution-cockpit'
+          }
+          default: {
+            $cockpit_package = 'tfm-rubygem-foreman_remote_execution-cockpit'
+          }
+        }
+      }
+      default: {
+        fail("${::hostname}: foreman_remote_execution cockpit integration does not support osfamily ${::osfamily}")
+      }
+    }
+
+    package { $cockpit_package:
+      ensure => present,
+    }
+    ->
+    service { 'foreman-cockpit':
+      ensure     => running,
+      enable     => true,
+      hasstatus  => true,
+      hasrestart => true,
+    }
+
+    file { '/etc/foreman/cockpit/cockpit.conf':
+      ensure  => file,
+      owner   => 'root',
+      group   => 'root',
+      mode    => '0644',
+      content => template('foreman/remote_execution_cockpit.conf.erb'),
+      notify  => Service['foreman-cockpit'],
+    }
+
+    file { '/etc/foreman/cockpit/foreman-cockpit-session.yml':
+      ensure  => file,
+      owner   => 'root',
+      group   => 'root',
+      mode    => '0644',
+      content => template('foreman/remote_execution_cockpit_session.yml.erb'),
+      notify  => Service['foreman-cockpit'],
+    }
+
+    foreman::config::apache::fragment { 'cockpit':
+      ssl_content => template('foreman/cockpit-apache-ssl.conf.erb'),
+    }
+
+    foreman_config_entry { 'remote_execution_cockpit_url':
+      value => "/webcon/=%{host}",
+    }
   }
 }
--- a/templates/cockpit-apache-ssl.conf.erb
+++ b/templates/cockpit-apache-ssl.conf.erb
@@ -1,0 +1,14 @@
+### File managed with puppet ###
+
+<Location /webcon>
+  LoadModule proxy_wstunnel_module modules/mod_proxy_wstunnel.so
+  LoadModule proxy_http_module modules/mod_proxy_http.so
+
+  ProxyPreserveHost On
+
+  RewriteEngine On
+  RewriteCond %{HTTP:Upgrade} =websocket [NC]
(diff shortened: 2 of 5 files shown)


manifests/plugin/remote_execution.pp
```diff
@@ -1,7 +1,70 @@
+# = Foreman Remote Execution
 # Installs foreman_remote_execution plugin
-class foreman::plugin::remote_execution {
+#
+# === Parameters:
+#
+# $cockpit:: Install cockpit support
+#
+class foreman::plugin::remote_execution (
+  Boolean $cockpit = $::foreman::plugin::remote_execution::params::cockpit,
```
ekohl: It's ok not to inherit if it's a trivial parameter.

manifests/plugin/remote_execution.pp
```diff
@@ -1,7 +1,70 @@
...
+            $cockpit_package = 'tfm-rubygem-foreman_remote_execution-cockpit'
+          }
+        }
+      }
+      default: {
+        fail("${::hostname}: foreman_remote_execution cockpit integration does not support osfamily ${::osfamily}")
+      }
+    }
+
+    package { $cockpit_package:
```
ekohl: You can replace this with `foreman::plugin { 'remote_execution-cockpit': }` and drop the whole package name determination.

manifests/plugin/remote_execution.pp
```diff
@@ -1,7 +1,70 @@
...
+      ensure => present,
+    }
+    ->
+    service { 'foreman-cockpit':
+      ensure     => running,
+      enable     => true,
+      hasstatus  => true,
```
ekohl: This defaults to true since puppet 2.7 and can be dropped:
```suggestion
```

manifests/plugin/remote_execution.pp
```diff
@@ -1,7 +1,70 @@
...
+      hasrestart => true,
```
ekohl: Same here
```suggestion
```

manifests/plugin/remote_execution.pp
```diff
@@ -1,7 +1,70 @@
...
+      mode    => '0644',
+      content => template('foreman/remote_execution_cockpit_session.yml.erb'),
+      notify  => Service['foreman-cockpit'],
+    }
+
+    foreman::config::apache::fragment { 'cockpit':
+      ssl_content => template('foreman/cockpit-apache-ssl.conf.erb'),
+    }
+
+    foreman_config_entry { 'remote_execution_cockpit_url':
```
ekohl: Why does this need to be configured? Isn't this the default and can we rely on that?

templates/cockpit-apache-ssl.conf.erb
```diff
@@ -0,0 +1,14 @@
+### File managed with puppet ###
+
+<Location /webcon>
+  LoadModule proxy_wstunnel_module modules/mod_proxy_wstunnel.so
```
ekohl: Please don't use LoadModule this way. In the class you should `include apache::mod::proxy_wstunnel`.

templates/cockpit-apache-ssl.conf.erb
```diff
@@ -0,0 +1,14 @@
...
+  LoadModule proxy_http_module modules/mod_proxy_http.so
```
ekohl: Same here with `include apache::mod::proxy_http`

templates/remote_execution_cockpit.conf.erb
```diff
@@ -0,0 +1,14 @@
+[WebService]
+LoginTitle = Foreman Cockpit
+UrlRoot = /webcon/
+Origins = https://<%= scope.lookupvar("::fqdn") %>
```
ekohl: I'd be nice to actually use the configured URL:
```suggestion
Origins = <%= scope.lookupvar("foreman::foreman_url") %>
```

templates/remote_execution_cockpit_session.yml.erb
```diff
@@ -0,0 +1,4 @@
+:foreman_url: https://<%= scope.lookupvar("::fqdn") %>
```
ekohl:
```suggestion
:foreman_url: <%= scope.lookupvar("foreman::foreman_url") %>
```

manifests/plugin/remote_execution.pp
iNecas: The default config is null and it actually determines whether the cockpit integration is enabled or not.

ekohl: Since `/webcon` is repeated a few times, perhaps this should be a variable? Doesn't have to be a class parameter but it's a good way to show that they're the same thing.

manifests/plugin/remote_execution.pp
iNecas: Updated

manifests/plugin/remote_execution.pp
iNecas: Updated

manifests/plugin/remote_execution.pp
iNecas: Updated

manifests/plugin/remote_execution.pp
iNecas: Updated

templates/cockpit-apache-ssl.conf.erb
iNecas: Updated

templates/cockpit-apache-ssl.conf.erb
iNecas: Updated

iNecas:
Ready for review agin

templates/cockpit-apache-ssl.conf.erb
```diff
@@ -0,0 +1,11 @@
+### File managed with puppet ###
+
+<Location /<%= scope.lookupvar("cockpit_subpath") %>>
```
ekohl: It's in the local scope so this is sufficient:

```suggestion
<Location /<%= @cockpit_subpath %>>
```

templates/cockpit-apache-ssl.conf.erb
```diff
@@ -0,0 +1,11 @@
...
+  ProxyPreserveHost On
+
+  RewriteEngine On
+  RewriteCond %{HTTP:Upgrade} =websocket [NC]
+  RewriteRule /<%= scope.lookupvar("cockpit_subpath") %>/(.*)           ws://127.0.0.1:9999/<%= scope.lookupvar("cockpit_subpath") %>/$1 [P]
```
ekohl:
```suggestion
  RewriteRule /<%= @cockpit_subpath %>/(.*)           ws://127.0.0.1:9999/<%= @cockpit_subpath %>/$1 [P]
```

templates/cockpit-apache-ssl.conf.erb
```diff
@@ -0,0 +1,11 @@
...
+  RewriteCond %{HTTP:Upgrade} !=websocket [NC]
+  RewriteRule /<%= scope.lookupvar("cockpit_subpath") %>/(.*)           http://127.0.0.1:9999/<%= scope.lookupvar("cockpit_subpath") %>/$1 [P]
```
ekohl:
```suggestion
  RewriteRule /<%= @cockpit_subpath %>/(.*)           http://127.0.0.1:9999/<%= @cockpit_subpath %>/$1 [P]
```

templates/remote_execution_cockpit.conf.erb
```diff
@@ -0,0 +1,14 @@
+[WebService]
+LoginTitle = Foreman Cockpit
+UrlRoot = /<%= scope.lookupvar("cockpit_subpath") %>/
```
ekohl:
```suggestion
UrlRoot = /<%= @cockpit_subpath %>/
```

manifests/plugin/remote_execution.pp
```diff
@@ -1,7 +1,57 @@
...
+#
+# === Parameters:
+#
+# $cockpit:: Install cockpit support
+#
+class foreman::plugin::remote_execution (
+  Boolean $cockpit = false,
+) {
   include ::foreman::plugin::tasks
+  include apache::mod::proxy_wstunnel
```
ekohl: Can you move this into the `if $cockpit` block?

manifests/plugin/remote_execution.pp
```diff
@@ -1,7 +1,57 @@
...
+  include apache::mod::proxy_http
 
   foreman::plugin {'remote_execution':
   }
+
+  $cockpit_subpath = 'webcon'
+
+  if $cockpit {
+    if $::osfamily != 'RedHat' {
```
ekohl: I'm a bit on the fence about this. Is there a reason why it can't work on Debian?

manifests/plugin/remote_execution.pp
```diff
@@ -1,7 +1,57 @@
...
+      fail("${::hostname}: foreman_remote_execution cockpit integration does not support osfamily ${::osfamily}")
+    }
+
+    foreman::plugin { 'remote_execution-cockpit': }
+    -> service { 'foreman-cockpit':
+      ensure => running,
+      enable => true,
+    }
+
+    file { '/etc/foreman/cockpit/cockpit.conf':
```
ekohl: If the package creates `/etc/foreman/cockpit` then it also needs a require on `Foreman::Plugin['remote_execution-cockpit']`. Otherwise you need to ensure the directory exists.

iNecas:
Unfortunately, my focus has already shifted quite far from this PR, therefore I would suggest someone else to pick this up to finish the comments. @ekohl with you Puppet experience, it seems like it might be quite low hanging fruit for you to shape it the way you need? Otherwise, reaching to @ares to try to find another set of hands for this one. It could be good also for actually sharing some knowledge about this functionality.

tstrych:
Not sure if this helps, but I was able with @adamruzicka help to install https://github.com/theforeman/foreman-packaging/pull/3816 and patch it with this PR and cockpit seems to work correctly.

Hopefully this will be merged soon, as this is last step from developer side of work, for now.

ekohl:
Replaced by https://github.com/theforeman/puppet-foreman/pull/760

templates/remote_execution_cockpit_session.yml.erb
```diff
@@ -0,0 +1,4 @@
+:foreman_url: <%= scope.lookupvar("foreman::foreman_url") %>
+:ssl_ca_file: <%= scope.lookupvar("foreman::client_ssl_ca") %>
```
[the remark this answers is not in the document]
iNecas: ack on that, it should probably be the server ca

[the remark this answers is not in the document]
iNecas: Thanks for taking over this.

PR #751, opened 2019-08-27
Fixes #27716 - Mark foreman::compute::*::version as advanced

ekohl:
Most users won't change this away from installed to a specific version or latest. Especially in the installer. By making it an advanced parameter, we hide this from most users who won't care.


ekohl:
Do not merge yet. This is still not effective.

ekohl:
It appears Kafo can't handle a class with only advanced parameters.

PR #754, opened 2019-08-28
Add supervisory_authority plugin

laugmanuel:
needs: https://github.com/theforeman/foreman-packaging/pull/3625


templates/foreman_supervisory_authority.yaml.erb
```diff
@@ -0,0 +1,12 @@
+:foreman_supervisory_authoriy:
```
ekohl: Is there a typo here?
```suggestion
:foreman_supervisory_authority:
```

manifests/plugin/supervisory_authority.pp
```diff
@@ -0,0 +1,45 @@
+# = Foreman supervisory_authority Plugin
+#
+# This class installs the supervisory_authority plugin and configuration file
+#
+# === Parameters:
```
ekohl: Would it make sense to split this up into `Parameters` and `Advanced Parameters`? If not, I prefer the puppet-strings syntax since it's the officially supported one. However, RDoc has the feature of parameter groups which we do use in our installer.

templates/foreman_supervisory_authority.yaml.erb
laugmanuel: Thank's for noticing this! This is actually a typo in the plugin itself (from which the settings file was copied). See: https://github.com/timogoebel/foreman_supervisory_authority/pull/6

laugmanuel: I fixed the template in this MR. I guess we should wait for the new plugin version to become available so we don't brake Foreman

manifests/plugin/supervisory_authority.pp
laugmanuel: I added the "Advanced Parameters" section with all parameters which are optional.
The doc-block was taken from another plugin (memcache) to be consistent with whats already there.
I think we should rework the RDoc in all the Puppet files at once to not have different styles of documenting stuff.
What do you think?

laugmanuel:
I did change the default values of some optional/advanced parameters to correspond with the default values of the elastic documentation.

ekohl:
Thanks!

PR #759, opened 2019-09-18
Release 11.0.3 with the actual fix for #26695

ekohl:
dca5d41234bf54470ec96e3858caabbc4970f8f6 was intended to reduce fb5d38f8bea69ad8c6cb1300ce4d803fd89aace3 to an implementation only change but ended up being a full revert. This removes the dead setting that actively breaks installations.


mmoll:
merged, bedankt @ekohl!

PR #761, opened 2019-10-16
Fixes #28067 - dynflow sidekiq services config

ezr-ondrej:
Install necessary sidekiq configs and dependencies.
- [x] merge theforeman/foreman#7249 to get the tests to pass.

### DIFF
--- a/manifests/config.pp
+++ b/manifests/config.pp
@@ -6,6 +6,19 @@ class foreman::config {
     Class['puppet::server::install'] -> Class['foreman::config']
   }
 
+  if $::foreman::jobs_manage_service {
+    if $::foreman::jobs_sidekiq_redis_url != undef {
+      $jobs_redis_url = $::foreman::jobs_sidekiq_redis_url
+    } else {
+      include ::redis
+      $jobs_redis_url = "redis://localhost:${::redis::port}/6"
+    }
+
+    file { '/etc/foreman/dynflow':
+      ensure => directory,
+    }
+  }
+
   concat::fragment {'foreman_settings+01-header.yaml':
     target  => '/etc/foreman/settings.yaml',
     content => template('foreman/settings.yaml.erb'),
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -179,13 +179,15 @@
 #
 # $telemetry_logger_level::       Telemetry debugging logs level
 #
-# $dynflow_pool_size::            How many workers should Dynflow use
+# $dynflow_pool_size::            How many threads per worker process should Dynflow use
 #
-# $jobs_service::                 Name of the service for running the background Dynflow executor
+# $jobs_manage_service::          Whether to manage the dynflow services
 #
 # $jobs_service_enable::          Whether the Dynflow executor should be enabled
 #
 # $jobs_service_ensure::          Whether the Dynflow executor should be running or stopped
+#
+# $jobs_sidekiq_redis_url::       If set, the redis server is not managed and we use the defined url to connect
 #
 # $hsts_enabled::                 Should HSTS enforcement in https requests be enabled
 #
@@ -278,9 +280,10 @@ class foreman (
   Boolean $telemetry_logger_enabled = $::foreman::params::telemetry_logger_enabled,
   Enum['DEBUG', 'INFO', 'WARN', 'ERROR', 'FATAL'] $telemetry_logger_level = $::foreman::params::telemetry_logger_level,
   Stdlib::Port $dynflow_pool_size = $::foreman::params::dynflow_pool_size,
-  String $jobs_service = $::foreman::params::jobs_service,
+  Boolean $jobs_manage_service = $::foreman::params::jobs_manage_service,
   Stdlib::Ensure::Service $jobs_service_ensure = $::foreman::params::jobs_service_ensure,
   Boolean $jobs_service_enable = $::foreman::params::jobs_service_enable,
+  Optional[Redis::RedisUrl] $jobs_sidekiq_redis_url = $::foreman::params::jobs_sidekiq_redis_url,
   Boolean $hsts_enabled = $::foreman::params::hsts_enabled,
   Array[Stdlib::HTTPUrl] $cors_domains = $::foreman::params::cors_domains,
 ) inherits foreman::params {
--- a/manifests/service.pp
+++ b/manifests/service.pp
@@ -8,13 +8,15 @@ class foreman::service(
   String $foreman_service = $::foreman::foreman_service,
   Stdlib::Ensure::Service $foreman_service_ensure = $::foreman::foreman_service_ensure,
   Boolean $foreman_service_enable = $::foreman::foreman_service_enable,
-  String $jobs_service = $::foreman::jobs_service,
-  Stdlib::Ensure::Service $jobs_service_ensure = $::foreman::jobs_service_ensure,
-  Boolean $jobs_service_enable = $::foreman::jobs_service_enable,
+  Boolean $jobs_manage_service = $::foreman::jobs_manage_service,
 ) {
-  service { $jobs_service:
-    ensure => $jobs_service_ensure,
-    enable => $jobs_service_enable,
+  if $jobs_manage_service {
+    foreman::dynflow::worker { 'orchestrator':
+      concurrency => 1,
+      queues      => ['dynflow_orchestrator'],
+    }
+
+    foreman::dynflow::worker { 'worker': }
   }
 
   if $apache {
--- a/metadata.json
+++ b/metadata.json
@@ -35,6 +35,10 @@
     {
       "name": "puppet/extlib",
       "version_requirement": ">= 3.0.0 < 5.0.0"
+    },
+    {
+      "name": "puppet/redis",
+      "version_requirement": ">= 5.0.0 < 6.0.0"
     }
   ],
   "requirements": [
--- a/spec/acceptance/foreman_basic_spec.rb
+++ b/spec/acceptance/foreman_basic_spec.rb
@@ -28,6 +28,12 @@ describe 'Scenario: install foreman' do
       manage_repo => false,
     }
 
+    if $facts['os']['family'] == 'RedHat' and $facts['os']['name'] != 'Fedora' {
+      class { 'redis::globals':
+        scl => 'rh-redis5',
+      }
+    }
+
     $directory = '/etc/foreman'
     $certificate = "${directory}/certificate.pem"
     $key = "${directory}/key.pem"
@@ -68,7 +74,12 @@ describe 'Scenario: install foreman' do
     it { is_expected.to be_running }
   end
 
-  describe service('dynflowd') do
+  describe service('dynflow-sidekiq@orchestrator') do
+    it { is_expected.to be_enabled }
+    it { is_expected.to be_running }
+  end
+
+  describe service('dynflow-sidekiq@worker') do
     it { is_expected.to be_enabled }
     it { is_expected.to be_running }
   end
--- a/spec/classes/foreman_service_spec.rb
+++ b/spec/classes/foreman_service_spec.rb
@@ -15,9 +15,7 @@ describe 'foreman::service' do
       foreman_service: 'foreman',
       foreman_service_ensure: 'running',
       foreman_service_enable: true,
-      jobs_service: 'dynflower',
-      jobs_service_ensure: 'stopped',
-      jobs_service_enable: false
+      jobs_manage_service: false
     }
   end
 
@@ -28,7 +26,6 @@ describe 'foreman::service' do
     context 'with ssl' do
       let(:params) { super().merge(ssl: true) }
       it { is_expected.to compile.with_all_deps }
-      it { is_expected.to contain_service('dynflower').with_ensure('stopped').with_enable(false) }
 
       it 'should restart passenger' do
         should contain_exec('restart_foreman')
--- a/spec/spec_helper_acceptance.rb
+++ b/spec/spec_helper_acceptance.rb
@@ -29,6 +29,9 @@ RSpec.configure do |c|
         on host, 'sed -i "s/keepcache=.*/keepcache=1/" /etc/yum.conf'
         # refresh check if cache needs refresh on next yum command
         on host, 'yum clean expire-cache'
+
+        host.install_package('centos-release-scl-rh')
+        host.install_package('rh-redis5-redis')
       end
     end
   end
--- a/templates/dynflow_worker.yml.erb
+++ b/templates/dynflow_worker.yml.erb
@@ -1,0 +1,5 @@
+:concurrency: <%= scope['concurrency'] %>
+:queues:
+<% scope['queues'].each do |queue| -%>
+  - <%= queue %>
+<% end %>
--- a/templates/settings.yaml.erb
+++ b/templates/settings.yaml.erb
@@ -72,8 +72,11 @@
     # logging level as in Logger::LEVEL
     :level: '<%= scope.lookupvar("foreman::telemetry_logger_level") %>'
 
+<% if scope.lookupvar("foreman::jobs_manage_service") -%>
 :dynflow:
   :pool_size: <%= scope.lookupvar("foreman::dynflow_pool_size") %>
+  :redis_url: <%= scope.lookupvar("foreman::config::jobs_redis_url") %>
+<% end -%>
 <% if scope.lookupvar("foreman::apache") && !scope.lookupvar("foreman::passenger") -%>
 
 # Configure reverse proxy headers
--- a/.fixtures.yml
+++ b/.fixtures.yml
@@ -12,6 +12,7 @@ fixtures:
     extlib:           'https://github.com/voxpupuli/puppet-extlib'
     postgresql:       'https://github.com/puppetlabs/puppetlabs-postgresql'
     puppet:           'https://github.com/theforeman/puppet-puppet'
+    redis:            'https://github.com/voxpupuli/puppet-redis'
     systemd:          'https://github.com/camptocamp/puppet-systemd'
     selinux_core:
       repo: "https://github.com/puppetlabs/puppetlabs-selinux_core"
--- a/Gemfile
+++ b/Gemfile
@@ -30,6 +30,7 @@ gem 'beaker', '>= 4.2.0', {"groups"=>["system_tests"]}
 gem 'beaker-docker', {"groups"=>["system_tests"]}
 gem 'beaker-hostgenerator', '>= 1.1.10', {"groups"=>["system_tests"]}
 gem 'beaker-puppet', {"groups"=>["system_tests"]}
+gem 'beaker-vagrant', {"groups"=>["system_tests"]}
 gem 'beaker-rspec', {"groups"=>["system_tests"]}
 gem 'beaker-module_install_helper', {"groups"=>["system_tests"]}
 gem 'beaker-puppet_install_helper', {"groups"=>["system_tests"]}
--- a/manifests/dynflow/worker.pp
+++ b/manifests/dynflow/worker.pp
@@ -1,0 +1,26 @@
+# Initialize the dynflow worker
+# @param concurrency
+#   Defines number of threads used for job processing
+define foreman::dynflow::worker (
+  String $service_name = $name,
+  Integer[1] $concurrency = $foreman::dynflow_pool_size,
+  Array[String] $queues = ['default', 'remote_execution'],
+  String $config_owner = 'root',
+  String $config_group = $::foreman::group,
+  Stdlib::Ensure::Service $service_ensure = $::foreman::jobs_service_ensure,
+  Boolean $service_enable = $::foreman::jobs_service_enable
+) {
+
+  file { "/etc/foreman/dynflow/${service_name}.yml":
+    ensure  => file,
+    owner   => $config_owner,
+    group   => $config_group,
+    mode    => '0644',
+    content => template('foreman/dynflow_worker.yml.erb'),
+  }
+
+  service { "dynflow-sidekiq@${service_name}":
+    ensure => $service_ensure,
+    enable => $service_enable,
+  }
+}
--- a/manifests/install.pp
+++ b/manifests/install.pp
@@ -25,6 +25,12 @@ class foreman::install {
     }
   }
 
+  if $::foreman::jobs_manage_service {
+    package { 'foreman-dynflow-sidekiq':
+      ensure => installed,
+    }
+  }
+
   if $::foreman::ipa_authentication and $::foreman::ipa_manage_sssd {
     package { 'sssd-dbus':
       ensure => installed,
--- a/manifests/params.pp
+++ b/manifests/params.pp
@@ -85,9 +85,11 @@ class foreman::params {
   $foreman_service_port = 3000
 
   # Define job processing service properties
-  $jobs_service = 'dynflowd'
+  $jobs_manage_service = true
   $jobs_service_ensure = 'running'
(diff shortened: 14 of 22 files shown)


manifests/config.pp
```diff
@@ -80,6 +80,35 @@
     ensure  => absent,
   }
 
+  if $::foreman::jobs_sidekiq {
+    if $::foreman::jobs_sidekiq_redis_url == undef {
+      class { '::redis':
+        default_install => false,
```
ezr-ondrej: I would rather keep this var passing if possible - we might need to include the class with default true in redis cache.

ekohl: So that's the part you can't do. Classes are singleton. That's why I'd prefer to keep it just to `include redis` and keep the control of it out of this module.

ezr-ondrej: Ok, including now instead.

manifests/service.pp
```diff
@@ -11,10 +11,30 @@
...
+    $sidekiq_enable = $jobs_service_enable
+    $dynflowd_ensure = 'stopped'
+    $dynflowd_enable = false
+  } else {
+    $dynflowd_ensure = $jobs_service_ensure
+    $dynflowd_enable = $jobs_service_enable
+    $sidekiq_enable = false
+    $sidekiq_ensure = 'stopped'
+  }
+  service { 'dynflow-sidekiq@orchestrator':
```
ekohl: I now realize that these services might be a bit more complicated. If we don't install the package, they won't exist. That means these should be within the `if $jobs_sidekiq`. The dynflow service always exists so can be outside of the `if`

ezr-ondrej: yeah righ, did that, just unsure how to clean them up... If only we would have an `foreman-maintain cleanup` taks :thinking: :rofl:

manifests/init.pp
```diff
@@ -305,6 +311,9 @@
   String $jobs_service = $::foreman::params::jobs_service,
   Stdlib::Ensure::Service $jobs_service_ensure = $::foreman::params::jobs_service_ensure,
   Boolean $jobs_service_enable = $::foreman::params::jobs_service_enable,
+  Boolean $jobs_sidekiq = $::foreman::params::jobs_sidekiq,
```
ehelms: Should these all be `dynflow_sidekiq` to keep them congruent with the service and configuration names?

ehelms: I am also a bit worried it may be confusing to users if they are able to set `jobs_service_enable: true` and `jobs_sidekiq: true` but then only the dynflow-sidekiq process is running. Their configuration settings would not reflect the state of the system. We should maybe then have these two options conflict as a validation in `init.pp` ?

manifests/config.pp
```diff
@@ -80,6 +80,34 @@
     ensure  => absent,
   }
 
+  if $::foreman::jobs_sidekiq {
+    if $::foreman::jobs_sidekiq_redis_url == undef {
+      include ::redis
+
+      redis::instance { 'dynflow':
```
ehelms: Does an instance refer to a "database" inside Redis or too the Redis process itself? If the latter, will this tie all future deployments and use of Redis (e.g. caching or Pulp 3) to an instance named for dynflow?

manifests/config.pp
```diff
@@ -80,6 +80,34 @@
...
+    }
+
+    file { '/etc/foreman/dynflow/orchestrator.yml':
+      owner   => 'root',
+      group   => $::foreman::group,
+      mode    => '0640',
+      content => template('foreman/dynflow_orchestrator.yml.erb'),
+    }
+
+    file { '/etc/foreman/dynflow/worker.yml':
```
ehelms: We are supporting one and only one worker out of the box right? So in the future this would get updated to deploy additional `worker.yml` files based on some user input?

ezr-ondrej:
[test continuous-integration/travis-ci/pr]

ezr-ondrej:
the tests doesn't seem to kick on

manifests/init.pp
ezr-ondrej: In current state, we are giving the option(s) new meaning. That is - `job_service_enable` - enable the service, it depends on `jobs_sidekiq` what service it's going to be.

manifests/config.pp
ezr-ondrej: The `instance` is completly separate proces, with its own config and process. So if we decide to use redis for caching, there should be another instance. This should mean dynflow has it's own instance and configuration - meaning not sharing it with other use cases of redis.

manifests/config.pp
ezr-ondrej: Yeah, exactly.

tbrisker:
Tests are failing?

ezr-ondrej:
> Tests are failing?

As we have it default on, it needs the package, that is not in yet

manifests/config.pp
ehelms: I suppose we can clean that up in the future as I think having multiple Redis instances may not be the cleanest design or requirement for our users. But generally I think I would prefer a single instance for Foreman if possible.

manifests/init.pp
ehelms: If I set `jobs_sidekiq` to true and `jobs_service_enable` to false, I think the current logic still enables the `dynflow-sidekiq` processes which would be unexpected based on `job_service_enable` controlling whether services are enabled or not.

I also still heavily lean towards naming conventions matching closer to the service, e.g. `jobs_sidekiq` becoming `jobs_dynflow_sidekiq`.

tbrisker:
To summarize our discussion from https://community.theforeman.org/t/release-meeting-notes-2019-10-23/15883 - this will be merged only post branching and will be supported in 1.25.

ekohl:
Adding the label Waiting on contributor to make it easier for me see it's not ready for merge when going over the open PRs/issues.

metadata.json
```diff
@@ -35,6 +35,10 @@
     {
       "name": "puppet/extlib",
       "version_requirement": ">= 3.0.0 < 5.0.0"
+    },
+    {
+      "name": "puppet/redis",
+      "version_requirement": ">= 4.2.0 < 5.0.0"
```
ekohl: FYI: I just released 5.0.0 which includes the support we want.
```suggestion
      "version_requirement": ">= 5.0.0 < 6.0.0"
```

I'm about to test this on a real system with this patch.

ezr-ondrej: Awesome! :heart:  Big :+1:

templates/dynflow_worker.yml.erb
```diff
@@ -0,0 +1,5 @@
+:concurrency: <%= scope['concurrency'] %>
+:queues:
```
ehelms: @ezr-ondrej @adamruzicka I don't know if you have considered or if it would be possible, but a future looking idea that could be nice is making the queues a setting that's changeable without restart. This would simplify the installer support, make the workers less reliant on a config file and restarts and potentially give the ability to dynamically scale and assign out queues based on workload from the application itself via user config or "smarts".

adamruzicka: It would be nice, but I'm not sure if sidekiq supports changing queue configuration without restarting the worker. Still, might be worth looking into it

ehelms:
Looks like this may be down fixing failing tests to finish it?

templates/settings.yaml.erb
```diff
@@ -74,6 +74,7 @@
 
 :dynflow:
```
ekohl: Just a general question, also to @adamruzicka. Why do we configure the dynflow pool size and redis URL here when we also have a worker config? Does the main Foreman ever use these settings?

spec/acceptance/foreman_basic_spec.rb
```diff
@@ -27,6 +27,9 @@
     class { '::apache::mod::passenger':
       manage_repo => false,
     }
+    class { 'redis::globals':
```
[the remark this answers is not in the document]
ezr-ondrej: This unfortunately doesn't work :
```
Warning: /Stage[main]/Redis::Config/File[/var/log/redis]: Skipping because of failed dependencies
       	Warning: /Stage[main]/Redis::Config/File[/var/opt/rh/rh-redis5/lib/redis]: Skipping because of failed dependencies
       	Warning: /Stage[main]/Redis::Ulimit/File[/etc/systemd/system/rh-redis5-redis.service.d/]: Skipping because of failed dependencies
       	Warning: /Stage[main]/Redis::Ulimit/File[/etc/systemd/system/rh-redis5-redis.service.d/limit.conf]: Skipping because of failed dependencies
       	Warning: /Stage[main]/Redis::Ulimit/Augeas[Systemd redis ulimit]: Skipping because of failed dependencies
       	Warning: /Stage[main]/Redis/Exec[systemd-reload-redis]: Skipping because of failed dependencies
       	Warning: /Stage[main]/Redis::Config/Redis::Instance[default]/File[/etc/opt/rh/rh-redis5/redis.conf.puppet]: Skipping because of failed dependencies
       	Warning: /Stage[main]/Redis::Config/Redis::Instance[default]/Exec[cp -p /etc/opt/rh/rh-redis5/redis.conf.puppet /etc/opt/rh/rh-redis5/redis.conf]: Skipping because of failed dependencies
       	Warning: /Stage[main]/Redis::Service/Service[rh-redis5-redis]: Skipping because of failed dependencies
```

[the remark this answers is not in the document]
ezr-ondrej: It seems to fail, because the package is already in :thinking:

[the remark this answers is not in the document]
ekohl: To be clear what I meant:
https://github.com/theforeman/puppet-foreman/blob/36aa5389b7da41d4770fa61a1c10a78239695735/spec/spec_helper_acceptance.rb#L26-L32
If you add `host.install_package('rh-redis5-redis')` there I think you'll avoid the SELinux idempotence issue.

spec/spec_helper_acceptance.rb
```diff
@@ -29,6 +29,9 @@
         on host, 'sed -i "s/keepcache=.*/keepcache=1/" /etc/yum.conf'
         # refresh check if cache needs refresh on next yum command
         on host, 'yum clean expire-cache'
+
+        host.install_package('centos-release-scl')
```
ekohl: Small optimization: it's sufficient to only install the RH set of SCLs and that's also what we do in real deployments.
```suggestion
        host.install_package('centos-release-scl-rh')
```

ezr-ondrej:
For real? xD The tests aren't broken because of me! :sob:
see: theforeman/foreman#7249 for the fix

spec/spec_helper_acceptance.rb
ezr-ondrej: Thanks, makes sense :)

templates/settings.yaml.erb
```diff
@@ -74,6 +74,7 @@
...
   :pool_size: <%= scope.lookupvar("foreman::dynflow_pool_size") %>
+  :redis_url: <%= scope.lookupvar("foreman::jobs_sidekiq_redis_url") || "redis://localhost:#{scope.lookupvar("foreman::jobs_sidekiq_redis_port") || 6379}/6" %>
```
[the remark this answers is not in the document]
ezr-ondrej: I can't do that I would need to do something fancy to add to the setting file, because at the point settings file is evaluated the `foreman::jobs` class is not.

templates/settings.yaml.erb
ezr-ondrej: Actually it doesn't need it at all, although do we want to configure the URL in the job config itself?

ezr-ondrej: For the record it would be harder to put it in sidekiq process setting as it is sidekiq config file now. So to put other optios (like sidekiq url) we would need to parse it by ourselves.

ohadlevy: just wondering, would that work with https://github.com/theforeman/foreman/pull/7153#discussion_r356590683 ?

manifests/jobs.pp
```diff
@@ -0,0 +1,22 @@
+# Configure dynflow
+class foreman::jobs {
```
ekohl: Bikeshedding: Should this be called `foreman::dynflow`?

templates/settings.yaml.erb
```diff
@@ -74,6 +74,7 @@
 
 :dynflow:
   :pool_size: <%= scope.lookupvar("foreman::dynflow_pool_size") %>
+  :redis_url: <%= scope.lookupvar("foreman::jobs_sidekiq_redis_url") || "redis://localhost:6379/6" %>
```
ekohl: This should use the Redis variable:

```suggestion
  :redis_url: <%= scope.lookupvar("foreman::jobs::redis_url") %>
```

You said this fails because the variable is not instantiated at that point which makes me think we should move logic from `foreman::jobs` to `foreman::config`.

An alternative would be to set all the dynflow settings (`pool_size`, `redis_url`) in `worker.yaml` but that might require changes in dynflow itself.

spec/classes/foreman_jobs_spec.rb
```diff
@@ -0,0 +1,28 @@
...
+describe 'foreman' do
+  on_os_under_test.each do |os, facts|
+    context "on #{os}" do
+      let(:facts) { facts }
+      let(:params) { {} }
+
+      context 'with default params' do
+        it { should contain_class('redis') }
+        it { should contain_redis__instance('default') }
+        it { should contain_file('/etc/foreman/dynflow/').with_ensure('directory') }
```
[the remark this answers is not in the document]
ezr-ondrej: No idea why it passes :man_shrugging: Fixed :heavy_check_mark:

manifests/jobs.pp
ezr-ondrej: We would have to rename the attributes as well then :worried:

templates/settings.yaml.erb
ezr-ondrej: If we put it to the `config.pp` we can't use `redis::port` as it is not defined yet.

manifests/config.pp
```diff
@@ -80,6 +80,17 @@
     ensure  => absent,
   }
 
+  if $::foreman::jobs_sidekiq_redis_url != undef {
```
ekohl: Should we wrap this whole block in `if $::foreman::jobs_manage_service {` like is done in `install.pp`?

manifests/service.pp
```diff
@@ -42,4 +34,11 @@
       enable => $foreman_service_enable,
     }
   }
+
+  foreman::dynflow::worker { 'orchestrator':
```
ekohl: Perhaps this should also be wrapped in `if $::foreman::jobs_manage_service {` rather than at the worker level.

manifests/config.pp
ezr-ondrej: probably, yeah it would be more straight forward.

ekohl:
I think this looks correct. The acceptance tests will fail on the specific worker tests, but others should pass when merging. Will wait for that.

spec/classes/foreman_service_spec.rb
```diff
@@ -14,10 +14,7 @@
       use_foreman_service: false,
       foreman_service: 'foreman',
       foreman_service_ensure: 'running',
-      foreman_service_enable: true,
-      jobs_service: 'dynflower',
-      jobs_service_ensure: 'stopped',
-      jobs_service_enable: false
+      foreman_service_enable: true
```
ekohl: Probably easier to set this to false so the tests pass. It's a bit complex to get the foreman variables and it's tested in other places as well. You can do it in `foreman_spec.rb` for example. This test is mostly there to test the apache logic that can be quite complex.

templates/settings.yaml.erb
```diff
@@ -74,6 +74,7 @@
 
 :dynflow:
   :pool_size: <%= scope.lookupvar("foreman::dynflow_pool_size") %>
+  :redis_url: <%= scope.lookupvar("foreman::config::jobs_redis_url") %>
```
ehelms: I am seeing this error when I test this with the full workflow:

```
[ WARN 2020-01-14T14:34:33 verbose]  Undefined variable '::foreman::config::jobs_redis_url'; 
```

ehelms:
I tested this out on a Katello nightly. Installation was successful but bats test are failing for most anything task related.

My forklift box:

```
dynflow:
  box: centos7-katello-nightly
  ansible:
    variables:
      foreman_repositories_environment: staging
      katello_repositories_environment: staging
      foreman_installer_module_prs:
        - theforeman/foreman/761
```

After running `vagrant up dynflow`, I ran the bats tests:

```
ansible-playbook playbook/bats.yml -l dynflow
```

templates/settings.yaml.erb
ekohl: In the `if` a line higher it should check `foreman::jobs_manage_service` because that guards whether the variable is defined at all. I also think the entire dynflow section could be guarded. If we don't manage the service, nothing should read its content.

ezr-ondrej:
I think you need this: https://github.com/theforeman/foreman-installer/pull/400

The failure seems to be in the ping command though. It fails on status `dynflowd` but I don't see it validated in the ping command :thinking:

ehelms:
Will run through a few tests and report back

ekohl:
I've kicked off https://ci.theforeman.org/job/foreman-nightly-deb-pipeline/263/ and will restart the Debian tests after that.

ehelms:
I'm getting less bats failures now, woot! But I am still seeing an issue with one related to package installation failing:

```
 ok 29 enable content view repo
    not ok 30 install katello-host-tools
    # (from function `tIsRedHatCompatible' in file os_helper.bash, line 4,
    #  from function `tPackageInstall' in file os_helper.bash, line 89,
    #  in test file fb-content-katello.bats, line 194)
    #   `tPackageInstall katello-host-tools && tPackageExists katello-host-tools' failed
    # Loaded plugins: fastestmirror, product-id, search-disabled-repos, subscription-
    #               : manager
    # 1 local certificate has been deleted.
    # Loading mirror speeds from cached hostfile
    #  * base: mirror.math.princeton.edu
    #  * centos-sclo-rh: mirror.fileplanet.com
    #  * epel: download-ib01.fedoraproject.org
    #  * extras: mirrors.mit.edu
    #  * updates: repo1.ash.innoscale.net
    # No package katello-host-tools available.
    # Error: Nothing to do
```

This makes me think one of the previous tasks is failing in some way or incomplete but is not registering as such. Needs further investigation before we merge. As this will surely result in a broken nightly.  

My steps were the same as before. I spin up a machine with this puppet module patch, and latest nightly and ran bats tests against it.

ekohl:
Perhaps it was an issue that the old dynflowd was still running? After https://github.com/theforeman/foreman-packaging/pull/4442 is merged, that should be fixed in a `yum update` but until then it might interfere.

ehelms:
It appeared dead when I checked:

```
[root@dynflow vagrant]# systemctl status dynflowd
● dynflowd.service - Foreman jobs daemon
   Loaded: loaded (/usr/lib/systemd/system/dynflowd.service; disabled; vendor preset: disabled)
   Active: inactive (dead)
     Docs: https://theforeman.org
```

mmoll:
Debian seems to fail on the dynflow-orchestrator service now :broken_heart:

adamruzicka:
```
+Jan 17 21:49:32 debian10-64.example.com systemd[1]: Started Foreman jobs daemon - orchestrator on sidekiq.
+Jan 17 21:49:33 debian10-64.example.com dynflow-sidekiq@orchestrator[9064]: 2020-01-17T21:49:33.658Z 9064 TID-gpi086500 INFO: ==================================================================
+Jan 17 21:49:33 debian10-64.example.com dynflow-sidekiq@orchestrator[9064]: 2020-01-17T21:49:33.659Z 9064 TID-gpi086500 INFO:   Please point sidekiq to a Rails 4/5 application or a Ruby file
+Jan 17 21:49:33 debian10-64.example.com dynflow-sidekiq@orchestrator[9064]: 2020-01-17T21:49:33.659Z 9064 TID-gpi086500 INFO:   to load your worker classes with -r [DIR|FILE].
+Jan 17 21:49:33 debian10-64.example.com dynflow-sidekiq@orchestrator[9064]: 2020-01-17T21:49:33.659Z 9064 TID-gpi086500 INFO: ==================================================================
+Jan 17 21:49:33 debian10-64.example.com dynflow-sidekiq@orchestrator[9064]: 2020-01-17T21:49:33.659Z 9064 TID-gpi086500 INFO: sidekiq [options]
+Jan 17 21:49:33 debian10-64.example.com dynflow-sidekiq@orchestrator[9064]:     -c, --concurrency INT            processor threads to use
+Jan 17 21:49:33 debian10-64.example.com dynflow-sidekiq@orchestrator[9064]:     -d, --daemon                     Daemonize process
+Jan 17 21:49:33 debian10-64.example.com dynflow-sidekiq@orchestrator[9064]:     -e, --environment ENV            Application environment
+Jan 17 21:49:33 debian10-64.example.com dynflow-sidekiq@orchestrator[9064]:     -g, --tag TAG                    Process tag for procline
+Jan 17 21:49:33 debian10-64.example.com dynflow-sidekiq@orchestrator[9064]:     -i, --index INT                  unique process index on this machine
+Jan 17 21:49:33 debian10-64.example.com dynflow-sidekiq@orchestrator[9064]:     -q, --queue QUEUE[,WEIGHT]       Queues to process with optional weights
+Jan 17 21:49:33 debian10-64.example.com dynflow-sidekiq@orchestrator[9064]:     -r, --require [PATH|DIR]         Location of Rails application with workers or file to require
+Jan 17 21:49:33 debian10-64.example.com dynflow-sidekiq@orchestrator[9064]:     -t, --timeout NUM                Shutdown timeout
+Jan 17 21:49:33 debian10-64.example.com dynflow-sidekiq@orchestrator[9064]:     -v, --verbose                    Print more verbose output
+Jan 17 21:49:33 debian10-64.example.com dynflow-sidekiq@orchestrator[9064]:     -C, --config PATH                path to YAML config file
+Jan 17 21:49:33 debian10-64.example.com dynflow-sidekiq@orchestrator[9064]:     -L, --logfile PATH               path to writable logfile
+Jan 17 21:49:33 debian10-64.example.com dynflow-sidekiq@orchestrator[9064]:     -P, --pidfile PATH               path to pidfile
+Jan 17 21:49:33 debian10-64.example.com dynflow-sidekiq@orchestrator[9064]:     -V, --version                    Print version and exit
+Jan 17 21:49:33 debian10-64.example.com dynflow-sidekiq@orchestrator[9064]:     -h, --help                       Show help
```
It looks like the systemd unit is somehow wrong

ekohl:
https://github.com/theforeman/foreman-packaging/pull/4620 intends to fix the Debian packaging side.

ehelms:
Looks like I am seeing two failures, but latest nightly has the same bats failures. So I am good from that perspective. We will need this to land close to the foreman-maintain change + a build of foreman-maintain or else nightly will fail with the status check.

adamruzicka:
I updated the f-m PR to have the correct redis service names

ekohl:
https://ci.theforeman.org/job/foreman-develop-package-release/343/ should include the Debian fix. Once complete I'll kick off https://ci.theforeman.org/job/foreman-nightly-deb-pipeline/ as well to get green tests.

> I updated the f-m PR to have the correct redis service names

This is https://github.com/theforeman/foreman_maintain/pull/292

ekohl:
https://ci.theforeman.org/job/foreman-nightly-deb-pipeline/268/ should make that available.

ekohl:
I think the same statsd failure is seen in master now, haven't looked into it yet.

mmoll:
Exactly :green_heart:

mmoll:
merged, díky @ezr-ondrej!

PR #762, opened 2019-10-16
Add options for rails_cache_store

dgoetz:
This allows to make use of redis as backend, but still defaults to file.

@timogoebel: Can you have a look as the feature was implemented by you and so you perhaps know best if the parameters make sense in this way?

### DIFF
--- a/README.md
+++ b/README.md
@@ -33,6 +33,35 @@ on the host this module is applied to. Databases will be created with using the
 `en_US.utf8` locale, which means a respective OS locale must be available on
 the database host. The database management can be disabled with `db_manage`.
 
+## Rails Cache support
+
+Foreman supports different backends as Rails cache. This is handled by this
+module using the parameter `rails_cache_store`. The parameter takes a hash
+containing the type and options specfic to the backend.
+
+The default is the file backend, configured via `{'type' => 'file'}`. To
+setup for redis use a hash similar to `{'type' => 'redis', 'urls' => ['localhost:8479/0'], 'options' => {'compress' => 'true', 'namespace' => 'foreman'}}`
+where `urls` takes an array of redis urls which get prepended with `redis://`
+and `options` using a hash with options from [rails](https://guides.rubyonrails.org/caching_with_rails.html#activesupport-cache-store)
+falling back to `{'compress' => 'true', 'namespace' => 'foreman'}` if no
+option is provided.
+
+An example configuration for activating the redis backend with a local instance
+could look like this:
+
+```puppet
+class { 'foreman':
+  rails_cache_store => {
+    'type' => 'redis',
+    'urls' => ['localhost:8479/0'],
+    'options' => {
+      'compress' => 'true',
+      'namespace' => 'foreman'
+    }
+  }
+}
+```
+
 ## Support policy
 
 At any time, the module supports two releases, however the previous version
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -196,6 +196,8 @@
 #
 # $foreman_service_puma_workers::         Number of workers for Puma. Relevant only when Puma service is used and ignored when Passenger is used.
 #
+# $rails_cache_store::            Set rails cache store
+#
 # === Keycloak parameters:
 #
 # $keycloak::                     Enable Keycloak support. Note this is limited
@@ -301,6 +303,7 @@ class foreman (
   Integer[0] $foreman_service_puma_threads_min = $foreman::params::foreman_service_puma_threads_min,
   Integer[0] $foreman_service_puma_threads_max = $foreman::params::foreman_service_puma_threads_max,
   Integer[0] $foreman_service_puma_workers = $foreman::params::foreman_service_puma_workers,
+  Hash[String, Any] $rails_cache_store = $foreman::params::rails_cache_store,
   Boolean $keycloak = $foreman::params::keycloak,
   String[1] $keycloak_app_name = $foreman::params::keycloak_app_name,
   String[1] $keycloak_realm = $foreman::params::keycloak_realm,
--- a/manifests/install.pp
+++ b/manifests/install.pp
@@ -47,4 +47,10 @@ class foreman::install {
       ensure => installed,
     }
   }
+
+  if $foreman::rails_cache_store['type'] == 'redis' {
+    package { 'foreman-redis':
+      ensure => installed,
+    }
+  }
 }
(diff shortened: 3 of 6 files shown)


dgoetz:
I think the failed tests are unrelated, because it is just Debian failing for "install foreman with prometheus" which runs fine on CentOS.

dgoetz:
@ekohl I have it included now (also in the tests). Tests are still running.

manifests/init.pp
```diff
@@ -307,6 +310,7 @@
   Boolean $jobs_service_enable = $::foreman::params::jobs_service_enable,
   Boolean $hsts_enabled = $::foreman::params::hsts_enabled,
   Array[Stdlib::HTTPUrl] $cors_domains = $::foreman::params::cors_domains,
+  Hash $rails_cache_store = $::foreman::params::rails_cache_store,
```
ehelms: Are the options for cache store well defined? I know from user issue reports that hashes are seen as complex and painful to send on the commandline when configuring through something like `foreman-installer`. If possible, breaking this out would be nice.

dgoetz: I fear it is too flexible. Minimum are the following lines in the config which could also be omitted:
```
:rails_cache_store:
  :type: file
```
For redis it would be something like this as minimum:
```
:rails_cache_store:
  :type: redis
  :urls:
     - redis://localhost:8479/0
  :options:
    :compress: true
    :namespace: foreman
```
But it could be multiple URLs and more options.

In Puppet code it would look like:
```
{'type' => 'file'}
```
or:
```
{'type' => 'redis', 'urls' => ['localhost:8479/0'], options => {'compress' => 'true', 'namespace' => 'foreman'}}
```

Not sure how it would look for the installer. I am also not sure if future will perhaps bring more cache store backends and they will have a different format. 

With this in mind I do not come to a good alternative, best other way I can think of:
- rails_cache_store_type as string
- rails_cache_store_redis_urls as array of strings (only used when rails_cache_store_type equals redis)
- rails_cache_store_redis_options as array of strings which will be split on a delimiter (only used when rails_cache_store_type equals redis)

manifests/init.pp
```diff
@@ -208,6 +208,9 @@
 #
 # $cors_domains::                 List of domains that show be allowed for Cross-Origin Resource Sharing. This requires Foreman 1.22+
 #
+# $rails_cache_store::            Set rails cache store. This requires Foreman 1.22+. Defaults to {'type' => 'file'}.
+#                                 For redis use {'type' => 'redis', 'urls' => ['localhost:8479/0'], options => {'compress' => 'true', 'namespace' => 'foreman'}}.
```
ekohl: In another issue we've had the problem that this is confusing for foreman-installer users. How about an example in README?

dgoetz: I added a paragraph to the README. Was this the one you wanted or did you mean the one of foreman-installer?

laugmanuel:
@dgoetz can I support you in any way on getting this merged? I would love to experiment with this in our release environment.

dgoetz:
@laugmanuel: No, it needs re-review and comments from @ekohl and @ehelms, so I get input for further adjustments if required.

dgoetz:
Rebased to resolve merge conflicts, tests failing on debian for statsd seem to be unrelated.

manifests/init.pp
ekohl: I think it can be dropped here now. You can also drop the 1.22+ part since the module is 2.0.0 only now.

manifests/install.pp
```diff
@@ -48,4 +48,10 @@
       ensure => installed,
     }
   }
+
+  if $::foreman::rails_cache_store['type'] == 'redis' {
```
ekohl: If type is not present in the hash, I think this will fail and the type doesn't enforce it currently.

manifests/params.pp
```diff
@@ -188,6 +188,14 @@
   $logging_layout = 'pattern'
   $loggers = {}
 
+  # Rails Cache Store
+  $rails_cache_store = { 'type' => 'file' }
+
+  # KeepAlive settings of Apache
```
ekohl: This looks like a result of a rebase with conflicts and can be dropped.

README.md
```diff
@@ -34,6 +34,19 @@ on the host this module is applied to. Databases will be created with using the
...
 the database host. The database management can be disabled with `db_manage`.
 
+## Rails Cache support
+
+Foreman supports different backends as Rails cache. This is handled by this
+module using the parameter `rails_cache_store`. The parameter takes a hash
+containing the type and options specfic to the backend.
+
+The default is the file backend, configured via `{'type' => 'file'}`. To
+setup for redis use a hash similar to `{'type' => 'redis', 'urls' => ['localhost:8479/0'], options => {'compress' => 'true', 'namespace' => 'foreman'}}`
```
ekohl: Would it make sense to add an example like:
```puppet
class { 'puppet':
  rails_cache_store => { 'type' => 'file' },
}
```

And also the longer example.

dgoetz:
I rebased and addressed the comments.
* I removed the additional lines from the rebase.
* An example was added.
* I tested how a hash missing a key is handled and it works fine (according to the puppet docs it is simple handled as undef).

ehelms:
@dgoetz This will need another rebase

dgoetz:
Rebased and tests are green except from beaker tests for puppet 5 testing journald, statsd and prometheus, not sure if this is related.

manifests/init.pp
```diff
@@ -301,6 +303,7 @@
   Integer[0] $foreman_service_puma_threads_min = $foreman::params::foreman_service_puma_threads_min,
   Integer[0] $foreman_service_puma_threads_max = $foreman::params::foreman_service_puma_threads_max,
   Integer[0] $foreman_service_puma_workers = $foreman::params::foreman_service_puma_workers,
+  Hash $rails_cache_store = $foreman::params::rails_cache_store,
```
ekohl, line 306:
```suggestion
  Hash[String, Any] $rails_cache_store = $foreman::params::rails_cache_store,
```

dgoetz: Adjusted and pushed, tests are running again

manifests/init.pp
ekohl: Could you have a look at this?

dgoetz: Shortened to "Set rails cache store" only

dgoetz:
Tests are green now, and I hope I have addressed everything.

ekohl:
Thanks!

PR #763, opened 2019-10-24
Release 13.0.0

ekohl:


ekohl:
Looks like it fails on the locale but nothing changed there so I'm going to ignore that for now. A quick look in the vagrant image suggests that it no longer generates the `en_US.UTF-8` locale by default.

PR #764, opened 2019-10-29
Fixes #28146 - Drop double leading slash from cockpit url

adamruzicka:


ekohl:
This may require tests to be updated as well. Will wait for Travis to finish.

adamruzicka:
Test failures seem unrelated

PR #767, opened 2019-11-04
Initial AzureRM support

apuntamb:
In reference with [foreman-installer/405](https://github.com/theforeman/foreman-installer/pull/405).


apuntamb:
@tbrisker any update?

tbrisker:
the build seems to still be failing, i know @ekohl was looking into it as it isn't related to this PR but i don't know if there's been any progress (or if we can merge this with red build)

apuntamb:
Sure @tbrisker .
@ekohl Could you please take a look?

evgeni:
yeah, 1.24 only, we didn't backport the packages further

PR #770, opened 2019-11-12
Update GHCG gem and drop rdoc

ekohl:


mmoll:
merged, bedankt @ekohl!

PR #772, opened 2019-11-21
Add Foreman AzureRM cli option

apuntamb:


apuntamb:
@ekohl Build finished with RED again. Not sure if tests failure in previous PR continues here or if they are related to this PR.

manifests/cli/azure.pp
```diff
@@ -0,0 +1,10 @@
+# = Hammer AzureRM plugin
```
ekohl: Can you simplify the whole docs to use `puppet-strings` style?
```puppet
# @summary This installs the AzureRM plugin for Hammer CLI
```

apuntamb: Done. Removed the `=== Parameters` section. Since there were no separate params to mention with `@param`.
Does that make sense?

apuntamb:
> Can you add it [here](https://github.com/theforeman/puppet-foreman/blob/2a34012e18320ddd31e31b1e7a00cf449a70c7a6/spec/acceptance/foreman_cli_plugins_spec.rb#L30-L36) as well? That'll at least ensure it can be installed.
> 
> I've purged the CDN cache so the Red Hat ones should now at least pass.

done.

manifests/cli/azure.pp
```diff
@@ -0,0 +1,8 @@
+# @summary This installs the AzureRM plugin for Hammer CLI
+#
```
ekohl: Can you remove this "empty" line and the one below? Leaving one is fine.

spec/acceptance/foreman_cli_plugins_spec.rb
```diff
@@ -30,6 +30,7 @@ class { '::foreman::cli':
     if $facts['osfamily'] == 'RedHat' {
       include ::foreman::cli::ansible
     }
+    include ::foreman::cli::azure
```
ekohl: This is only packaged in RPMs so belongs in the `if` block above.

apuntamb:
> Will let the tests continue before merging.

Sure.

apuntamb:
rpm tests are green now.

ekohl:
Thanks!

PR #776, opened 2019-11-26
Add check for file to migrate and seed database

ehelms:
This is working to replace RPMs running migration steps by driving the database migration and seed via  the existence of files. There are two approaches that could be taken:

 1) Create a file when migration or seed occurs, have users or packages remove those files to signal migration/seed is needed
 2) Have users or packages create files to signal a migration/seed is needed and have the intsaller only run these steps if those files exist and then remove them

I went back and forth and landed on a more installer centric view via #1. I perceived benefits were:

 * the installer just works initially since it does not rely on something creating a file first
 * this allows some tracking of the last time migrations or seeds were ran via the file timestamp
 * packages can remove the files to signal the need for installation

I'm open to option #2 as an alternative this just puts a burden on external creation of these files for the initial installer run (which can reasonably be expected but feels bad).

### DIFF
--- a/manifests/database.pp
+++ b/manifests/database.pp
@@ -25,13 +25,28 @@ class foreman::database {
       value => false,
       dry   => true,
     }
-    ~> foreman::rake { 'db:migrate': }
+    ~> foreman::rake { 'db:migrate':
+      unless => '/usr/bin/test -f /etc/foreman/database_migrated',
+    } ~>
+    file { '/etc/foreman/database_migrated':
+      ensure  => 'present',
+      replace => 'no',
+      content => Timestamp.new().strftime('%Y-%m-%dT%H:%M:%S%:z'),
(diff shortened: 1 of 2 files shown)


manifests/database.pp
```diff
@@ -25,13 +25,28 @@
...
+      ensure  => 'present',
+      replace => 'no',
+      content => Timestamp.new().strftime('%Y-%m-%dT%H:%M:%S%:z'),
+      mode    => '0644',
+    }
     ~> foreman_config_entry { 'db_pending_seed':
       value => false,
       dry   => true,
+    } ~>
+    file { '/etc/foreman/database_seeded':
```
ekohl: You should do this after seeding.

ekohl:
Given #778, should we close this out?

PR #778, opened 2019-12-02
Run migrations if there are pending migrations

ehelms:
This drops relying on a settings entry in favor of lettings Rails
tell us when there are migrations that need to be applied.

### DIFF
--- a/manifests/database.pp
+++ b/manifests/database.pp
@@ -6,7 +6,7 @@ class foreman::database {
     contain $db_class
 
     if $::foreman::db_manage_rake {
-      Class[$db_class] ~> Foreman_config_entry['db_pending_migration']
+      Class[$db_class] ~> Foreman::Rake['db:migrate']
     }
   }
 
@@ -21,11 +21,9 @@ class foreman::database {
       'SEED_LOCATION'         => $::foreman::initial_location,
     }
 
-    foreman_config_entry { 'db_pending_migration':
-      value => false,
-      dry   => true,
+    foreman::rake { 'db:migrate':
+      unless => '/usr/sbin/foreman-rake db:abort_if_pending_migrations',
     }
-    ~> foreman::rake { 'db:migrate': }
     ~> foreman_config_entry { 'db_pending_seed':
       value => false,
       dry   => true,
--- a/manifests/rake.pp
+++ b/manifests/rake.pp
@@ -4,6 +4,7 @@ define foreman::rake(
   $timeout  = undef,
   $user     = $::foreman::user,
   $app_root = $::foreman::app_root,
+  $unless   = undef,
 ) {
   # https://github.com/rodjek/puppet-lint/issues/327
   # lint:ignore:arrow_alignment
@@ -12,8 +13,9 @@ define foreman::rake(
     user        => $user,
     environment => sort(join_keys_to_values(merge({'HOME' => $app_root}, $environment), '=')),
     logoutput   => 'on_failure',
-    refreshonly => true,
+    refreshonly => $unless =~ Undef,
     timeout     => $timeout,
+    unless      => $unless,
   }
   # lint:endignore
 }
--- a/spec/defines/foreman_rake_spec.rb
+++ b/spec/defines/foreman_rake_spec.rb
@@ -8,55 +8,44 @@ describe 'foreman::rake' do
       on_supported_os['redhat-7-x86_64']
     end
 
-    context 'without parameters' do
-      # These parameters are inherited normally, but here we cheat for performance
-      let :params do
-        {
-          user: 'foreman',
-          app_root: '/usr/share/foreman'
-        }
-      end
+    # These parameters are inherited normally, but here we cheat for performance
+    let :params do
+      {
+        user: 'foreman',
+        app_root: '/usr/share/foreman'
+      }
+    end
 
-      it {
+    context 'without parameters' do
+      it do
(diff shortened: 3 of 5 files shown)


ehelms:
This is an alternative solution to (https://github.com/theforeman/puppet-foreman/pull/776) for at least handling the migrate part of the database.

manifests/database.pp
```diff
@@ -21,11 +21,9 @@
       'SEED_LOCATION'         => $::foreman::initial_location,
     }
 
-    foreman_config_entry { 'db_pending_migration':
-      value => false,
-      dry   => true,
+    foreman::rake { 'db:migrate':
+      unless => "/usr/sbin/foreman-rake db:abort_if_pending_migrations"
```
ekohl: Linting will fail:

```suggestion
      unless => '/usr/sbin/foreman-rake db:abort_if_pending_migrations',
```

ehelms:
I went that route originally -- assume if migration that seed should happen. And had the same thought, seed doesn't require a migration. The seeds are tougher because by design they ought to be idempotent but I think your desire is that the exec is idempotent. So rather than let `db:seed` determine if changes are necessary, you prefer the exec determine that before running the `db:seed`. The current logic we live by is if Foreman or a plugin updates, try to db:seed. If the updates were all controlled inside the installer we could use that to notify the db:seed exec. However, we can't control that. I think that leaves us with using a file for seeds unfortunately if the desire is to avoid running db:seed on each puppet run.

manifests/rake.pp
```diff
@@ -14,6 +15,7 @@
     logoutput   => 'on_failure',
     refreshonly => true,
```
ekohl: This needs to be `false` when unless is set. Otherwise the whole trick still doesn't work.

ehelms:
A plugin adding notifications is an example of seed with no migration, e.g. https://github.com/theforeman/foreman_ansible/commit/d37ad23f88e2f1f93e36da977bb66a1b4664de4a#diff-90c017e7983b50d96418476d81b442db

ekohl:
You still need to change the `refreshonly` code.



> A plugin adding notifications is an example of seed with no migration, e.g. [theforeman/foreman_ansible@d37ad23#diff-90c017e7983b50d96418476d81b442db](https://github.com/theforeman/foreman_ansible/commit/d37ad23f88e2f1f93e36da977bb66a1b4664de4a#diff-90c017e7983b50d96418476d81b442db)

Without this change that isn't covered either. Let's get this in and think about seeding in the next stage.

manifests/rake.pp
```diff
@@ -4,16 +4,24 @@
   $timeout  = undef,
   $user     = $::foreman::user,
   $app_root = $::foreman::app_root,
+  $unless   = undef,
 ) {
+  if $unless == undef {
+    $refreshonly = true
+  } else {
+    $refreshonly = false
+  }
```
ehelms: @ekohl does this make sense?

spec/defines/foreman_rake_spec.rb
```diff
@@ -23,7 +23,7 @@
           .with_user('foreman')
           .with_environment(['HOME=/usr/share/foreman'])
           .with_logoutput('on_failure')
-          .with_refreshonly(true)
+          .with_refreshonly(false)
```
ekohl: This doesn't look correct because unless is not specified. In https://github.com/ekohl/puppet-foreman/pull/new/fixes I've written a test case for this.

manifests/rake.pp
```diff
@@ -4,16 +4,24 @@
...
+
   # https://github.com/rodjek/puppet-lint/issues/327
   # lint:ignore:arrow_alignment
   exec { "foreman-rake-${title}":
     command     => "/usr/sbin/foreman-rake ${title}",
     user        => $user,
     environment => sort(join_keys_to_values(merge({'HOME' => $app_root}, $environment), '=')),
     logoutput   => 'on_failure',
-    refreshonly => true,
+    refreshonly => $refreshonly,
```
ekohl: Short version:
```suggestion
    refreshonly => $unless =~ Undef,
```

spec/defines/foreman_rake_spec.rb
```diff
@@ -8,63 +8,66 @@
...
+      it do
         should contain_exec('foreman-rake-db:migrate')
           .with_command('/usr/sbin/foreman-rake db:migrate')
           .with_user('foreman')
           .with_environment(['HOME=/usr/share/foreman'])
           .with_timeout(60)
           .with_logoutput('on_failure')
-          .with_refreshonly(true)
-      }
+          .with_refreshonly(false)
```
ekohl:
```suggestion
          .with_refreshonly(true)
```

ehelms:
Looks like further debugging needed as its failing for CentOS 7.

ekohl:
Are you reading what I'm writing?
> Also odd that prometheus is failing. I already disabled that on Debian but now it's failing on centos as well. Perhaps that really is broken.

Since then it also looks like master is showing the same issue. That means I'd be ok with just merging this. Alternatively first merging a PR that disables the test similar to https://github.com/theforeman/puppet-foreman/commit/c0b961134552c01efe58d44b4a0839d48fe2b2ed

ehelms:
Sorry, I was. The output just made me think it was a regular curl status check that was failing. I do see now that the test description mentions prometheus.

PR #779, opened 2019-12-05
Fixes #28436 - Add keycloak support

ekohl:
This is a draft since it hasn't been tested end to end. It's here to allow end-to-end testing using forklift.

https://github.com/theforeman/forklift/pull/1065 is a related PR to install a Keycloak server.


manifests/settings.pp
```diff
@@ -41,4 +42,10 @@
       value => $email_smtp_password,
     }
   }
+
```
rahulbajaj0509: From what I understand, this condition would check if the setting `authorize_login_delegation_auth_source_user_autocreate` is set to `External`. So would it imply that the user needs to set this setting before running the installer?

manifests/config/apache.pp
```diff
@@ -200,6 +203,28 @@
...
+      package => 'mod_auth_openidc',
+    }
+
+
+    # This file is generated by keycloak-httpd-client-install and that manages
+    # the content. The command would be:
+    #
+    # keycloak-httpd-client-install --app-name ${keycloak_app_name} --keycloak-server-url $KEYCLOAK_URL --keycloak-admin-username $KEYCLOAK_USER --keycloak-realm ${keycloak_realm} --keycloak-admin-realm master --keycloak-auth-role root-admin --client-type openidc --client-hostname ${servername} --protected-locations /users/extlogin
+    #
+    # If $suburi is used, --location-root should also be passed in
```
rahulbajaj0509: Not sure what `$suburi` is and where/how it can we used. Can you share some insights?

manifests/settings.pp
ekohl: No, it sets the setting.

manifests/config/apache.pp
ekohl: Not sure how much it's used, but if `foreman_url` is passed in as `https://foreman.example.com/foreman` then `$suburi` will be set. Probably ok to ignore it for now.

manifests/settings.pp
```diff
@@ -41,4 +42,10 @@
...
+  if $keycloak {
+    foreman_config_entry { 'authorize_login_delegation_auth_source_user_autocreate':
+      value =>  'External',
+    }
+  }
```
rahulbajaj0509: This will not be a part of the PR anymore, am I correct @ekohl ?

ekohl: Yes, it was discussed to drop it.

ekohl:
Updated to not configure the settings and require the user to do so.

manifests/config/apache.pp
```diff
@@ -200,6 +203,28 @@
     include ::apache::mod::intercept_form_submit
     include ::apache::mod::lookup_identity
     include ::apache::mod::auth_kerb
+  } elsif $keycloak {
+    apache::mod { 'auth_openidc':
+      package => 'mod_auth_openidc',
+    }
```
rahulbajaj0509: Should we also install keycloak-httpd-client-install package too?

ekohl: I was debating that, but decided against it. I think the instruction should be:
* `yum install keycloak-httpd-client-install`
* `keycloak-httpd-client-install ....`
* `foreman-installer --foreman-keycloak true --foreman-keycloak-realm ...`

In theory the installer could run the install command, but then we need to pass credentials and I think that's something we probably don't want.

ntkathole:
@ekohl can you please rebase? 
```
stderr_lines:
  - From https://github.com/theforeman/puppet-foreman
  - ' * [new ref]         refs/pull/779/head -> pr'
  stdout: |-
    Auto-merging spec/classes/foreman_spec.rb
    Auto-merging manifests/init.pp
    Auto-merging manifests/config/apache.pp
    CONFLICT (content): Merge conflict in manifests/config/apache.pp
    Automatic merge failed; fix conflicts and then commit the result.
  stdout_lines: <omitted>
```

rahulbajaj0509:
@ekohl 

I think this should good once the conflict is solved, can you please solve the conflicts and rebase?

ekohl:
Rebased to fix the merge conflict.

ntkathole:
@ekohl @rahulbajaj0509 I am not able to test this PR as installer is failing with issue mentioned at https://github.com/theforeman/puppet-certs/pull/266#issuecomment-567159522 I tried to revert the PR changes but still installer is failing with `Failed to call refresh: Proxy centos7-katello-devel.nkathole.example.com cannot be retrieved: unknown error (response 503)`

ekohl:
> @ekohl @rahulbajaj0509 I am not able to test this PR as installer is failing with issue mentioned at [theforeman/puppet-certs#266 (comment)](https://github.com/theforeman/puppet-certs/pull/266#issuecomment-567159522) I tried to revert the PR changes but still installer is failing with `Failed to call refresh: Proxy centos7-katello-devel.nkathole.example.com cannot be retrieved: unknown error (response 503)`

That's unrelated, but should be resolved in the latest nightly packages.

ehelms:
@ekohl Seems fairly straight forward of a change, if you circle back to the tests I am happy to merge

manifests/config/apache.pp
```diff
@@ -83,6 +83,9 @@
   Boolean $ipa_authentication = $::foreman::ipa_authentication,
   Hash[String, Any] $http_vhost_options = {},
   Hash[String, Any] $https_vhost_options = {},
+  Boolean $keycloak = $foreman::keycloak,
+  String[1] $keycloak_app_name = $foreman::keycloak_app_name,
+  String[1] $keycloak_realm = $foreman::keycloak_realm,
```
ehelms: Do these need to be `$::` prefixed?

ekohl: Technically no and I'd also like to remove that prefix everywhere else. Just haven't gotten around to it. It was needed to make Puppet 3 and Puppet 4 behave in the same way but luckily those dark days are behind us. In this case I'm going to add them to stay consistent with the rest of the file.

ehelms:
Needs a rebase

spec/classes/foreman_config_apache_spec.rb
```diff
@@ -59,6 +59,15 @@
             should contain_file('/usr/share/foreman/config/environment.rb').with_owner('foreman')
           end
         end
+
+        describe 'with keycloak' do
+          let(:params) { super().merge(keycloak: true) }
+
+          it { should compile.with_all_deps }
+          it { should contain_apache_mod('auth_openidc') }
```
ehelms: This needs an additional underscore `apache__mod`

ehelms:
I see a dependency cycle error as well in the tests but I am less clear on how to resolve that one.

PR #780, opened 2019-12-13
Drop database seeding and let Foreman app handle

ehelms:
For lack of a better option, I went with creating a file on seed that can be removed by RPMs. This will allow RPMs to be simplified and still indicate when seeding is needed. Within the katello scenario, we can also drop this post upgrade task:

https://github.com/theforeman/foreman-installer/blob/develop/katello/hooks/post/30-upgrade.rb#L16

### DIFF
--- a/manifests/database.pp
+++ b/manifests/database.pp
@@ -22,13 +22,10 @@ class foreman::database {
     foreman::rake { 'db:migrate':
       unless => '/usr/sbin/foreman-rake db:abort_if_pending_migrations',
     }
-    ~> foreman_config_entry { 'db_pending_seed':
-      value => false,
-      dry   => true,
-    }
-    ~> foreman::rake { 'db:seed':
+    ~> Foreman::Rake['apipie:cache:index']
+
+    foreman::rake { 'db:seed':
       environment => delete_undef_values($seed_env),
     }
-    ~> Foreman::Rake['apipie:cache:index']
   }
 }


manifests/database.pp
```diff
@@ -24,12 +24,15 @@
     foreman::rake { 'db:migrate':
       unless => '/usr/sbin/foreman-rake db:abort_if_pending_migrations',
     }
-    ~> foreman_config_entry { 'db_pending_seed':
-      value => false,
-      dry   => true,
-    }
     ~> foreman::rake { 'db:seed':
       environment => delete_undef_values($seed_env),
+      unless      => '/usr/bin/test -f /etc/foreman/database_seeded',
```
ekohl: Puppet's exec has `creates` which natively implements this. Perhaps that should be exposed?

manifests/database.pp
```diff
@@ -24,12 +24,15 @@
...
+    }
+    ~> file { '/etc/foreman/database_seeded':
```
ekohl: I think that `/var/lib/foreman` is a better place for this.

manifests/database.pp
ehelms: I can go that route.  I was starting with not overloading the rake defines to see what you'd say before I went and added creates support :)

ekohl: I'm actually thinking about including the file creation in the rake task:

```puppet
define foreman::rake(
  Boolean $create_file = false,
) {
  exec { '...':
      # ...
  }

  if $create_file {
    file { "/var/lib/foreman/rake-${name}":
      ensure => file,
      # ...
      require => Exec['...'],
    }
  }
}

foreman::rake { 'db:seed':
  environment => delete_undef_values($seed_env),
  create_file => true,
}
```

On the other hand I'm also considering actually doing this in the seed job itself.

ekohl:
This removes the `db:migrate` implies `db:seed` logic. Can that create any problems?

manifests/database.pp
ehelms: Seed job itself? LIke a special rake task that runs seeds and creates the file? I considered that option as well.

ekohl: Yes, enhance the task. We already do so today: https://github.com/theforeman/foreman/blob/ae96cb70130968bdcc8b7a9a9af7edc7c6e0d9a8/lib/tasks/config.rake#L171-L184 (which needs to be replaced anyway if you're changing this in the installer)

ehelms:
Making a note that https://github.com/theforeman/foreman/blob/ad06a2d1361ea091e3849bd69831efcef469a320/lib/tasks/upgrade.rake#L15-L16 needs updating

manifests/database.pp
ehelms: Opened a PR to core here -- https://github.com/theforeman/foreman/pull/7262

ehelms:
Opened a PR to add the file creation in core: https://github.com/theforeman/foreman/pull/7262

manifests/database.pp
```diff
@@ -24,13 +26,14 @@
     foreman::rake { 'db:migrate':
       unless => '/usr/sbin/foreman-rake db:abort_if_pending_migrations',
     }
-    ~> foreman_config_entry { 'db_pending_seed':
-      value => false,
-      dry   => true,
+    ~> file { $database_seeded_file:
+      ensure  => 'absent',
```
ehelms: @ekohl thoughts on this logic to trigger seeding?

manifests/database.pp
```diff
@@ -24,13 +26,14 @@
...
     }
-    ~> foreman::rake { 'db:seed':
+    ~> Foreman::Rake['apipie:cache:index']
```
ehelms: Moved this here to trigger cache index on any migration

ehelms:
This is updated to reflect the new strategy in the linked Foreman core PR

ehelms:
Good catch. Would keeping the db:seed call but changing it to essentially only run once work?

ekohl:
Technically that's pretty much what the current implementation does. Given our priorities, we can keep this till last. First make sure all seeding happens at start up so we can drop it from packaging, then figure this out.

PR #783, opened 2019-12-19
Drop the selinux parameter

ekohl:
This was needed before we had foreman-selinux but that was introduced in Foreman 1.1. By now we can rely on only the selinux fact.


mmoll:
merged, bedankt @ekohl!

PR #784, opened 2019-12-19
Drop listen_on_interface

ekohl:
This parameter doesn't work well since you need to configure Apache slightly different as well since out of the box it doesn't work. Users who care about this functionality can use Hiera with the `http_vhost_options` and `https_vhost_options` parameters to set the ip directly.

### DIFF
--- a/manifests/config/apache.pp
+++ b/manifests/config/apache.pp
@@ -4,9 +4,6 @@
 #
 # $app_root::                 Root of the application.
 #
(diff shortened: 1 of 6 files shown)


mmoll:
tests fail :broken_heart:

ekohl:
Rebased to fix the merge conflict and updated the tests.

mmoll:
merged, bedankt @ekohl!

PR #785, opened 2019-12-19
Drop keepalive parameters

ekohl:
These settings can be set server wide and Apache already has sane defaults here. If needed, these can be tuned via Hiera using `http_vhost_options` and `https_vhost_options`.


mmoll:
merged, bedankt @ekohl!

PR #787, opened 2019-12-26
Refactor CLI plugin specs

ekohl:
By creating single files with cached facts reduces duplication and also speeds up execution.


mmoll:
Tests fail :broken_heart:

ekohl:
Reduced to just the CLI plugin testing. I realized the regular plugins already get the performance benefit of using cached facts via the support class so it's less important.

mmoll:
merged, bedankt @ekohl!

PR #789, opened 2020-01-23
Fixes #28848: Add dynflow worker with hosts_queue

ehelms:


manifests/service.pp
```diff
@@ -17,6 +17,12 @@
     }
 
     foreman::dynflow::worker { 'worker': }
+
+    if $jobs_hosts_queue_worker {
+      foreman::dynflow::worker { 'worker-hosts-queue':
+        queues => 'hosts_queue'
```
adamruzicka: If i'm reading [this](https://github.com/theforeman/puppet-foreman/blob/master/manifests/dynflow/worker.pp#L7) right, `queues` should be an array of strings

ehelms: Good catch, updated

ehelms:
Associated installer PR -- https://github.com/theforeman/foreman-installer/pull/457

manifests/service.pp
```diff
@@ -17,6 +18,12 @@
     }
 
     foreman::dynflow::worker { 'worker': }
+
+    if $jobs_worker_hosts_queue {
```
ekohl: I think this needs a way to disable the worker if it's false, otherwise it'll never be stopped.

manifests/service.pp
```diff
@@ -17,6 +18,12 @@
...
+      foreman::dynflow::worker { 'worker-hosts-queue':
+        queues => ['hosts_queue']
```
ekohl:
```suggestion
        queues => ['hosts_queue'],
```

ehelms:
Alternative implementation that deploys this no matter what for Katello -- https://github.com/theforeman/puppet-katello/pull/318

PR #790, opened 2020-01-23
Fixes #28955 - Added puma configuration

sthirugn:
~~I am creating this PR sooner with `DO NOT MERGE` tag to let everyone know how I am proceeding.~~
Any feedback is appreciated.

### About the configuration file puma_production.erb
- I have so far picked only 2 parameters - `threads` and `workers`. More will be picked in the future if need be. See [sample config](https://github.com/puma/puma/blob/f22dc6172235a57c56c28e3074c76ce421170b1c/examples/config.rb) for more info.

### Pending work in this PR
- [x] Externalize the parameters in the template (as environment variables?)
- [x] workers=4 is my test configuration. I need to test further to tune this to an appropriate default value.

### DIFF
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -193,6 +193,12 @@
 #
 # $cors_domains::                 List of domains that show be allowed for Cross-Origin Resource Sharing. This requires Foreman 1.22+
 #
+# $foreman_service_puma_threads_min::     Minimum number of threads for Puma. Relevant only when Puma service is used and ignored when Passenger is used.
+#
+# $foreman_service_puma_threads_max::     Maximum number of threads for Puma. Relevant only when Puma service is used and ignored when Passenger is used.
+#
+# $foreman_service_puma_workers::         Number of workers for Puma. Relevant only when Puma service is used and ignored when Passenger is used.
+#
 class foreman (
   Stdlib::HTTPUrl $foreman_url = $::foreman::params::foreman_url,
   Boolean $unattended = $::foreman::params::unattended,
@@ -286,6 +292,9 @@ class foreman (
   Optional[Redis::RedisUrl] $jobs_sidekiq_redis_url = $::foreman::params::jobs_sidekiq_redis_url,
   Boolean $hsts_enabled = $::foreman::params::hsts_enabled,
   Array[Stdlib::HTTPUrl] $cors_domains = $::foreman::params::cors_domains,
+  Integer[0] $foreman_service_puma_threads_min = $::foreman::params::foreman_service_puma_threads_min,
+  Integer[0] $foreman_service_puma_threads_max = $::foreman::params::foreman_service_puma_threads_max,
+  Integer[0] $foreman_service_puma_workers = $::foreman::params::foreman_service_puma_workers,
 ) inherits foreman::params {
   if $db_sslmode == 'UNSET' and $db_root_cert {
     $db_sslmode_real = 'verify-full'
--- a/manifests/params.pp
+++ b/manifests/params.pp
@@ -83,6 +83,9 @@ class foreman::params {
   $foreman_service_ensure = 'running'
   $foreman_service_enable = true
   $foreman_service_port = 3000
+  $foreman_service_puma_threads_min = 0
+  $foreman_service_puma_threads_max = 16
+  $foreman_service_puma_workers = 2
 
   # Define job processing service properties
   $jobs_manage_service = true
--- a/templates/foreman.service-overrides.erb
+++ b/templates/foreman.service-overrides.erb
@@ -6,3 +6,6 @@ Environment=FOREMAN_HOME=<%= scope['foreman::app_root'] %>
 Environment=FOREMAN_BIND=<%= scope['foreman::foreman_service_bind'] %>
 <% end -%>
 Environment=FOREMAN_PORT=<%= scope['foreman::foreman_service_port'] %>
+Environment=FOREMAN_PUMA_THREADS_MIN=<%= scope['foreman::foreman_service_puma_threads_min'] %>
+Environment=FOREMAN_PUMA_THREADS_MAX=<%= scope['foreman::foreman_service_puma_threads_max'] %>
+Environment=FOREMAN_PUMA_WORKERS=<%= scope['foreman::foreman_service_puma_workers'] %>


manifests/config.pp
```diff
@@ -93,6 +93,19 @@
     ensure  => absent,
   }
 
+  # adding puma configuration
+  file { "${::foreman::app_root}/config/puma":
+    ensure  => directory,
+  }
+
+  file { "${::foreman::app_root}/config/puma/production.rb":
```
ekohl: I think this should actually live in http://github.com/theforeman/foreman. Since it's Ruby, it can use `ENV` and we can set that in our systemd overrides. This way container deployments can also benefit from this.

sthirugn: Maybe like creating `puma/production.rb` in https://github.com/theforeman/foreman/tree/develop/config. If so, where will the template (templates/puma_production.erb) live?

ekohl: I think you can avoid to template it. In pure Ruby:
```ruby
# Configure "min" to be the minimum number of threads to use to answer
# requests and “max” the maximum.
#
# The default is "0, 16".
#
threads ENV.fetch('FOREMAN_PUMA_THREADS_MIN', 0).to_i, ENV.fetch('FOREMAN_PUMA_THREADS_MAX, 16).to_i


# === Cluster mode ===

# How many worker processes to run.
#
# The default is "0".
#
workers ENV.fetch('FOREMAN_PUMA_WORKERS', 4).to_i
```

Untested, but I suspect this will work. Then you should be able to set these environment variables via systemd overrides.

sthirugn: @ekohl I tried the configuration you gave above in `/usr/share/foreman/config/puma/production.rb`

I provided overrides in file `/usr/share/foreman-installer/modules/foreman/templates/foreman.service-overrides.erb` like the following:

```
[Service]
User=<%= scope['foreman::user'] %>
<% if scope['foreman::foreman_service_bind'] -%>
Environment=FOREMAN_BIND=<%= scope['foreman::foreman_service_bind'] %>
<% end -%>
Environment=FOREMAN_PORT=<%= scope['foreman::foreman_service_port'] %>
Environment=FOREMAN_PUMA_WORKERS=8
Environment=FOREMAN_PUMA_THREADS_MIN=4
Environment=FOREMAN_PUMA_THREADS_MAX=12
```

Initially I thought it was working, but later figured that it is not really picking up the environment variable values defined. I am looking into this further.

ehelms: Just adding notes from discussion, I tested this method out and it worked great for me. We will want a Redmine issue since we'll be adding a baseline configuration file to Foreman core codebase as well as this overrides file.

ehelms:
Something to consider is the db pool size we used given the threading nature of Puma. We likely want to tune this based on these settings.

https://devcenter.heroku.com/articles/concurrency-and-database-connections

sthirugn:
> You should also add the values to `manifests/init.pp`.

Thank you, @ekohl   I added params to init.pp and confirmed that I can override them through custom-hiera.yaml parameters like the following:
```sh
# cat /etc/foreman-installer/custom-hiera.yaml | tail -n 3
foreman::foreman_puma_threads_min: 1
foreman::foreman_puma_threads_max: 14
foreman::foreman_puma_workers: 6
```

ekohl:
It shouldn't be needed to specify it via hiera and should show up as `--foreman-foreman-puma-threads-min`.

manifests/init.pp
```diff
@@ -286,6 +292,9 @@
   Optional[Redis::RedisUrl] $jobs_sidekiq_redis_url = $::foreman::params::jobs_sidekiq_redis_url,
   Boolean $hsts_enabled = $::foreman::params::hsts_enabled,
   Array[Stdlib::HTTPUrl] $cors_domains = $::foreman::params::cors_domains,
+  Integer[0] $foreman_puma_threads_min = $::foreman::params::foreman_puma_threads_min,
```
ekohl: I'd group these together with the other service options to at least keep some logic to it even though it probably doesn't matter that much.

manifests/init.pp
```diff
@@ -193,6 +193,12 @@
 #
 # $cors_domains::                 List of domains that show be allowed for Cross-Origin Resource Sharing. This requires Foreman 1.22+
 #
+# $foreman_puma_threads_min::     Minimum number of threads for puma
```
ekohl: Would it make sense to say this is only relevant when the Puma service is used but ignored when passenger is used?

ehelms: I also don't think we need the `foreman_` in front. These are already scoped by the puppet module itself.

ehelms: The related scoping is because there is a service and its named foreman so `foreman_service_` makes sense. If you want to tie this specifically to that, I would namespace them as `foreman_service_puma_threads_min` but I don't feel that its required to namespace them.

sthirugn: > The related scoping is because there is a service and its named foreman so `foreman_service_` makes sense. If you want to tie this specifically to that, I would namespace them as `foreman_service_puma_threads_min` but I don't feel that its required to namespace them.
I am going to use `foreman_service_` just to match with the other foreman service params. It will be easy to debug/refer these params in future.

manifests/init.pp
sthirugn: Do you mean `foreman_service`, `foreman_service_ensure`, `foreman_service_enable`, `foreman_service_port`? I don't see them in init.pp. Should I add those?

sthirugn:
> It shouldn't be needed to specify it via hiera and should show up as `--foreman-foreman-puma-threads-min`.

Good to know this.

sthirugn:
> Something to consider is the db pool size we used given the threading nature of Puma. We likely want to tune this based on these settings.
> 
> https://devcenter.heroku.com/articles/concurrency-and-database-connections

I am testing with different configurations. I will post here when I have more info.

sthirugn:
Updated default puma workers to 2 as per https://github.com/theforeman/foreman/pull/7421#discussion_r378565214

manifests/init.pp
```diff
@@ -193,6 +193,12 @@
 #
 # $cors_domains::                 List of domains that show be allowed for Cross-Origin Resource Sharing. This requires Foreman 1.22+
 #
+# $foreman_service_puma_threads_min::     Minimum number of threads for puma. Relevant only when Puma service is used and ignored when Passenger is used.
```
ekohl: Bit of a bikeshedding: should this leave out `puma` in the variable name? The use of `puma` vs `Puma` is also a bit inconsistent.

```suggestion
# $foreman_service_threads_min::     Minimum number of threads for Puma. Relevant only when Puma service is used and ignored when Passenger is used.
```

sthirugn: I think calling out puma explicitly in the variable name will help the users under it as `puma threads` vs. just `foreman threads`.

manifests/init.pp
```diff
@@ -193,6 +193,12 @@
 #
 # $cors_domains::                 List of domains that show be allowed for Cross-Origin Resource Sharing. This requires Foreman 1.22+
 #
+# $foreman_service_puma_threads_min::     Minimum number of threads for Puma. Relevant only when Puma service is used and ignored when Passenger is used.
+#
+# $foreman_service_puma_threads_max::     Maximum number of threads for Puma. Relevant only when Puma service is used and ignored when Passenger is used.
+#
+# $foreman_service_puma_workers::         Number of workers for Puma. Relevant only when Puma service is used and ignored when Passenger is used.
+
```
ekohl: The empty line here breaks some tools. The comment block must be attached to the class.
```suggestion
#
```

manifests/init.pp
sthirugn: @ekohl Updated lowecase/uppercase inconsistencies, let me know what you think

manifests/init.pp
sthirugn: Good catch, fixed it now. thanks!

ekohl:
Tests are still running. The person merging this should make sure to squash the commits.

ehelms:
Same failures as master, mergin

ekohl:
@sthirugn the dash in the commit message appears to be needed for the PR processor. Now it wasn't checked. I also didn't check during the review, but let's be mindful about this in the future. Now it didn't hard fail because this repo doesn't hard require a Redmine issue.

sthirugn:
> @sthirugn the dash in the commit message appears to be needed for the PR processor. Now it wasn't checked. I also didn't check during the review, but let's be mindful about this in the future. Now it didn't hard fail because this repo doesn't hard require a Redmine issue.

ack thank you for pointing this out.

PR #791, opened 2020-01-24
Refs #28067: Ensure dynflow worker config exists before service

ehelms:
If the dynflow worker service starts before the configuration file
has been deployed the worker will fail to start.

### DIFF
--- a/manifests/dynflow/worker.pp
+++ b/manifests/dynflow/worker.pp
@@ -18,9 +18,9 @@ define foreman::dynflow::worker (
     mode    => '0644',
     content => template('foreman/dynflow_worker.yml.erb'),
   }
-
-  service { "dynflow-sidekiq@${service_name}":
-    ensure => $service_ensure,
-    enable => $service_enable,
+  ~> service { "dynflow-sidekiq@${service_name}":
+    ensure    => $service_ensure,
+    enable    => $service_enable,
+    subscribe => Class['foreman::database'],
   }
 }


ehelms:
Once we added the Katello hosts_queue worker this was found to break nightly.

manifests/dynflow/worker.pp
```diff
@@ -11,15 +11,15 @@
   Boolean $service_enable = $::foreman::jobs_service_enable
 ) {
 
+  Class['foreman::database::postgresql'] ~>
```
ehelms: Not sure if this is correct way to handle. The issue is that a worker can involve connecting to the database and therefore shouldn't be started until database setup has completed.

ehelms:
There is one other issue at the moment as well. If the worker boots, in Katello's case, the seeds will run per our new in app method. Those seeds can fail due to no Candlepin being up yet and crash the worker. We can:

 1) Add something to the seeds code to not run if its a dynflow worker assuming we can detect that
 2) Update katello seeds to fail gracefully
 3) Add a relationship to dynflow worker startup in puppet-katello to ensure Candlepin or App is started before hand

manifests/dynflow/worker.pp
```diff
@@ -11,15 +11,18 @@
   Boolean $service_enable = $::foreman::jobs_service_enable
 ) {
 
+  if $::foreman::db_manage {
+    Class['foreman::database::postgresql'] ~> Service["dynflow-sidekiq@${service_name}"]
```
ekohl: I think this chaining is better with `foreman::database` because then it starts after `db:migrate` and `db:seed` as well. It also has the conditional on `db_manage`. So you can just use `subscribe => Class['foreman::database::postgresql']` on the `service`

manifests/dynflow/worker.pp
```diff
@@ -18,9 +18,9 @@
...
     content => template('foreman/dynflow_worker.yml.erb'),
   }
-
-  service { "dynflow-sidekiq@${service_name}":
-    ensure => $service_ensure,
-    enable => $service_enable,
+  ~> service { "dynflow-sidekiq@${service_name}":
+    ensure    => $service_ensure,
+    enable    => $service_enable,
+    subscribe => Class['foreman::database'],
```
ehelms: Like this? @ekohl

ehelms:
Not obvious to me what the error here is

PR #800, opened 2020-02-14
Make foreman::config::apache standalone

ekohl:
By not relying on the foreman class parameters, it becomes easier to use this class without Foreman. This can be useful when deploying it using a git checkout rather than packages. An example of that is https://github.com/theforeman/puppet-katello_devel/pull/225

### DIFF
--- a/manifests/config.pp
+++ b/manifests/config.pp
@@ -94,6 +94,33 @@ class foreman::config {
   }
 
   if $::foreman::apache  {
+    class { 'foreman::config::apache':
+      passenger               => $::foreman::passenger,
+      app_root                => $::foreman::app_root,
+      passenger_ruby          => $::foreman::passenger_ruby,
+      priority                => $::foreman::vhost_priority,
+      servername              => $::foreman::servername,
+      serveraliases           => $::foreman::serveraliases,
+      server_port             => $::foreman::server_port,
+      server_ssl_port         => $::foreman::server_ssl_port,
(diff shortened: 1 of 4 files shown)


ehelms:
So far I am running into:

```
    [ERROR 2020-02-14T14:21:11 verbose]  Evaluation Error: Error while evaluating a Resource Statement, Duplicate declaration: Class[Foreman::Config::Apache] is already declared at (file: /usr/share/foreman-installer/modules/katello_devel/manifests/apache.pp, line: 6); cannot redeclare (file: /usr/share/foreman-installer/modules/foreman/manifests/config.pp, line: 97) (file: /usr/share/foreman-installer/modules/foreman/manifests/config.pp, line: 97, column: 5) on node centos7-devel.war.example.com
```

My guess is this is due to puppet-pulpcore integration (https://github.com/theforeman/puppet-foreman_proxy_content/blob/master/manifests/init.pp#L292).

ekohl:
I think that's actually a valid concern because that unconditionally pulls in Foreman from packages again just to get the servername. That could also `include foreman::config::apache` and then use `$foreman::config::apache::servername`.

ekohl:
https://github.com/theforeman/puppet-foreman_proxy_content/pull/243 would probably work there, but depends on this.

ehelms:
The change to puppet-katello is indeed needed to let the devel install pass. With all 3 of those changes though I got a successful install!

When you get a few moments, if you could file a Redmine issue, link all of them up and then I can pass through each with a review. This will unblock devel having Pulp 3 which I know a number of developers are needing.

Thanks for the re-factor here!

Issue #796, opened 2020-02-11
AD registration with GSS proxy fails


kuwv:
Installer fails to register satellite to AD due to bad default while querying sssd_services. This working in 6.4 and fails in 6.5 and 6.6. Here are the procedures:

https://access.redhat.com/documentation/en-us/red_hat_satellite/6.5/html/administering_red_hat_satellite/chap-red_hat_satellite-administering_red_hat_satellite-configuring_external_authentication#sect-Red_Hat_Satellite-Administering_Red_Hat_Satellite-Configuring_External_Authentication-Using_Active_Directory

```
Evaluation Error: Operator '[]' is not applicable to an Undef Value. (file: /usr/share/foreman-installer/modules/fore
man/manifests/config.pp, line: 136, column: 43) on node
```

kuwv:
Current versions are:
```
$ puppet --version
5.5.12
$ facter --version
3.11.8 (commit eb5f71136af5012f3a7169ed3a77a111c1e4d765)
```

Seems the facts might not be populating correctly due to: https://puppet.com/docs/puppet/6.10/custom_facts.html#concept-243

```
Note: Facter 3.0 removed the Ruby implementations of some features and replaced them with a custom facts API. Any custom fact that requires one of the Ruby files previously stored in lib/facter/util fails with an error.
```

kuwv:
Guess this isn't monitored. :(

ekohl:
I wonder if b0f88ed52539ca74cf58017d97bbb8327b083e1f is the cause of this. However, it should not fail with that error since it does check if the facts exist. I'm currently looking at writing some tests.

ekohl:
I spent some time on adding tests and could only get them to work with string keys in https://github.com/theforeman/puppet-foreman/pull/801. Would you mind testing if this is also a solution to your problem?

kuwv:
@ekohl My apologies, I hadn't seen that you commented. I will give it a check sometime this week.

ekohl:
Do you by any chance have the `ipa` module installed? That also provides the `ipa` fact and the conflict is an issue. I'll rename the fact in this module to `foreman_ipa` to avoid the naming issue.

PR #801, opened 2020-02-18
Use string keys for custom facts

ekohl:
While adding tests, this was the only way to get a result. When using symbols, it returned nil instead of a hash.

Possible solution for GH-796. If this is a fix, then this needs a Redmine issue before merging.

### DIFF
--- a/spec/unit/facter/ipa_spec.rb
+++ b/spec/unit/facter/ipa_spec.rb
@@ -1,0 +1,40 @@
+require 'spec_helper'
+
+describe 'ipa' do
+  subject { Facter.value(:ipa) }
+
+  before do
+    Facter.reset
+    Facter.loadfacts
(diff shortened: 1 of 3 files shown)


spec/unit/facter/ipa_spec.rb
```diff
@@ -0,0 +1,40 @@
+require 'spec_helper'
+
+describe 'ipa' do
+  subject { Facter.value(:ipa) }
+
+  before do
+    Facter.reset
+    Facter.loadfacts
+    allow(Facter::Util::Sssd).to receive(:ipa_value).with('global/realm').and_return(default_realm)
```
evgeni:
```
     NameError:
       uninitialized constant Facter::Util::Sssd
```

ekohl:
This was tested by a user and it doesn't work.

PR #804, opened 2020-02-28
Add instructions to run specific tests

sthirugn:


ehelms:
This will get wiped away if not added to https://github.com/theforeman/foreman-installer-modulesync/blob/master/moduleroot/CONTRIBUTING.md.erb which I think would be good for all modules to get.

sthirugn:
@ehelms thanks, closing this in favor of https://github.com/theforeman/foreman-installer-modulesync/pull/108

PR #805, opened 2020-03-02
Fixes #29244: Default to disabling SSLv3

ehelms:
This is mostly until EL5 EOLs.


ehelms:
Should I file an issue to drop this parameter from this module? My logic for the question is that if we shouldnt be doing this here, we probably should remove the ability to.

ehelms:
Replaced by https://github.com/theforeman/foreman-installer/pull/480

ekohl:
> Should I file an issue to drop this parameter from this module? My logic for the question is that if we shouldnt be doing this here, we probably should remove the ability to.

I was actually thinking about that. It was introduced in 0602750cf3e97a51e789de8d962a0b3316dd9709 but e40c8dced27f8c59c40009f0ce5806af71702c3b allows you to set this via hiera. Given this is not the primary way we should be setting this, I think it should be dropped but hadn't gotten around to it.

PR #807, opened 2020-03-03
Fixes #29255 - Set plugin config file mode to 0640

ekohl:
This sets the config file mode to 0640 by default because they may contain secrets such as credentials. To keep it readable for Foreman, the group is modified to $foreman::group.


ekohl:
Rebased and updated the tests.

PR #808, opened 2020-03-05
Bump version to 15.0.0

ekohl:
Various changes have gone in that make this a new major version.


ekohl:
Before merging this I'll also prepare PRs for other modules that depend on this so we don't break the installer nightly.

PR #809, opened 2020-03-05
Use plugin_prefix to determine plugin packages

ekohl:
Previously this was duplicating the plugin prefix for packages that didn't follow the convention. By using regsubst, this can be avoided. This removes the need for separate params classes.


ekohl:
Backwards incompatible because this drops the package parameter

mmoll:
merged, bedankt @ekohl!

PR #814, opened 2020-03-20
Refs #29144 - Use systemd socket activation

ekohl:
When using systemd socket activation, it's important that the ListenSocket matches what Puma binds on. Otherwise it may fail. This happens when is configured on [::]:3000 (dual stack) and Puma on 0.0.0.0:3000. Puma will then attempt to bind and fail because the port is already in use.

The service bind is now made explicit because systemd's ListenSocket=3000 binds on :: where Puma by default binds on 0.0.0.0:3000. This is IPv4-only, but is what was done prior to this as well. Apache is configured dual stack and the recommended deployment.

Note this depends on https://github.com/theforeman/foreman/pull/7536 and https://github.com/theforeman/foreman-packaging/pull/4894 (deb) / https://github.com/theforeman/foreman-packaging/pull/4895 (rpm).

### DIFF
--- a/manifests/config.pp
+++ b/manifests/config.pp
@@ -43,10 +43,22 @@ class foreman::config {
     ensure => absent,
   }
 
+  $listen_socket = $foreman::foreman_service_bind ? {
+    Stdlib::IP::Address::V6 => "[${foreman::foreman_service_bind}]:${foreman::foreman_service_port}",
+    default                 => "${foreman::foreman_service_bind}:${foreman::foreman_service_port}",
+  }
+
   if $::foreman::use_foreman_service {
-    systemd::dropin_file { 'installer.conf':
-      unit    => "${::foreman::foreman_service}.service",
-      content => template('foreman/foreman.service-overrides.erb'),
+    systemd::dropin_file { 'foreman-socket':
+      filename => 'installer.conf',
+      unit     => "${::foreman::foreman_service}.socket",
+      content  => template('foreman/foreman.socket-overrides.erb'),
+    }
+
+    systemd::dropin_file { 'foreman-service':
+      filename => 'installer.conf',
+      unit     => "${::foreman::foreman_service}.service",
+      content  => template('foreman/foreman.service-overrides.erb'),
     }
   }
 
@@ -103,7 +115,7 @@ class foreman::config {
       serveraliases           => $::foreman::serveraliases,
       server_port             => $::foreman::server_port,
       server_ssl_port         => $::foreman::server_ssl_port,
-      proxy_backend           => "http://${::foreman::foreman_service_bind}:${::foreman::foreman_service_port}/",
+      proxy_backend           => "http://${listen_socket}/",
       ssl                     => $::foreman::ssl,
       ssl_ca                  => $::foreman::server_ssl_ca,
       ssl_chain               => $::foreman::server_ssl_chain,
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -333,7 +333,7 @@ class foreman (
     $foreman_service_bind = '127.0.0.1'
   } else {
     $use_foreman_service = true
-    $foreman_service_bind = undef
+    $foreman_service_bind = '0.0.0.0'
   }
 
   include ::foreman::repo
--- a/spec/classes/foreman_spec.rb
+++ b/spec/classes/foreman_spec.rb
@@ -78,7 +78,8 @@ describe 'foreman' do
             .with_content(/adapter: postgresql/)
         end
 
-        it { should contain_systemd__dropin_file('installer.conf') }
+        it { should contain_systemd__dropin_file('foreman-socket') }
+        it { should contain_systemd__dropin_file('foreman-service') }
 
         it { should contain_file('/usr/share/foreman').with_ensure('directory') }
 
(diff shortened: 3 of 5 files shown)


manifests/config.pp
```diff
@@ -44,9 +44,21 @@
...
-      content => template('foreman/foreman.service-overrides.erb'),
+    $listen_socket = $foreman::foreman_service_bind ? {
+      Stdlib::IP::Address::V6 => "[${foreman::foreman_service_bind}]:${foreman::foreman_service_port}",
+      default                 => "${foreman::foreman_service_bind}:${foreman::foreman_service_port}",
+    }
+
+    systemd::dropin_file { 'foreman-socket':
+      filename => 'installer.conf',
+      unit     => "${::foreman::foreman_service}.socket",
+      content  => "[Service]\nListenSocket=\nListenSocket=${listen_socket}\n",
```
ehelms: This is harder to read than say a template or multiline content, but I think what I am seeing is this example output:

```
[Service]
ListenSocket=
ListenSocket=127.0.0.1:3000
```

If so, what is the empty ListenSocket doing?

manifests/config.pp
```diff
@@ -44,9 +44,21 @@
...
+    }
+
+    systemd::dropin_file { 'foreman-service':
+      filename => 'installer.conf',
```
ehelms: Was curious why we call this installer.conf ?

manifests/config.pp
ekohl: It clears the built in from `foreman.socket`, similar to how we use `ExecStart=`. Otherwise it's in addition to since `ListenSocket` can be repeated.

I first had it without the `ListenSocket=` and then it was short enough but now I agree it may be a bit less readable. Because it's short, I'll have a look at using heredoc instead.

manifests/config.pp
ekohl: We already did this before, but since you can't define `systemd::dropin_file` twice with 'installer.conf' I had to be explicit. This creates `/etc/systemd/system/foreman.service.d/installer.conf` and I thought this was the most logical name for it. `systemctl cat foreman` shows you how it would look like:

```text
# /usr/lib/systemd/system/foreman.service
[Unit]
Description=Foreman
Documentation=https://theforeman.org
After=network.target remote-fs.target nss-lookup.target
Requires=foreman.socket

[Service]
Type=simple
User=foreman
TimeoutSec=300
WorkingDirectory=/usr/share/foreman
ExecStart=/usr/share/foreman/bin/rails server --environment $FOREMAN_ENV --port $FOREMAN_PORT --binding $FOREMAN_BIND
Environment=FOREMAN_ENV=production FOREMAN_PORT=3000 FOREMAN_BIND=0.0.0.0

SyslogIdentifier=foreman

[Install]
WantedBy=multi-user.target

# /etc/systemd/system/foreman.service.d/installer.conf
[Service]
User=foreman
Environment=FOREMAN_ENV=production
Environment=FOREMAN_HOME=/usr/share/foreman
Environment=FOREMAN_BIND=127.0.0.1
Environment=FOREMAN_PORT=3000
Environment=FOREMAN_PUMA_THREADS_MIN=0
Environment=FOREMAN_PUMA_THREADS_MAX=16
Environment=FOREMAN_PUMA_WORKERS=2
```

manifests/config.pp
ekohl: And for completeness, `systemctl cat foreman.socket`

```text
# /etc/systemd/system/foreman.socket
[Unit]
Description=Puma HTTP Server Accept Sockets

[Socket]
ListenStream=3000

# AF_UNIX domain socket
# SocketUser, SocketGroup, etc. may be needed for Unix domain sockets
# ListenStream=/run/puma.sock

# Socket options matching Puma defaults
NoDelay=true
ReusePort=true
Backlog=1024

[Install]
WantedBy=sockets.target

# /etc/systemd/system/foreman.socket.d/installer.conf
[Service]
ListenSocket=
ListenSocket=127.0.0.1:3000
```

ekohl:
I've updated it to reuse `$listen_socket` in the Apache reverse proxy config. This makes it even clearer that these are directly linked. In a later iteration we can have a look at using unix sockets which can prevent attacks (currently anyone on the local host can send HTTP(S) headers impersonating hosts)

ehelms:
Looks like a linting error

ekohl:
https://github.com/voxpupuli/puppet-lint-unquoted_string-check/issues/4 is the lint issue.

ehelms:
Hrmm, can we work around it for now?

ekohl:
https://github.com/voxpupuli/puppet-lint-unquoted_string-check/pull/9 is the bugfix. Working on releasing a fix for it.

ekohl:
Restarted the failed tests now that the new version is out.

mmoll:
tests fail :broken_heart:

ekohl:
I wonder why https://travis-ci.org/github/theforeman/puppet-foreman/builds/666888226 is not in the status checks

ehelms:
Seems Travis might be having a rough time

ekohl:
:green_heart: now

ekohl:
I'm not entirely sure if this is breaking or not. I think it still works on Foreman 2.0 since we don't manage `foreman.socket`. However, perhaps we need to. Labeled as enhancement for now.

PR #815, opened 2020-03-25
Refactor repository handling

ekohl:
This removes the repo parameters from the main class, in favor of a standalone class that can be included. It uses an anchor because that can be collected in the main class to keep the correct dependency chaining while using composition.

Still a draft because I need to clean up some more things but now it's at least a backup.


manifests/repo.pp
```diff
@@ -1,22 +1,43 @@
...
+    if $configure_scl_repo {
+      Foreman::Repos['foreman'] -> Package['foreman-release-scl']
     }
   }
 
-  class { 'foreman::repos::extra':
-    configure_epel_repo => $configure_epel_repo,
-    configure_scl_repo  => $configure_scl_repo,
+  if $configure_scl_repo {
+    package {'foreman-release-scl':
```
ehelms, line 35: Was the intent to move this to centos-release-scl-rh instead?

ekohl: No. I first want to merge it this way and rely on it. I want to drop `foreman-release-scl` in a separate PR but with this merged, I can at least start to clean up acceptance tests in puppet-foreman_proxy which will help with EL8.

ehelms:
Gotcha. Does this need an associated installer migration to go with it?

ekohl:
I don't think so. We'll stop supporting managing the repositories since the parameters no longer exist, the installer should remove them. That said, it probably makes sense to track it in a Redmine issue which I'll create now.

ekohl:
> Installation of packages like `foreman-postgresql`, `rh-redis5-redis`, and `rh-ruby25-ruby` is failing in the CentOS7 acceptance test.

Incorrect fact. Pushing an update.

mmoll:
tests fail :broken_heart:

mmoll:
merged, bedankt @ekohl!

PR #822, opened 2020-04-08
Only install foreman-release-scl on CentOS EL 7

ehelms:


manifests/params.pp
```diff
@@ -113,11 +111,21 @@
         $passenger_ruby_package = 'tfm-rubygem-passenger-native'
         $plugin_prefix = 'tfm-rubygem-foreman_'
       }
+
+      case $::operatingsystem {
+        'CentOS': {
+          $configure_scl_repo = (versioncmp($facts['operatingsystemmajrelease'], '7') >= 0)
```
evgeni, line 117: shouldn't this be `== 0` as otherwise it will also install it on EL8?

evgeni:
test failures on ScientificLinux and RHEL related.

ehelms:
Updated and simplified. I incorrectly thought RHEL required RH SCL.

PR #826, opened 2020-04-17
Drop the separate rails repository

ekohl:
Foreman 2.0 dropped this and needs to be removed.


ekohl:
Updated the tests. Ubuntu acceptance tests failing on journald, but I suspect that dynflow took just too long to start up. Essentially a race condition and not related to the files I changed.

mmoll:
merged, bedankt @ekohl!

PR #827, opened 2020-04-17
Clean up acceptance test with shared examples

ekohl:
This is taking a part of https://github.com/theforeman/puppet-foreman/pull/699 but leaves out the beaker-hiera changes and shared context to get it in a mergeable state.


ekohl:
Updated, copied the shared example from a branch that also included another change already (moving the certs to `/etc/foreman-certs`). Should pass now.

mmoll:
merged, bedankt @ekohl!

PR #828, opened 2020-04-17
Fixes #29212 - support el8

wbclark:
This is a WIP which incorporates the PRs #826 & #827 .

### DIFF
--- a/spec/acceptance/foreman_cli_plugins_spec.rb
+++ b/spec/acceptance/foreman_cli_plugins_spec.rb
@@ -11,7 +11,7 @@ describe 'Scenario: install foreman-cli + plugins without foreman' do
       password    => 'changeme',
     }
 
-    if $facts['osfamily'] == 'RedHat' {
+    if $facts['os']['family'] == 'RedHat' {
       include foreman::cli::ansible
       include foreman::cli::azure
     }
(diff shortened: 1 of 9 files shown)


wbclark:
When I test locally with EL7 it's failing due to the known issue: https://projects.theforeman.org/issues/29429

With EL8, an error occurs  when running puppet apply: `Error: Could not find a suitable provider for foreman_config_entry`

mmoll:
@wbclark please rebase :)

spec/acceptance/foreman_cli_plugins_spec.rb
```diff
@@ -11,9 +11,9 @@ class { 'foreman::cli':
       password    => 'changeme',
     }
 
-    if $facts['osfamily'] == 'RedHat' {
-      include foreman::cli::ansible
-      include foreman::cli::azure
+    if $facts['os']['family'] == 'RedHat' {
+      include ::foreman::cli::ansible
+      include ::foreman::cli::azure
```
ekohl, line 16: This is the old style we got rid of
```suggestion
      include foreman::cli::ansible
      include foreman::cli::azure
```

ekohl:
In this module, you need to ensure the `en` locales are present due to how we configure postgresql. I wonder if we should ensure this in `foreman::install`.

wbclark:
> In this module, you need to ensure the `en` locales are present due to how we configure postgresql. I wonder if we should ensure this in `foreman::install`.

I've opened #837 to do this for the RedHat family of operating systems when the database is managed.

wbclark:
test failure is due to: https://projects.theforeman.org/issues/29429

mmoll:
merged, thanks @wbclark!

PR #832, opened 2020-04-29
Use legacy facts for puppetmaster defaults

ekohl:
875c8377f5dfadbe1cd69635d0925765e6c2a83b started to use modern facts.  However, for these branches there is no coverage and facterdb can't find these modern facts. This causes issues in puppet-puppet which does call these branches. Ignoring lint is the quickest fix now.

Longer term these classes will be split out to its own module as theforeman/puppetserver_foreman so there can be test coverage.


ekohl:
Turns out this was incorrect. Sadly there is no coverage for this.

PR #834, opened 2020-04-30
Switch AIO detection to use aio_agent_version fact

ekohl:
This fact is a core fact that's only set on AIO installs. Setting this custom fact is also easier than the structured fact in our test suite which increases reliability of our tests.

This is finishing what https://github.com/theforeman/puppet-foreman/pull/832 intended to fix.

### DIFF
--- a/manifests/params.pp
+++ b/manifests/params.pp
@@ -133,7 +133,7 @@ class foreman::params {
     }
   }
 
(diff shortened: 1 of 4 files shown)


mmoll:
merged, bedankt @ekohl!

PR #835, opened 2020-05-04
Ensure Foreman is provisioned before configuring cockpit

ekohl:
Previously foreman_config_entry could be executed before the database was available. This ensures the database is ready and the plugin installed before attempting to change a config setting.

### DIFF
--- a/manifests/plugin/remote_execution/cockpit.pp
+++ b/manifests/plugin/remote_execution/cockpit.pp
@@ -48,6 +48,7 @@ class foreman::plugin::remote_execution::cockpit {
   }
 
   foreman_config_entry { 'remote_execution_cockpit_url':
-    value => "${cockpit_path}/=%{host}",
+    value   => "${cockpit_path}/=%{host}",
+    require => Class['foreman::database'],
   }
 }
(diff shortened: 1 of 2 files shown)


manifests/plugin/remote_execution/cockpit.pp
```diff
@@ -48,6 +48,7 @@
   }
 
   foreman_config_entry { 'remote_execution_cockpit_url':
-    value => "${cockpit_path}/=%{host}",
+    value   => "${cockpit_path}/=%{host}",
+    require => Class['foreman::database'],
```
ehelms, line 52: Does this ensure that all of the class has executed before this will execute? e.g. db:seed

ekohl: Yes, those resources are contained within the class (if they're managed).

ehelms: Thanks, I just loose track of puppet and how it build it's dependency cycle. Sounds like we could clean this up a bit?

https://github.com/theforeman/puppet-katello/blob/master/manifests/application.pp#L36

ekohl: Yes, that would be better. I've had a look at doing this in the type definition, but since the config part is also used within that, we can't always require the class.

ekohl:
Looks like we need to use the SCL postgres due to Rails 6. I'll update https://github.com/theforeman/puppet-foreman/pull/699

lpramuk:
What's ETA here? This is blocking QE automation. Can we expect it to be in presnap 0.2  ?

mmoll:
@ekohl please rebase :)

PR #837, opened 2020-05-05
Refs #29212 - install en_US.utf-8 locale when required

wbclark:


manifests/install.pp
```diff
@@ -1,6 +1,13 @@
 # Install the needed packages for foreman
 class foreman::install {
 
+  if $foreman::db_manage and $facts['os']['family'] == 'RedHat' {
```
wbclark, line 4: I think it's pretty simple to re-work this for other OSes as well if needed, but what about other locales? We seem to be declaring `postgresql::server::db` with `en_US.utf8` hardcoded so I think I can ignore them?

wbclark:
The package exists for EL8 only, causing the acceptance test to fail on EL8. I'm not aware on EL7 equivalent so I can add an additional condition to only install it on EL8.

@ekohl given this, do you still think it makes sense to install it in `manifests/install.pp` as opposed to only in the EL8 acceptance test environment?

manifests/install.pp
ekohl: I agree. The DB is defined as en_US so this needs to be present.

manifests/install.pp
```diff
@@ -1,6 +1,13 @@
...
+    package { 'glibc-langpack-en':
+      ensure => installed,
+      before => Class['foreman::database'],
```
ekohl, line 7: This is a bit redundant since init.pp does the same thing.

ekohl:
https://github.com/theforeman/foreman-packaging/pull/5171 does this in packaging and might be easier.

manifests/install.pp
```diff
@@ -1,6 +1,13 @@
 # Install the needed packages for foreman
 class foreman::install {
 
+  if $foreman::db_manage and $facts['os']['family'] == 'RedHat' {
+    package { 'glibc-langpack-en':
```
evgeni, line 5: AFAIK this package does not exist on EL7, so you'd need to guard that somewhere.

evgeni:
> [theforeman/foreman-packaging#5171](https://github.com/theforeman/foreman-packaging/pull/5171) does this in packaging and might be easier.

So I've been wondering the same. Especially as the installer needs to be run with the right locale too (and I think we force it somewhere), so the package needs to be present *before* the installer runs?

evgeni:
That said, it definitely won't hurt.

ekohl:
> Especially as the installer needs to be run with the right locale too (and I think we force it somewhere), so the package needs to be present _before_ the installer runs?

It needs to be present before the database is created.

wbclark:
Closing in favor of https://github.com/theforeman/foreman-packaging/pull/5171

PR #839, opened 2020-05-11
Prefix ipa and sssd facts with foreman_

ekohl:
This prevents a collision with the ipa fact from the ipa module.


mmoll:
merged, bedankt @ekohl!

tbrisker:
can we cherry-pick this for foreman 2.0.1? I'm guessing that backward incompatible label means it's complicated?

ekohl:
In the cherry picked version we can keep in the old fact and just copy it in case anyone relies on that. It shouldn't matter for us.

PR #840, opened 2020-05-12
Allow puppet/redis 6.x

ekohl:
6.x was breaking since it made repository management modules soft dependencies. Since we don't use these, it's not breaking here.

### DIFF
--- a/metadata.json
+++ b/metadata.json
@@ -42,7 +42,7 @@
     },
     {
       "name": "puppet/redis",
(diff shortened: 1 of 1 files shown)


ekohl:
Ubuntu failed with an unrelated error:
```
May 12 10:52:46 ubuntu1804-64.example.com dynflow-sidekiq@orchestrator[6872]: More than 1 row in migrator table
```

PR #843, opened 2020-05-14
Fixes #29817 - Implement a dynflow worker pool

ekohl:


ehelms:
I like where this is headed. As part of this update design to allow more scaling, it feels prudent to consider whether the design can handle specifying both number of workers and type of work any of the workers are handling. For example, if I want to be able to spin up 3 workers dedicated to REX and 2 dedicated to handling Katello host actions and 1 set to default.

I'd be happy to get this in as a first pass, if the design won't require a re-work to support the above.

ekohl:
The idea is that for every pool you can set the number of workers and threads for each worker. Similar to Puma, you can't set the number of threads for each individual worker. IMHO that's too complex and no real value.

This is breaking and Katello probably needs to be modified to even work:
https://github.com/theforeman/puppet-katello/blob/990b7d2c78a31d271064e912760f8dc0fe0a6f2d/manifests/application.pp#L84-L86
That would be a different pool and in katello we would introduce a parameter to tune that pool size. As you can see, orchestrator is not a pool but a single worker.

There is no code purge unknown workers / pools. That required too much guessing, thought the custom fact could provide it since the information is there.

ekohl:
So there currently is an idempotence issue where the fact is calculated before the package is installed. Then the package creates worker.yml. On the next run, it finds worker.yml that shouldn't be there and removes it.

manifests/dynflow/pool.pp
```diff
@@ -0,0 +1,51 @@
...
+#
+# @param workers
+#   Defines number of workers used for job processing. Set to 0 to clean up old
+#   instances with the service name.
+#
+# @param concurrency
+#   Defines number of threads used for job processing
+#
+# @param queues
+#   The Queues this worker should process
```
ehelms, line 15: Should this read "queues this pool of workers should process" ?

ehelms:
Looking through the examples in the tests, I expected to see setting a worker / queue combination but I still just see setting a number of workers. I am still confused as the end user configuring dynflow what I would be configuring for my deployment.

ekohl:
I haven't spent that much time yet on the usability part, only focused on the technical aspect to see if it was feasible. That's the case, but given the packaging changes needed I'm proposing we release this module with a major version number now to release rc1 and do a breaking change soon after for rc2.

ekohl:
I've opened https://community.theforeman.org/t/foreman-dynflow-pool/18955 because I need some design answers that I'm currently blocked on. It's also good for visibility.

manifests/params.pp
```diff
@@ -82,10 +82,16 @@
...
   $jobs_service_ensure = 'running'
   $jobs_service_enable = true
-  $jobs_sidekiq_redis_url = undef
+  $dynflow_redis_url = undef
+
+  $dynflow_manage_services = true,
+  $dynflow_orchestrator_ensure = 'present',
+  $dynflow_worker_instances = 1,
+  $dynflow_worker_concurrency = 5,
+  $dynflow_redis_url = undef,
```
ehelms, line 94: No comma's here :)

lib/facter/dynflow.rb
```diff
@@ -0,0 +1,11 @@
+Facter.add(:foreman_dynflow) do
+  setcode do
+    directory = '/etc/foreman/dynflow'
+    extension = '.yml'
+    return unless File.exist?(directory)
+
+    Dir[File.join(directory, "*#{extension}")].map do |config|
+      File.basename(config, extension)
+    end
```
wbclark, line 9:
```suggestion
    if File.exist?(directory)
      Dir[File.join(directory, "*#{extension}")].map do |config|
        File.basename(config, extension)
      end
    else
      {}
    end
```

To prevent annoying `Error when resolving custom fact: unexpected return` output in the case where it is being applied for the first time.

ekohl: Rather than `{}` I think you should return `nil`

wbclark: You're absolutely right. `nil` is what actually worked for me in my testing. I made a typo in the suggestion here.

ekohl:
Updated. It no longer uses return in the fact and I've fixed some incorrect variables.

ekohl:
Rebased, no other changes.

ehelms:
I'd be curious your thoughts on the implications this has for PostgreSQL connections, this change provides a vector for users to scale up the number of workers and exhaust the default number of available connections (which is 100 I believe) as I understand things. I am wondering if this needs to be part of tuning profiles in the installer, or hook that does a rough calculation and either warns the user or configures the available connections to a value higher than the calculated value of needed connections (this is complicated by externally managed databases).

ehelms (approved):
Looking like these changes will require an installer migration to drop the old parameters or rename them for upgrades.

ekohl:
Removal automatically happens, but renaming the pool size to concurrency is probably maintaining the configured installation.

ehelms:
@ekohl would you like to merge this and then follow up? or wait on the migration PR?

jlsherrill:
@ekohl any chance we could get this for katello 4.1/Foreman 2.5, i more specifically care about the 'host' queue workers

ehelms:
@ekohl this was last ran against Travis, could you rebase it so that we can see the GH actions based tests run?

jlsherrill:
@ekohl this is really really desired for 4.1 and let users better scale performance, is anything holding this back?

ehelms:
Thanks @ekohl !

PR #846, opened 2020-05-20
Fixes #29892 - Use server certs for websockets

ekohl:
Prior to this the user needed to specify the certs both for the server and websockets. In practice these are pretty much always the same files.  By using undef + pick() the option to specify these is maintained, but the defaults are better.


mmoll:
merged, bedankt @ekohl!

PR #847, opened 2020-05-27
fixes #29938 - change default logging layout

domitea:
Complementary PR for https://github.com/theforeman/foreman/pull/7220 to make default new kind of logger mentioned in liked PR

### DIFF
--- a/README.md
+++ b/README.md
@@ -76,6 +76,9 @@ previous stable release.
 
 This module targets Foreman 2.0+.
 
+If you're running Foreman instance older than Foreman 2.2 you need to change logging layout parameter from 
+`multiline_request_pattern` to `multiline_pattern` due to compatibility.  
+
 ## Types and providers
 
 `foreman_config_entry` can be used to manage settings in Foreman's database, as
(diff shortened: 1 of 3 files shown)


ekohl:
If we want to make this active for all existing installations, it also needs an installer migration.

tbrisker:
Core has been merged, @domitea please add the migration as suggested

domitea:
Migration added in https://github.com/theforeman/foreman-installer/pull/557, also Readme was affected by this change.

README.md
```diff
@@ -74,7 +74,7 @@ previous stable release.
 
 ### Foreman version compatibility notes
 
-This module targets Foreman 2.0+.
+This module targets Foreman 2.2+.
```
ekohl, line 77: We don't do this, see the support policy. Instead, we would write a new paragraph that on Foreman versions older than 2.2 you need to change the logging layout parameter to `pattern`.

ekohl:
And you need to add the new value to the valid values in `Enum`.

domitea:
@ekohl thanks fro mentioning enum! Readme updated.

tbrisker:
@ekohl what's the status here?

ekohl (approved):

ekohl:
It sort of slipped form my radar.

PR #862, opened 2020-06-24
Drop Puppet integration bits

ekohl:
These are moved to their own theforeman/puppetserver_foreman module.  This allows us to actually test the additional operating systems that Foreman doesn't support, but the integration does.

Draft since it depends on https://github.com/theforeman/puppet-puppetserver_foreman/pull/1. The Foreman manual also refers to these files and needs to be updated. The README of theforeman/puppet also needs to be updated to reflect this.


ekohl:
Superseded by https://github.com/theforeman/puppet-foreman/pull/933

PR #863, opened 2020-06-25
Fixes #30078 - add parameter to accept a hostgroup config hash

apatelKmd:


templates/default_hostgroup.yaml.erb
```diff
@@ -0,0 +1,10 @@
+---
+:default_hostgroup:
+  :facts_map:
+  <%= @hostgroups.each do |group, fact_kv| -%>
```
ekohl, line 4: It'd a good idea to keep it sorted to avoid rewrites (same below)
```suggestion
  <%= @hostgroups.sort.each do |group, fact_kv| -%>
```

manifests/plugin/default_hostgroup.pp
```diff
@@ -1,4 +1,20 @@
...
+# = Foreman Default Hostgroup
+#
+# This class installs the default_hostgroup plugin and optionally manages the configuration file
+#
+# === Parameters:
+#
+# $hostgroups:: An hash of hostgroup names and facts to add to the configuration
+#
+class foreman::plugin::default_hostgroup (
+  Hash $hostgroups = {},
```
ekohl, line 10: Would it be a good idea to further specify this?
```suggestion
  Hash[String, Hash[String, Variant[String, Integer]]] $hostgroups = {},
```

manifests/plugin/default_hostgroup.pp
```diff
@@ -1,4 +1,20 @@
-# Installs foreman_default_hostgroup plugin
-class foreman::plugin::default_hostgroup {
-  foreman::plugin {'default_hostgroup': }
+# = Foreman Default Hostgroup
+#
+# This class installs the default_hostgroup plugin and optionally manages the configuration file
+#
+# === Parameters:
+#
+# $hostgroups:: An hash of hostgroup names and facts to add to the configuration
```
ekohl, line 7: For simple documentation (when there are no parameter groups) we prefer puppet-strings' yardoc format:

```suggestion
# @summary This class installs the default_hostgroup plugin and optionally manages the configuration file
#
# @param hostgroups An hash of hostgroup names and facts to add to the configuration
```

manifests/plugin/default_hostgroup.pp
```diff
@@ -1,4 +1,20 @@
...
+  Hash $hostgroups = {},
+){
+  if $hostgroups {
+    foreman::plugin {'default_hostgroup':
+      config => template('foreman/default_hostgroup.yaml.erb')
+    }
+  }
+  else {
+    foreman::plugin {'default_hostgroup': }
+  }
```
ekohl, line 19: I'd prefer something like:
```suggestion
  if $hostgroups {
    $config = template('foreman/default_hostgroup.yaml.erb')
  } else {
    $config = undef
  }

  foreman::plugin {'default_hostgroup':
    config => $config,
  }
```

Also wonder if `$hostgroups` doesn't always evaluate to true and you need an explicit empty check, but there's a test so that should be covered.

manifests/plugin/default_hostgroup.pp
apatelKmd: I like it.

templates/default_hostgroup.yaml.erb
apatelKmd: I am big fan of sorted hash, I rushed this one.

spec/classes/plugin/default_hostgroup_spec.rb
```diff
@@ -2,4 +2,38 @@
...
+        'Redhat':
+          'osfamily': 'RedHat'
+          'lsbdistcodename': 'Santiago'
+        'Osx/common':
+          'osfamily': 'Darwin'
+        'Base':
+          'hostname': '.*'
+      }
+
+hostgroups = <<EOF
```
ekohl, line 18: You can indent this line. You can even indent the content using `<<~`.

ekohl:
Would you mind squashing the commits? Also use `Fixes #30078 - <message>`. Looks like now the check doesn't detect it.

spec/classes/plugin/default_hostgroup_spec.rb
apatelKmd: I was trying to execute rspec locally but did not find a way to do it. Can you share how to execute the spec test locally.

apatelKmd: The `~` removes the indentation and I am trying to test the exact content.

ekohl: https://github.com/theforeman/puppet-foreman/blob/master/CONTRIBUTING.md#getting-started has a bit more, but the tl;dr is:
```
bundle install
bundle exec rake spec_prep
bundle exec rspec spec/classes/plugin/default_hostgroup_spec.rb
```

manifests/plugin/default_hostgroup.pp
```diff
@@ -1,4 +1,20 @@
...
+#
+class foreman::plugin::default_hostgroup (
+  Array[Hash] $hostgroups = [],
+){
+  if empty($hostgroups) {
+    $config = undef
+    $config_file = undef
+  } else {
+    $config = template('foreman/default_hostgroup.yaml.erb')
+    $config_file = "${foreman::plugin_config_dir}/default_hostgroup.yaml"
```
apatelKmd, line 13: The config file path uses a prefix `foreman` in the plugin type. Based on the default_hostgroup plugin it expects the file name `default_hostgroup.yaml` and not the `foreman-default_hostgroup.yaml`.

apatelKmd:
I want to know how can I use the parameter with multiple hashes. For e.g, I want to add the same parameters as described in the spec test. Can you help with it?

`foreman-installer --foreman-plugin-default-hostgroup-hostgroups ???`

ekohl:
[kafo](https://github.com/theforeman/kafo) is the engine behind exposing parameters. In particular https://github.com/theforeman/kafo#hash-arguments is relevant. However, I don't know how it would deal with a hash inside a hash. I think there's even a bug in that `--param=key:value` doesn't work and needs to be `--param key:value`. Perhaps `--param key:nested_key:value` works. Otherwise that might be a possible RFE.

Another solution is to put it in custom-hiera.yaml. This is not really what we want (answers for exposed classes should be preferred), but if it's the only way, it has to be that way. Then you would use (from the top of my head):
```yaml
foreman::plugin::default_hostgroup::hostgroups:
  key:
    nested_key: value
```

apatelKmd:
Thanks @ekohl for clarifying the use of the parameter. Use of `custom-hiera.yaml` will do. What version of the foreman-installer will include the update? If we know the future plan we can prepare our installer arguments accordingly.

ekohl:
At this moment we're in the Foreman 2.2 development cycle. If it's merged in the next few weeks, that'll be the version to contain it.

apatelKmd:
@ekohl AFAIK, the changes are complete. Please let me know if you want me to make an update?

manifests/plugin/default_hostgroup.pp
```diff
@@ -1,4 +1,20 @@
...
+  }
+
+  foreman::plugin {'default_hostgroup':
+    config      => $config,
+    config_file => $config_file,
```
ekohl, line 18: You can set this unconditionally:
```suggestion
    config_file => "${foreman::plugin_config_dir}/default_hostgroup.yaml",
```

The file is only managed if the config is set:
https://github.com/theforeman/puppet-foreman/blob/9bb72cee31c3d7265f4f83c9dc222f7205848fed/manifests/plugin.pp#L46-L55

manifests/plugin/default_hostgroup.pp
```diff
@@ -1,4 +1,20 @@
-# Installs foreman_default_hostgroup plugin
-class foreman::plugin::default_hostgroup {
-  foreman::plugin {'default_hostgroup': }
+# @summary This class installs the default_hostgroup plugin and optionally manages the configuration file
+#
+# @param hostgroups An array of hashes of hostgroup names and facts to add to the configuration
+#
+class foreman::plugin::default_hostgroup (
+  Array[Hash] $hostgroups = [],
```
ekohl, line 6: I think you can further narrow it down:
```suggestion
  Array[Hash[String, Hash]] $hostgroups = [],
```

apatelKmd: Done

manifests/plugin/default_hostgroup.pp
apatelKmd: Done

apatelKmd:
@ekohl I have incorporated your suggestions followed by squash and rebase.

spec/classes/plugin/default_hostgroup_spec.rb
```diff
@@ -2,4 +2,45 @@
...
+    it { is_expected.to compile.with_all_deps }
+    it { is_expected.to contain_file('/etc/foreman/plugins/default_hostgroup.yaml').with_content(hostgroups) }
+  end
+
+  context 'no config hash' do
+    it { is_expected.not_to contain_file('/etc/foreman/plugins/default_hostgroup.yaml') }
+  end
+
 end
+
```
ekohl, line 46: We should really enable rubocop to catch these, but for now I'll suggest it manually. Not sure if Github captures it well, but please drop the empty lines
```suggestion
end
```

apatelKmd: Done

apatelKmd:
Lint test failed, I pushed an update to resolve the issue.

ekohl:
Thanks!

PR #865, opened 2020-07-01
Expand error message on SP retrieval

ekohl:
It may not be clear to everyone that this is the response from Foreman. Often we see people only post the installer log but the that's not helpful when the Foreman application had an error.

### DIFF
--- a/lib/puppet/provider/foreman_smartproxy/rest_v3.rb
+++ b/lib/puppet/provider/foreman_smartproxy/rest_v3.rb
@@ -4,7 +4,7 @@ Puppet::Type.type(:foreman_smartproxy).provide(:rest_v3, :parent => Puppet::Type
   def proxy
     @proxy ||= begin
       r = request(:get, 'api/v2/smart_proxies', :search => %{name="#{resource[:name]}"})
-      raise Puppet::Error.new("Proxy #{resource[:name]} cannot be retrieved: #{error_message(r)}") unless success?(r)
+      raise Puppet::Error.new("Proxy #{resource[:name]} cannot be retrieved from Foreman: #{error_message(r)}") unless success?(r)
       JSON.load(r.body)['results'][0]
     end
   end


lib/puppet/provider/foreman_smartproxy/rest_v3.rb
```diff
@@ -4,7 +4,7 @@
   def proxy
     @proxy ||= begin
       r = request(:get, 'api/v2/smart_proxies', :search => %{name="#{resource[:name]}"})
-      raise Puppet::Error.new("Proxy #{resource[:name]} cannot be retrieved: #{error_message(r)}") unless success?(r)
+      raise Puppet::Error.new("Proxy #{resource[:name]} cannot be retrieved from Foreman: #{error_message(r)}") unless success?(r)
```
ekohl, line 7: Should we suggest to consult the Foreman logs as well?

https://github.com/theforeman/puppet-foreman/blob/746229243abb98995981f90d1a777e096d201108/lib/puppet/provider/foreman_smartproxy/rest_v3.rb#L70 does this

tbrisker, line 7: would it make sense to point to production.log on the foreman for more info? I'm assuming the error message in most cases here might not be very informative regarding why it failed other than http status code

ekohl: Yes, most often it was HTTP 5xx which people didn't understand. That can be in Foreman or Apache or in between. On HTTP 4xx we can display an error:

https://github.com/theforeman/puppet-foreman/blob/746229243abb98995981f90d1a777e096d201108/lib/puppet/provider/foreman_resource/rest_v3.rb#L97-L99

Sadly, I think we don't often have this.

ehelms:
Given our discussions today around these error messages, do you want to get this in a small fix or meld it into the greater idea of displaying more cases based on error code?

ekohl:
For context: @wbclark, @ehelms and I had a discussion about this and potential other improvements. I feel that we should capture those and do it right the first time. There's a list of those, but essentially what https://github.com/theforeman/puppet-foreman/pull/865#discussion_r448325914 started and more. Thus it's a draft again.

ehelms:
Redmine issue to capture the desired intent: https://projects.theforeman.org/issues/30284

PR #867, opened 2020-07-21
Fixes #30456 - Fix missing icons on /pub page

adamruzicka:
The requests for `/icons/*` were being proxy-passed to puma, which didn't know about any icons and returned 404s.


adamruzicka:
Test failures seem unrelated

ekohl:
Yes, it's one we see more often. I think it has to do with sidekiq returning it has started before it actually has. https://projects.theforeman.org/issues/30275 would probably solve that.

PR #868, opened 2020-07-31
added foreman_datacenter

Zenya:


wbclark (approved):

wbclark:
Tested this on EL7 and EL8 and found that it works nicely. Thanks @Zenya -- sorry for the delay in reviewing.

PR #871, opened 2020-08-12
Fix comparison code in test

ekohl:


ekohl:
I'm still quite hesitant to just rewrite the tests. They're huge patches and for now they're just deprecated, not planned to be removed AFAIK. Consistency is good thing, but just changing things for the sake of changing them is a lot of work.

What I do find surprising is that this passed both before and after. That shouldn't happen. Looking into this.

ekohl:
I suspect it modifies the facts in place which causes the compilation of the catalog to happen with the other value. Not a lot we can do about that.

PR #872, opened 2020-08-19
Fixes #30535 - Set HTTP headers for proxy requests

hsahmed:


templates/auth_kerb.conf.erb
```diff
@@ -10,6 +10,17 @@
...
+  RewriteRule .* - [E=HTTP_REMOTE_USER:%{REMOTE_USER}]
+  RequestHeader set REMOTE_USER %{REMOTE_USER}e
+  RewriteRule .* - [E=HTTP_REMOTE_USER_EMAIL:%{REMOTE_USER_EMAIL}]
+  RequestHeader set REMOTE_USER_EMAIL %{REMOTE_USER_EMAIL}e
+  RewriteRule .* - [E=HTTP_REMOTE_USER_FIRSTNAME:%{REMOTE_USER_FIRSTNAME}]
+  RequestHeader set REMOTE_USER_FIRSTNAME %{REMOTE_USER_FIRSTNAME}e
+  RewriteRule .* - [E=HTTP_REMOTE_USER_LASTNAME:%{REMOTE_USER_LASTNAME}]
+  RequestHeader set REMOTE_USER_LASTNAME %{REMOTE_USER_LASTNAME}e
+  RewriteRule .* - [E=HTTP_REMOTE_USER_GROUPS:%{REMOTE_USER_GROUPS}]
+  RequestHeader set REMOTE_USER_GROUPS %{REMOTE_USER_GROUPS}e
```
ehelms, line 23: Should each of these also come with a corresponding line to unset the value before Apache sets it?

hsahmed: Since set replaces the existing value, the unset should not really be required. I have added unset anyways since I have seen other projects follow that practice sometimes.

templates/auth_kerb.conf.erb
```diff
@@ -8,6 +8,20 @@
   KrbAuthRealms <%= @facts['foreman_ipa']['default_realm'] %>
   Krb5KeyTab <%= scope.lookupvar('::foreman::http_keytab') %>
   KrbLocalUserMapping On
+
+  # Set headers for proxy requests
+  RequestHeader unset REMOTE_USER
+  RequestHeader unset REMOTE_USER_EMAIL
+  RequestHeader unset REMOTE_USER_FIRSTNAME
+  RequestHeader unset REMOTE_USER_LASTNAME
+  RequestHeader unset REMOTE_USER_GROUPS
```
tbrisker, line 17: If i'm not mistaken, this template is only used if the installer is used to set up external auth, while we would want to always clear these headers if they come from the user

hsahmed: The only place in foreman code these headers are read is in the sso/apache.rb which goes to extlogin path before reading them, one more place reads REMOTE_USER for what appears to be session management (in app/controllers/application_controller.rb), I have not modified it there and it is not affecting anything during my tests. Since these are not read anywhere else in Foreman, I don't think we have to clear them every time.

tbrisker: I think one concern @ekohl raised on the other pr is that someone might turn on the setting on foreman by mistake without configuring apache accordingly. These headers should never originate from the user, so I believe it should be safe to always clear them.
the other place that uses `REMOTE_USER` should also be updated to the new header - it doesn't break anything, but for remote users there is some different handling of session expiry.

hsahmed: I have amended the commit to unset them in the main apache configuration for clearing the RequestHeaders. Do you thing there is need to change "REMOTE_USER environment variable" to "REMOTE_USER HTTP header" in settings description and translations?

tbrisker: That sounds like a good idea to update the description, don't worry about the translations as they are generated automatically as part of the release process.

hsahmed: I have pushed the updated code with updated translation and tests.

templates/lookup_identity.conf.erb
```diff
@@ -3,6 +3,7 @@
   LookupUserAttr email REMOTE_USER_EMAIL
   LookupUserAttr firstname REMOTE_USER_FIRSTNAME
   LookupUserAttr lastname REMOTE_USER_LASTNAME
+  LookupUserGroups REMOTE_USER_GROUPS :
   LookupUserGroupsIter REMOTE_USER_GROUP
```
ekohl, line 7: Just to verify: the old `REMOTE_USER_GROUP_N` env var is still present, right? That's good to have for backwards compatibility. Just to be clear: I like the new implementation more but I want to understand the impact.

ekohl (commented):
From a quick glance this looks correct. I'll be out until the next week, but I'll try to give it a spin then. It looks like this is backwards compatible which always makes me happy.

hsahmed:
@ekohl Yes, I have left the old env variables as is.

templates/lookup_identity.conf.erb
hsahmed: Yes I have left the old env variables as is.

wiad:
I'm having some problems making this work, using `mod_auth_cas`. This module sets it's own headers, like `HTTP_CAS_MAILADDRESS` and I'm not succeding in using those headers to create the ones needed by Foreman (like `HTTP_REMOTE_USER_EMAIL`), they only turn out with a value of `(null)`. 

The ugly workaround is to edit `app/services/sso/apache.rb` Foreman and replace the headers there with my CAS headers.

templates/auth_kerb.conf.erb
```diff
@@ -8,6 +8,14 @@
   KrbAuthRealms <%= @facts['foreman_ipa']['default_realm'] %>
   Krb5KeyTab <%= scope.lookupvar('::foreman::http_keytab') %>
   KrbLocalUserMapping On
+
+  # Set headers for proxy requests
```
sbuzonas, line 12: Setting these here won't apply to the intercept for form login.  I put them in lookup_identity which covers `^/users/(ext)?login$` and it fixes both form intercept and ticket authentication

tbrisker:
@hsahmed could you update according to the latest comments please?

hsahmed:
> @hsahmed could you update according to the latest comments please?

I have moved the headers to lookup_identity, however I have not been able to test it yet. I will test and amend the commit later today.

hsahmed:
I have moved the headers to lookup_identity and they're now working for both SSO and intercept login.

ehelms (approved):
@ekohl final thoughts before merging?

ekohl (approved):

sbuzonas:
Upon reapplying this patch to after an upgrade, I noticed an inconsistency.  The `unset REMOTE_USER_USER_GROUPS`, but it's really passing `REMOTE_USER_GROUPS`

tbrisker:
Thanks for noticing @sbuzonas ! I've opened https://github.com/theforeman/puppet-foreman/pull/896 to fix that.

PR #879, opened 2020-08-20
Changed managed user resource to system type

danielsreichenbach:
Foreman packages by default deploy users as system users. This ensures the module does the same.


ekohl (commented):
Makes sense, but should the same be done in the group?

ekohl:
Replaced by https://github.com/theforeman/puppet-foreman/pull/971

PR #880, opened 2020-08-20
Fixes #29649 - Handle missing realm/server in IPA config

hboetes:
Fixes #29649 - Handle missing realm/server in IPA config


lib/facter/sssd.rb
```diff
@@ -1,14 +1,14 @@
...
   Facter.add(:foreman_ipa, :type => :aggregate) do
     {
-      :default_realm => 'global/realm',
-      :default_server => 'global/server',
+      :default_server => 'target[.=~regexp("domain/.*")][1]/ipa_server',
+      :default_realm  => 'target[.="sssd"]/domains'
     }.each do |key, path|
       chunk(key) do
-        val = Facter::Util::Sssd.ipa_value(path)
+        val = Facter::Util::Sssd.sssd_value(path)
```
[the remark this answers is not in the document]
hboetes: OK, I have a new version, this should be better.

hboetes:
And to show the output I get now:
```
# /opt/puppetlabs/bin/facter --custom-dir ~/src/puppet-foreman/lib/facter foreman_ipa foreman_sssd
foreman_ipa => {
  default_realm => "EXAMPLE.COM"
}
foreman_sssd => {
  server => [
    "gandalf.example.com",
    "olorin.example.com",
    "mithrandir.example.com"
  ],
  services => [
    "nss",
    "pam",
    "sudo",
    "ssh"
  ]
}

```

ekohl:
Is it needed to split the files? They're so similar.

hboetes:
If you want them to be in the same files the `sssd.rb` files should be renamed to something more generic. What would you propose?

ekohl:
My main problems is the duplication. AFAIK the filename isn't really that relevant for the actual functionality. Perhaps `realm.rb` since it's used to integrate with a realm?

hboetes:
I like that. Just a sec.

lib/facter/util/realm.rb
```diff
@@ -15,6 +15,15 @@ def self.ipa_value(path)
...
 
+    def self.aug_value(lens, file, path)
+      Augeas::open(nil, nil, Augeas::NO_MODL_AUTOLOAD) do |aug|
+        aug.transform(:lens => lens, :incl => file)
+        aug.load
+        aug.set('/augeas/context', "/files#{file}")
+        aug.get(path)
+      end
+    end
+
```
ekohl, line 26: This method is not needed anymore.

lib/facter/realm.rb
```diff
@@ -1,32 +1,30 @@
...
   Facter.add(:foreman_ipa, :type => :aggregate) do
     {
       :default_realm => 'global/realm',
       :default_server => 'global/server',
     }.each do |key, path|
       chunk(key) do
-        val = Facter::Util::Sssd.ipa_value(path)
+        val = Facter::Util::Realm.ipa_value(path)
         {key => val} if val
+        # == Fact: foreman_sssd
```
ekohl, line 12: This belongs a few lines below.

hboetes:
Hard to focus today. Thanks for your suggestions.

lib/facter/realm.rb
```diff
@@ -1,32 +1,31 @@
...
       end
     end
   end
-
   # == Fact: foreman_sssd
   Facter.add(:foreman_sssd, :type => :aggregate) do
     {
       :services => 'target[.="sssd"]/services',
+      :server => 'target[.=~regexp("domain/.*")][1]/ipa_server',
```
ekohl, line 20: Since this is a list, should it be `servers`?

hboetes: It is an array to should at least have one member. Since the original configuration option is `ipa_server`, I went with `server`. I think it is relevant if it is either one.

hboetes:
I've read the check logs twice now and I have no idea why they failed. Last time it was xenial that failed, now it's centos-8 that fails. And I didn't change anything in between.

ekohl:
Yes, there is a bit of a timing problem. The service start returns before it really started. Unrelated issue to this.

lib/facter/realm.rb
ekohl: I still think it should be `servers` here since it's an array. The name should give a hint about the data type to expect.

hboetes: OK, done.

ekohl (commented):
So I think that all that's left is to modify the call to `/usr/sbin/ipa-getkeytab` to pick the right server. What I'd probably do is something like this:

```puppet
if fact('foreman_ipa.default_server') {
  $ipa_server = fact('foreman_ipa.default_server')
} elseif fact('foreman_sssd.servers') and !empty(fact('foreman_sssd.servers')) {
  $ipa_server = fact('foreman_sssd.servers')[0]
} else {
  $ipa_server = undef
}

unless $ipa_server and fact('foreman_ipa.default_realm') {
  fail()
}
```

Not sure if just taking the first server is the best. Perhaps there's a better way.

hboetes:
> Not sure if just taking the first server is the best. Perhaps there's a better way.

Now that you mention it, from the sssd-ipa manpage: https://linux.die.net/man/5/sssd-ipa 
```
If no servers are specified, the back end automatically uses service discovery to try to find a server. Optionally, the user may choose to use both fixed server addresses and service discovery by inserting a special keyword, "_srv_", in the list of servers. The order of preference is maintained. This feature is useful if, for example, the user prefers to use service discovery whenever possible, and fall back to a specific server when no servers can be discovered using DNS. 
```

So in case the first server returns _srv_ or nothing at all you still could do a dns lookup, as described here: https://www.freeipa.org/page/Howto/Updating_FreeIPA_system_DNS_records_on_a_remote_DNS_server

```
[user@ipa ~]$ ipa dns-update-system-records --dry-run
 IPA DNS records:
   _kerberos-master._tcp.example.com. 86400 IN SRV 0 100 88 ipa.example.com.
   _kerberos-master._udp.example.com. 86400 IN SRV 0 100 88 ipa.example.com.
   _kerberos._tcp.example.com. 86400 IN SRV 0 100 88 ipa.example.com.
   _kerberos._udp.example.com. 86400 IN SRV 0 100 88 ipa.example.com.
   _kerberos.example.com. 86400 IN TXT "EXAMPLE.COM"
   _kpasswd._tcp.example.com. 86400 IN SRV 0 100 464 ipa.example.com.
   _kpasswd._udp.example.com. 86400 IN SRV 0 100 464 ipa.example.com.
   _ldap._tcp.example.com. 86400 IN SRV 0 100 389 ipa.example.com.
   _ntp._udp.example.com. 86400 IN SRV 0 100 123 ipa.example.com.
   ipa-ca.example.com. 86400 IN A 192.0.2.36
   ipa-ca.example.com. 86400 IN AAAA 2001:db8:0:224e:21a:4aff:fe23:1523
```

Or like this, which requires a valid Kerberos ticket:

```
# ipa server-find
---------------------
3 IPA servers matched
---------------------
  Server name: gandalf.example.com
  Min domain level: 1
  Max domain level: 1

  Server name: mithrandir.example.com
  Min domain level: 1
  Max domain level: 1

  Server name: olorin.example.com
  Min domain level: 1
  Max domain level: 1
----------------------------
Number of entries returned 3
----------------------------
```
So, TLDR, non of these methods are guaranteed to work. I'll ask around at the #freeipa channel on freenode, they know a lot.

hboetes:
Perhaps the much simpler solution is to keep checking if freeipa is properly installed out of the installer and document it properly.

ekohl:
The `fail()` is there because we want to verify if IPA is installed but because the variable is used here:
https://github.com/theforeman/puppet-foreman/blob/e3bf011200d2519895627e29f0cb458dc47338b7/manifests/config.pp#L176

This PR doesn't address that part yet which makes it incomplete. If the `-s` parameter is optional and can be left out, then that's the correct fix.

hboetes:
OK, let me share my discussion in the freeipa channel with two redhat employees who work on freeipa:
```
19:11        Han| ab, rcrit, what would be the simplest way to figure out a valid freeipa server? We're trying to integrate freeipa in the foreman: 
                  https://github.com/theforeman/puppet-foreman/pull/880#issuecomment-682936823
19:12        Han| So far we're searching /etc/ipa/default.conf and /etc/sssd/sssd.conf
19:13        Han| That user probably won't have a valid kerberos ticket, sssd may not contain a list, I've never seen it in default.conf
19:14        Han| This is used in a program that installs theforeman and tries to figure out if freeipa is installed.
19:17      rcrit| ipa-getkeytab will use the IPA server configured in /etc/ipa/default.conf if no -s is passed in
19:17      rcrit| you can do your own discovery with: dig -t SRV ldap._tcp.example.com
19:17      rcrit| but not everyone uses DNS discovery
19:25        Han| ipa-getkeytab requires a valid kerberos ticket. I prefer to avoid that. Besides, there is no server defined in our default.conf.
19:26      rcrit| you can't get a keytab without authentication
19:36        Han| True, perhaps this shouldn't be automated.
19:54         ab| Han: ipa-getkeytab accepts LDAP bind too
20:49        Han| ab, then we'd still have to figure out the name of a valid server, right? Or am I missing something?
09:26         ab| yes
```
I think that rcrit's first remark at 19:17 answers your question.

ekohl:
That sounds like we don't need the `default_server` fact at all and can just omit the parameter then. Looks like a much more elegant solution. The `default_realm` check is still needed because it's used in `templates/auth_kerb.conf.erb`.

ekohl:
> That sounds like we don't need the `default_server` fact at all and can just omit the parameter then. Looks like a much more elegant solution. The `default_realm` check is still needed because it's used in `templates/auth_kerb.conf.erb`.

I think https://github.com/theforeman/puppet-foreman/pull/935 would implement this. Could you try it out? Ideally also on a fresh deployment where it creates a new keytab file. I should set up my own IPA instance and test it, but right now I don't have that.

hboetes:
I think I can try this patch whilst running the next upgrade. It will probably complain, after which I can apply this patch and then see what happens. I will keep you posted.

hboetes:
Hmmm that's odd, I can't find that file:
```
root@theforeman /usr/share/foreman-installer/modules/foreman # patch -p1 -i ~/935.diff 
patching file lib/facter/sssd.rb
patching file manifests/config.pp
can't find file to patch at input line 49
Perhaps you used the wrong -p or --strip option?
The text leading up to this was:
--------------------------
|diff --git a/spec/classes/foreman_config_ipa_spec.rb b/spec/classes/foreman_config_ipa_spec.rb
|index 58102811..fffb92ec 100644
|--- a/spec/classes/foreman_config_ipa_spec.rb
|+++ b/spec/classes/foreman_config_ipa_spec.rb
--------------------------
File to patch: ^C
```

And I still got:

```
2021-04-20 11:29:40 [ERROR ] [configure] Evaluation Error: Error while evaluating a Function Call, theforeman: The system does not seem to be IPA-enrolled (file: /usr/share/foreman-installer/modules/foreman/manifests/config.pp, line: 154, column: 9) on node theforeman.example.com
```

So I commented out those lines and proceeded as usual.

ekohl:
Closing in favor of https://github.com/theforeman/puppet-foreman/pull/935

PR #882, opened 2020-09-07
Fixes #30789 - Set DB pool size dynamically

ekohl:
Every thread in the Puma worker can open a database connection. This means it needs to be taken into account to avoid exhasuting the pool.

### DIFF
--- a/manifests/config.pp
+++ b/manifests/config.pp
@@ -35,6 +35,12 @@ class foreman::config {
     mode  => '0640',
   }
 
+  if $foreman::use_foreman_service {
+    $db_pool = max($foreman::db_pool, $foreman::foreman_service_puma_threads_max)
+  } else {
+    $db_pool = $foreman::db_pool
+  }
+
   file { '/etc/foreman/database.yml':
     owner   => 'root',
     group   => $foreman::group,
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -83,7 +83,8 @@
 #
 # $db_root_cert::                 Root cert used to verify SSL connection to postgres
 #
-# $db_pool::                      Database 'production' size of connection pool
+# $db_pool::                      Database 'production' size of connection pool. When running as a reverse proxy,
+#                                 the value of `$foreman_service_puma_threads_max` is used if it's higher than `$db_pool`.
 #
 # $db_manage_rake::               if enabled, will run rake jobs, which depend on the database
 #
--- a/templates/database.yml.erb
+++ b/templates/database.yml.erb
@@ -27,6 +27,4 @@
 <% unless (password = scope.lookupvar("::foreman::db_password")) == 'UNSET' -%>
   password: "<%= password %>"
 <% end -%>
-<% unless (pool = scope.lookupvar("::foreman::db_pool")) == 'UNSET' -%>
-  pool: <%= pool %>
-<% end -%>
+  pool: <%= @db_pool %>


ekohl:
Note I didn't test this at all. This is here mostly for review and discussion about the approach. Also makes it easy to test.

manifests/config.pp
```diff
@@ -35,6 +35,12 @@
     mode  => '0640',
   }
 
+  if $foreman::use_foreman_service {
+    $db_pool = max($foreman::db_pool, $foreman::foreman_service_puma_threads_max)
+  } else {
+    $foreman::db_pool
```
ehelms, line 41:
```suggestion
    $db_pool = $foreman::db_pool
```

manifests/config.pp
```diff
@@ -35,6 +35,12 @@
     mode  => '0640',
   }
 
+  if $foreman::use_foreman_service {
+    $db_pool = max($foreman::db_pool, $foreman::foreman_service_puma_threads_max)
```
ehelms, line 39: I think to be fully safe, this has to also consider `dynflow_pool_size` if for some reason that was set higher than the max puma threads we would need to set the db_pool to match it.

evgeni: should that be `max(db_pool, puma, dynflow)` or `max(db_pool, puma+dynflow)`?

ekohl: What I've understood from @adamruzicka is that dynflow uses [sequel](https://rubygems.org/gems/sequel) and thus its own pool. However, it copies the config from Rails. Perhaps also the pool but I'm sure he can correct me.

I think the pool size was actually unused btw.

adamruzicka: `dynflow_pool_size` shouldn't be used for anything since we moved sidekiq and even when it was previously used, dynflow knew how to increase the number of connections in the pool to work no matter which value was configured there. We can safely ignore it here

ehelms: Right now, `dynflow_pool_size` determines the concurrency for workers (https://github.com/theforeman/puppet-foreman/blob/master/manifests/dynflow/worker.pp#L25). If this concurrency value was set higher than the db_pool configured in the Rails `database.yml`, would that imply that a worker could need more connections than we configure it for @adamruzicka ?

adamruzicka: Ah, forgot it is being used for that.

With https://github.com/Dynflow/dynflow/pull/362 the worker calculates how many connections it should need in the worst case and then it reconfigures its pool to have `max($calculated, $db_pool_size_from_database.yml)` connections available

ehelms: Ahh, so behind the scenes the worker could calculate a value that exceeds the `db_pool` configured through Rails, but the workers are using Sequel anyway so they are not bound by the db_pool defined in Rails?

adamruzicka: > but the workers are using Sequel anyway so they are not bound by the db_pool defined in Rails?

Not exactly. They are not bound by `db_pool`, but not because they are using Sequel.

The worker boots, takes AR configuration (what is set in `db_pool`), calculates how much connections it thinks will need, picks the bigger number, overrides the original AR configuration with the new value and uses that. Then it takes the new configuration for AR and derives configuration for Sequel from it.

manifests/config.pp
ekohl: Fixed

ehelms (approved):

PR #883, opened 2020-09-08
Fixes #30803: Bind to socket for Puma and Apache

ehelms:
The goal of this change is to switch to using local unix sockets to bind Puma to, and also to have Apache reverse proxy to increasing the security of the communication from the reverse proxy to the socket. 

I am looking to understand if I need to enforce permissions and user-group on the socket to enhance security. There is also currently SELinux denials that are associated with this change we will need to account for. On my test system they are:

```
type=AVC msg=audit(1599591946.533:10070): avc:  denied  { connectto } for  pid=6405 comm="httpd" path="/run/puma.sock" scontext=system_u:system_r:httpd_t:s0 tcontext=system_u:system_r:foreman_rails_t:s0 tclass=unix_stream_socket
```

SELinux policy changes needed to address -- https://github.com/theforeman/foreman-selinux/pull/113

### DIFF
--- a/manifests/config.pp
+++ b/manifests/config.pp
@@ -53,12 +53,13 @@ class foreman::config {
     ensure => absent,
   }
 
-  $listen_socket = $foreman::foreman_service_bind ? {
-    Stdlib::IP::Address::V6 => "[${foreman::foreman_service_bind}]:${foreman::foreman_service_port}",
-    default                 => "${foreman::foreman_service_bind}:${foreman::foreman_service_port}",
-  }
-
   if $foreman::use_foreman_service {
+    if $foreman::apache {
+      include apache
+    }
+
+    $listen_stream = regsubst($foreman::foreman_service_bind, 'unix://|tcp://', '')
+
     systemd::dropin_file { 'foreman-socket':
       filename => 'installer.conf',
       unit     => "${foreman::foreman_service}.socket",
@@ -125,7 +126,7 @@ class foreman::config {
       serveraliases           => $foreman::serveraliases,
       server_port             => $foreman::server_port,
       server_ssl_port         => $foreman::server_ssl_port,
-      proxy_backend           => "http://${listen_socket}/",
+      proxy_backend           => $foreman::foreman_service_bind,
       ssl                     => $foreman::ssl,
       ssl_ca                  => $foreman::server_ssl_ca,
       ssl_chain               => $foreman::server_ssl_chain,
--- a/manifests/config/apache.pp
+++ b/manifests/config/apache.pp
@@ -105,7 +105,7 @@ class foreman::config::apache(
   Array[Stdlib::Fqdn] $serveraliases = [],
   Stdlib::Port $server_port = 80,
   Stdlib::Port $server_ssl_port = 443,
-  Stdlib::HTTPUrl $proxy_backend = 'http://localhost:3000/',
+  Pattern['^(https?|unix)://'] $proxy_backend = 'unix:///run/foreman.sock',
   Hash $proxy_params = {'retry' => '0'},
   Array[String] $proxy_no_proxy_uris = ['/pulp', '/pulp2', '/streamer', '/pub', '/icons'],
   Boolean $ssl = false,
@@ -133,6 +133,12 @@ class foreman::config::apache(
   String[1] $keycloak_realm = 'ssl-realm',
 ) {
   $docroot = "${app_root}/public"
+
+  if $proxy_backend =~ 'unix://' {
+    $_proxy_backend = "${proxy_backend}|http://${servername}/"
+  } else {
+    $_proxy_backend = regsubst($proxy_backend, 'tcp://', 'http://')
+  }
 
   if $foreman_url {
     $suburi_parts = split($foreman_url, '/')
@@ -186,7 +192,7 @@ class foreman::config::apache(
     }
 
     include apache::mod::proxy_wstunnel
-    $websockets_backend = regsubst($proxy_backend, 'http://', 'ws://')
+    $websockets_backend = regsubst($_proxy_backend, 'http://', 'ws://')
 
     $vhost_http_internal_options = {
       'proxy_preserve_host' => true,
@@ -205,7 +211,7 @@ class foreman::config::apache(
       'proxy_pass'          => {
         'no_proxy_uris' => $proxy_no_proxy_uris,
         'path'          => pick($suburi, '/'),
-        'url'           => $proxy_backend,
+        'url'           => $_proxy_backend,
         'params'        => $proxy_params,
       },
       'rewrites'            => [
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -325,10 +325,10 @@ class foreman (
 
   if $apache {
     $use_foreman_service = ! $passenger
-    $foreman_service_bind = '127.0.0.1'
+    $foreman_service_bind = 'unix:///run/foreman.sock'
   } else {
     $use_foreman_service = true
-    $foreman_service_bind = '0.0.0.0'
+    $foreman_service_bind = 'tcp://0.0.0.0:3000'
   }
 
   include foreman::install
--- a/manifests/service.pp
+++ b/manifests/service.pp
@@ -40,9 +40,15 @@ class foreman::service(
   }
 
   if $use_foreman_service {
-    service { $foreman_service:
+    service { "${foreman_service}.socket":
       ensure => $foreman_service_ensure,
       enable => $foreman_service_enable,
+    }
+
+    service { $foreman_service:
+      ensure  => $foreman_service_ensure,
+      enable  => $foreman_service_enable,
+      require => Service["${foreman_service}.socket"],
     }
   }
 }
--- a/templates/foreman.service-overrides.erb
+++ b/templates/foreman.service-overrides.erb
@@ -3,7 +3,6 @@ User=<%= scope['foreman::user'] %>
 Environment=FOREMAN_ENV=<%= scope['foreman::rails_env'] %>
 Environment=FOREMAN_HOME=<%= scope['foreman::app_root'] %>
 Environment=FOREMAN_BIND=<%= scope['foreman::foreman_service_bind'] %>
-Environment=FOREMAN_PORT=<%= scope['foreman::foreman_service_port'] %>
 Environment=FOREMAN_PUMA_THREADS_MIN=<%= scope['foreman::foreman_service_puma_threads_min'] %>
 Environment=FOREMAN_PUMA_THREADS_MAX=<%= scope['foreman::foreman_service_puma_threads_max'] %>
 Environment=FOREMAN_PUMA_WORKERS=<%= scope['foreman::foreman_service_puma_workers'] %>
--- a/templates/foreman.socket-overrides.erb
+++ b/templates/foreman.socket-overrides.erb
@@ -1,3 +1,7 @@
 [Socket]
 ListenStream=
-ListenStream=<%= @listen_socket %>
+ListenStream=<%= @listen_stream %>
+<% if scope['foreman::apache'] -%>
+SocketUser=<%= scope['apache::user'] %>
+SocketMode=0600
+<% end -%>
--- a/manifests/params.pp
+++ b/manifests/params.pp
@@ -74,7 +74,6 @@ class foreman::params {
   $foreman_service = 'foreman'
   $foreman_service_ensure = 'running'
   $foreman_service_enable = true
-  $foreman_service_port = 3000
   $foreman_service_puma_threads_min = 0
   $foreman_service_puma_threads_max = 16
   $foreman_service_puma_workers = 2
--- a/spec/classes/foreman_config_apache_spec.rb
+++ b/spec/classes/foreman_config_apache_spec.rb
@@ -229,14 +229,14 @@ describe 'foreman::config::apache' do
                 .with_proxy_pass(
                   "no_proxy_uris" => ['/pulp', '/pulp2', '/streamer', '/pub', '/icons'],
                   "path"          => '/',
-                  "url"           => 'http://localhost:3000/',
+                  "url"           => 'unix:///run/foreman.sock|http://foo.example.com/',
                   "params"        => { "retry" => '0' },
                 )
                 .with_rewrites([
                   {
                     'comment'      => 'Upgrade Websocket connections',
                     'rewrite_cond' => '%{HTTP:Upgrade} =websocket [NC]',
-                    'rewrite_rule' => '/(.*) ws://localhost:3000/$1 [P,L]',
+                    'rewrite_rule' => '/(.*) unix:///run/foreman.sock|ws://foo.example.com/$1 [P,L]',
                   },
                 ])
             end
@@ -261,14 +261,14 @@ describe 'foreman::config::apache' do
                 .with_proxy_pass(
                   "no_proxy_uris" => ['/pulp', '/pulp2', '/streamer', '/pub', '/icons'],
                   "path"          => '/',
-                  "url"           => 'http://localhost:3000/',
+                  "url"           => 'unix:///run/foreman.sock|http://foo.example.com/',
                   "params"        => { "retry" => '0' },
                 )
                 .with_rewrites([
                   {
                     'comment'      => 'Upgrade Websocket connections',
                     'rewrite_cond' => '%{HTTP:Upgrade} =websocket [NC]',
-                    'rewrite_rule' => '/(.*) ws://localhost:3000/$1 [P,L]',
+                    'rewrite_rule' => '/(.*) unix:///run/foreman.sock|ws://foo.example.com/$1 [P,L]',
                   },
                 ])
             end
--- a/spec/classes/foreman_spec.rb
+++ b/spec/classes/foreman_spec.rb
@@ -154,8 +154,8 @@ describe 'foreman' do
 
         it { should compile.with_all_deps }
         it { should contain_class('foreman::config::apache').with_passenger(false) }
-        it { should contain_systemd__dropin_file('foreman-socket').with_filename('installer.conf').with_unit('foreman.socket').with_content(/^ListenStream=127\.0\.0\.1:3000$/) }
-        it { should contain_systemd__dropin_file('foreman-service').with_filename('installer.conf').with_unit('foreman.service').with_content(/^Environment=FOREMAN_BIND=127.0.0.1$/) }
+        it { should contain_systemd__dropin_file('foreman-socket').with_filename('installer.conf').with_unit('foreman.socket').with_content(/^ListenStream=\/run\/foreman\.sock$/) }
+        it { should contain_systemd__dropin_file('foreman-service').with_filename('installer.conf').with_unit('foreman.service').with_content(/^Environment=FOREMAN_BIND=unix:\/\/\/run\/foreman\.sock$/) }
         it do
           should contain_concat__fragment('foreman_settings+01-header.yaml')
             .with_content(/^:ssl_client_dn_env: HTTP_SSL_CLIENT_S_DN$/)
--- a/spec/support/acceptance/examples.rb
+++ b/spec/support/acceptance/examples.rb
@@ -19,8 +19,8 @@ shared_examples 'the foreman application' do
     it { is_expected.to be_listening }
   end
 
-  describe port(3000) do
-    it { is_expected.to be_listening.on('127.0.0.1').with('tcp') }
+  describe file('/run/foreman.sock') do
+    it { should be_socket }
   end
 
   describe command("curl -s --cacert /etc/foreman-certs/certificate.pem https://#{host_inventory['fqdn']} -w '\%{redirect_url}' -o /dev/null") do


ehelms:
Combines with https://github.com/theforeman/foreman/pull/7972

ehelms:
Note: I expect tests to fail but I want feedback on the approach before I do a bunch of updates.

templates/foreman.service-overrides.erb
```diff
@@ -3,7 +3,9 @@ User=<%= scope['foreman::user'] %>
 Environment=FOREMAN_ENV=<%= scope['foreman::rails_env'] %>
 Environment=FOREMAN_HOME=<%= scope['foreman::app_root'] %>
 Environment=FOREMAN_BIND=<%= scope['foreman::foreman_service_bind'] %>
+<%- if scope['foreman::foreman_service_port'] %>
 Environment=FOREMAN_PORT=<%= scope['foreman::foreman_service_port'] %>
+<%- end %>
```
adamruzicka, line 8: If I recall correctly, we have `FOREMAN_PORT=3000` straight in the service unit file and this file contains just overrides. So if `$foreman_service_port` is undefined, the default will be used. How does puma handle having bind set to a path and port to a number?

evgeni: My reading of https://github.com/puma/puma/blob/v4.3.3/lib/puma/binder.rb#L89-L175 says `puma` doesn't care about a port when it's doing `unix` sockets. However I remember @ehelms having fun with `rails server` yesterday, so it's probably good to test.

evgeni: But this will be interesting, see https://github.com/theforeman/foreman/pull/7972#discussion_r485439002

manifests/config.pp
```diff
@@ -119,7 +123,7 @@
       serveraliases           => $foreman::serveraliases,
       server_port             => $foreman::server_port,
       server_ssl_port         => $foreman::server_ssl_port,
-      proxy_backend           => "http://${listen_socket}/",
+      proxy_backend           => "${foreman_service_bind}|http://127.0.0.1/",
```
evgeni, line 126: This is missing a port?

ehelms: As I understand it, this is required but is only used to determine the Host value sent with the request:

> In 2.4.7 and later, support for using a Unix Domain Socket is available by using a target which prepends unix:/path/lis.sock|. For example, to proxy HTTP and target the UDS at /home/www.socket, you would use unix:/home/www.socket|http://localhost/whatever/. Since the socket is local, the hostname used (in this case localhost) is moot, but it is passed as the Host: header value of the request.

templates/foreman.service-overrides.erb
ehelms: If you look at the related PR (https://github.com/theforeman/foreman/pull/7972) the idea is to drop the default setting of those environment variables in the systemd file and rely on defaults in the Puma config or defaults from Puma itself if a user deploys this on their own without customizing through the installer.

templates/foreman.socket-overrides.erb
```diff
@@ -1,3 +1,4 @@
 [Socket]
 ListenStream=
 ListenStream=<%= @listen_socket %>
+User=httpd
```
ekohl, line 4: This needs to be a variable. On Debian the user is different. You can use `$apache::user` for that value. I also think this should be behind an if and only do so when Apache is configured. We can also narrow down the permissions.
```suggestion
<% if scope['foreman::apache'] -%>
User=<%=scope['apache::user'] %>
Mode=0700
<% end -%>
```

templates/foreman.service-overrides.erb
ekohl: I think we can set it unconditionally since `foreman.service` also sets it hardcoded. If you don't want a value, you must clear it.

ehelms: Would that mean you end up with:

```
Environment=FOREMAN_PORT=undef
```

ekohl: I shouldn't do such superficial code reviews. Didn't see it could be undef.

However, thinking more about it (and spending some time on this today) I have some more thoughts.

It kind of depends on how you want to deal with the Foreman side. One option is to change things around. Rather than exposing a `FOREMAN_BIND` (which is right now really `FOREMAN_HOST`) and a `FOREMAN_PORT`, perhaps we want to expose `FOREMAN_BIND` as what Puma accepts. So `tcp://127.0.0.1:3000` or `unix:///run/foreman.sock`.

So I'm suggesting to consider getting rid of `FOREMAN_PORT` altogether.

manifests/config.pp
evgeni: Ah, okay!

evgeni: But should it be localhost then, or the host the request came in on?

templates/foreman.service-overrides.erb
ehelms: I think that's OK. This puts more onus on the user who wants to customize and configure to know the exact right combination rather than specifying an IP and Port which feels a bit more natural from a user input perspective. However, I really only expect container based users to really get in here and change these values.

ekohl: On the other hand, it allows users to tweak the binding. Puma can also bind on HTTPS. This allows users to run Foreman without Apache over HTTPS.

And containers don't use systemd so that's mostly irrelevant.

ekohl (commented):
Since the current implementation depends on a Foreman change, we'll need to be careful about compatibility. Right now the module is compatible with Foreman 2.0, as specified in README.md. We want to support current stable + development.

Ideally we'd write any Foreman patch in such a way that it's compatible with both the old `FOREMAN_BIND` (without a protocol + port) and the new (with). Then cherry pick that into a stable branch. In the compatibility notes we'd then write that you need at least Foreman 2.1.x (2.1.2 at the time of writing?). I'd also like to cherry pick this specific patch to 2.1 when it's finished.

manifests/init.pp
```diff
@@ -320,10 +320,10 @@
 
   if $apache {
     $use_foreman_service = ! $passenger
-    $foreman_service_bind = '127.0.0.1'
+    $foreman_service_bind = 'unix:///run/puma.sock'
```
ehelms, line 323: One open question was whether this path sufficed or we should put it under `/run/httpd` or `/run/foreman`

manifests/config.pp
ehelms: > Since the current implementation depends on a Foreman change, we'll need to be careful about compatibility. Right now the module is compatible with Foreman 2.0, as specified in README.md. We want to support current stable + development.
> 
> Ideally we'd write any Foreman patch in such a way that it's compatible with both the old `FOREMAN_BIND` (without a protocol + port) and the new (with). Then cherry pick that into a stable branch. In the compatibility notes we'd then write that you need at least Foreman 2.1.x (2.1.2 at the time of writing?). I'd also like to cherry pick this specific patch to 2.1 when it's finished.

I'm not clear what you are asking for here. This change and the [foreman change](https://github.com/theforeman/foreman/pull/7972) go hand-in-hand and both would need to get back ported at the same time.

templates/foreman.socket-overrides.erb
```diff
@@ -1,3 +1,7 @@
 [Socket]
 ListenStream=
-ListenStream=<%= @listen_socket %>
+ListenStream=<%= @foreman_service_bind %>
+<% if scope['foreman::apache'] -%>
+SocketUser=<%=scope['apache::user'] %>
```
ehelms, line 5: When I run this in the installer, this value is always blank.

manifests/config.pp
```diff
@@ -110,6 +105,8 @@
   }
 
   if $foreman::apache  {
+    User[$apache::user] -> Systemd::Dropin_file['foreman-socket']
```
ehelms, line 108: The line at https://github.com/theforeman/puppet-foreman/pull/883/files#diff-d6542e555b7c0207c63c969d650e1036R5 always ends up blank when I run this and so I was attempting to define a relationship that would ensure it. This did not seem to work either.

manifests/config/apache.pp
```diff
@@ -105,7 +105,7 @@
   Array[Stdlib::Fqdn] $serveraliases = [],
   Stdlib::Port $server_port = 80,
   Stdlib::Port $server_ssl_port = 443,
-  Stdlib::HTTPUrl $proxy_backend = 'http://localhost:3000/',
+  String $proxy_backend = 'http://localhost:3000/',
```
ekohl, line 108: I'd prefer this to be a bit stricter and think this should work:
```suggestion
  Pattern['^(https?|unix)://'] $proxy_backend = 'http://localhost:3000/',
```

ehelms:
Unit tests are now passing but the acceptance tests won't pass until https://github.com/theforeman/foreman/pull/7972 is available as an RPM

manifests/config/apache.pp
```diff
@@ -105,7 +105,7 @@
   Array[Stdlib::Fqdn] $serveraliases = [],
   Stdlib::Port $server_port = 80,
   Stdlib::Port $server_ssl_port = 443,
-  Stdlib::HTTPUrl $proxy_backend = 'http://localhost:3000/',
+  Pattern['^(http?|unix)://'] $proxy_backend = 'unix:///run/foreman.sock|http://localhost/',
```
ekohl, line 108:
```suggestion
  Pattern['^(https?|unix)://'] $proxy_backend = 'unix:///run/foreman.sock|http://localhost/',
```

ekohl (changes requested):

manifests/config/apache.pp
ehelms: Why `https` here? Doesn't that rule out binding to `0.0.0.0` or `localhost` without https? (i.e. the way we were deploying previously)

Or are you trying to enforce either deploying via unix socket or running foreman service bound to localhost/0.0.0.0 with Puma performing SSL?

ekohl: If you don't want to allow https, then you can leave out `?`
```suggestion
  Pattern['^(http|unix)://'] $proxy_backend = 'unix:///run/foreman.sock|http://localhost/',
```

However, a HTTPS backend is technically valid and IMHO the data type shouldn't forbid it. So while it's unlikely, I'd just allow it in the data type.

ekohl:
My only concern now is the compatibility. Per https://github.com/theforeman/puppet-foreman#support-policy we support multiple versions and this now only works with develop. In https://github.com/theforeman/foreman/pull/7972#issuecomment-701350350 I asked to cherry pick this to stable branches.

The reason we support multiple versions is that the common upgrade procedure is to first update all your Puppet modules and then the Foreman version itself.

ehelms:
> My only concern now is the compatibility. Per https://github.com/theforeman/puppet-foreman#support-policy we support multiple versions and this now only works with develop.

We have no testing that guarantees this. I'm not saying that invalidates what the README says, just that if we do aim to meet this constraining requirement that we should expand our matrix to have our acceptance tests meet this case. Having the master branch of the module support develop *and* the current stable release seems like it could get mucky at times (e.g. a major Foreman version bump).

I am not as a familiar with the puppet workflows to know why somebody would fully upgrade their modules and then Foreman itself without doing them in tandem. Can you share?

ekohl:
> We have no testing that guarantees this. I'm not saying that invalidates what the README says, just that if we do aim to meet this constraining requirement that we should expand our matrix to have our acceptance tests meet this case. Having the master branch of the module support develop _and_ the current stable release seems like it could get mucky at times (e.g. a major Foreman version bump).

Yes, it would be good to have this testing but it's a policy we've kept for many years. I have often used git master versions on my stable Foreman release. The fact we've gotten issues/PRs in the past when we did break things shows it's not just me who does this. I strongly value that we keep this.



> I am not as a familiar with the puppet workflows to know why somebody would fully upgrade their modules and then Foreman itself without doing them in tandem. Can you share?

It's common to use the latest version of the modules with a stable version of the RPMs. I can only speak for myself, but this allowed me to send in a PR with a fix without having to wait for some backport. Another example may be that you want to split up the upgrade process. Updating may include updating the Apache and PostgreSQL modules as well so you first verify that doesn't break anything.

ehelms:
When I test a clean install, along with the selinux patch, I get a clean install. When I test an upgrade I run into an issue related to the ownership of the `foreman.socket` file that has me stumped.

On a clean install, the `foreman.sock` file gets created as:

```
[root@pipeline-up-foreman-nightly-centos7 vagrant]# ls -la /run/foreman.sock 
srwx------. 1 apache apache 0 Oct  1 18:41 /run/foreman.sock
```

On upgrade, this gets created as:

```
[root@pipeline-up-foreman-nightly-centos7 vagrant]# ls -la /run/foreman.sock 
srwx------. 1 root root 0 Oct  1 18:41 /run/foreman.sock
```

The upgrade fails as Apache does not have permission to see the socket file. My guess is this is some bit of ordering between whether `foreman.socket` or `foreman.service` creates the file first. But that is pure speculation. Looking for some help and ideas as to how to help mitigate this.

ekohl:
I think the problem is that we relabel in `%posttrans`:

https://github.com/theforeman/foreman-packaging/blob/475a1b24fd683e7171d19634ee16419bdbcf6f1a/packages/foreman/foreman-selinux/foreman-selinux.spec#L104-L108

However, the service restarts in `%postun`:

https://github.com/theforeman/foreman-packaging/blob/475a1b24fd683e7171d19634ee16419bdbcf6f1a/packages/foreman/foreman/foreman.spec#L1015-L1017

I think that means the new SELinux policy is loaded after the service is restarted. However, this is just a theory.

Another potential problem is the alias rules in SELinux that we've seen some issues with. `/run` is set to an equivalent of `/var/run` and perhaps that's also playing a role here?

Another theory is that on restart the socket isn't recreated because the socket is never restarted. The RPM package doesn't and the Puppet module doesn't manage `foreman.socket`.

manifests/service.pp
```diff
@@ -44,5 +44,10 @@
       ensure => $foreman_service_ensure,
       enable => $foreman_service_enable,
     }
+
+    service { "${foreman_service}.socket":
```
ehelms, line 48: Adding management of the socket here seems to have done the trick in my upgrade testing. @ekohl is there any reason not to add this?

ekohl: Initially I left it out to remain compatible with older versions which didn't ship the `.socket` file. Back then it wasn't as relevant, now it is. I think we started to ship it in 2.1 so it deserves a note in the compatibility section that 2.0 is no longer supported (technically users could ship it themselves rather than in the RPMs, but I never went that far in documenting).

Note the **think** part. I didn't check if it was 2.1. Perhaps it was 2.0

ehelms: Checking git history, it was 2.1 that it was released as part of.

ehelms: You know me I'm all for the main branch reflecting the state of the art in the project; however, I will at least ask if this is the kind of thing that warrants a fact or some other method to conditionally wrap it based on Foreman version?

ehelms:
This fails only on Ubuntu1804 with:

```
  Oct 14 20:07:40 ubuntu1804-64.example.com foreman[37139]: Exiting
  Oct 14 20:07:40 ubuntu1804-64.example.com foreman[37139]: /usr/share/foreman/vendor/ruby/2.5.0/gems/puma-4.3.6/lib/puma/binder.rb:320:in `unlink': Permission denied @ apply2files - /run/foreman.sock (Errno::EACCES)
  Oct 14 20:07:40 ubuntu1804-64.example.com foreman[37139]:         from /usr/share/foreman/vendor/ruby/2.5.0/gems/puma-4.3.6/lib/puma/binder.rb:320:in `rescue in add_unix_listener'
  Oct 14 20:07:40 ubuntu1804-64.example.com foreman[37139]:         from /usr/share/foreman/vendor/ruby/2.5.0/gems/puma-4.3.6/lib/puma/binder.rb:317:in `add_unix_listener'
  Oct 14 20:07:40 ubuntu1804-64.example.com foreman[37139]:         from /usr/share/foreman/vendor/ruby/2.5.0/gems/puma-4.3.6/lib/puma/binder.rb:151:in `block in parse'
  Oct 14 20:07:40 ubuntu1804-64.example.com foreman[37139]:         from /usr/share/foreman/vendor/ruby/2.5.0/gems/puma-4.3.6/lib/puma/binder.rb:90:in `each'
  Oct 14 20:07:40 ubuntu1804-64.example.com foreman[37139]:         from /usr/share/foreman/vendor/ruby/2.5.0/gems/puma-4.3.6/lib/puma/binder.rb:90:in `parse'
  Oct 14 20:07:40 ubuntu1804-64.example.com foreman[37139  ]:         from /usr/share/foreman/vendor/ruby/2.5.0/gems/puma-4.3.6/lib/puma/runner.rb:161:in `load_and_bind'
  Oct 14 20:07:40 ubuntu1804-64.example.com foreman[37139]:         from /usr/share/foreman/vendor/ruby/2.5.0/gems/puma-4.3.6/lib/puma/cluster.rb:413:in `run'
  Oct 14 20:07:40 ubuntu1804-64.example.com foreman[37139]:         from /usr/share/foreman/vendor/ruby/2.5.0/gems/puma-4.3.6/lib/puma/launcher.rb:172:in `run'
  Oct 14 20:07:40 ubuntu1804-64.example.com foreman[37139]:         from /usr/share/foreman/vendor/ruby/2.5.0/gems/puma-4.3.6/lib/rack/handler/puma.rb:73:in `run'
  Oct 14 20:07:40 ubuntu1804-64.example.com foreman[37139]:         from /usr/share/foreman/vendor/ruby/2.5.0/gems/rack-2.2.3/lib/rack/server.rb:327:in `start'
  Oct 14 20:07:40 ubuntu1804-64.example.com foreman[37139]:         from /usr/share/foreman/vendor/ruby/2.5.0/gems/railties-6.0.3.4/lib/rails/commands/server/server_command.rb:39:in `start'
  Oct 14 20:07:40 ubuntu1804-64.example.com foreman[37139]:         from /usr/share/foreman/vendor/ruby/2.5.0/gems/railties-6.0.3.4/lib/rails/commands/server/server_command.rb:147:in `block in perform'
  Oct 14 20:07:40 ubuntu1804-64.example.com foreman[37139]:         from /usr/share/foreman/vendor/ruby/2.5.0/gems/railties-6.0.3.4/lib/rails/commands/server/server_command.rb:138:in `tap'
  Oct 14 20:07:40 ubuntu1804-64.example.com foreman[37139]:         from /usr/share/foreman/vendor/ruby/2.5.0/gems/railties-6.0.3.4/lib/rails/commands/server/server_command.rb:138:in `perform'
  Oct 14 20:07:40 ubuntu1804-64.example.com foreman[37139]:         from /usr/share/foreman/vendor/ruby/2.5.0/gems/thor-1.0.1/lib/thor/command.rb:27:in `run'
  Oct 14 20:07:40 ubuntu1804-64.example.com foreman[37139]:         from /usr/share/foreman/vendor/ruby/2.5.0/gems/thor-1.0.1/lib/thor/invocation.rb:127:in `invoke_command'
  Oct 14 20:07:40 ubuntu1804-64.example.com foreman[37139]:         from /usr/share/foreman/vendor/ruby/2.5.0/gems/thor-1.0.1/lib/thor.rb:392:in `dispatch'
  Oct 14 20:07:40 ubuntu1804-64.example.com foreman[37139]:         from /usr/share/foreman/vendor/ruby/2.5.0/gems/railties-6.0.3.4/lib/rails/command/base.rb:69:in `perform'
  Oct 14 20:07:40 ubuntu1804-64.example.com foreman[37139]:         from /usr/share/foreman/vendor/ruby/2.5.0/gems/railties-6.0.3.4/lib/rails/command.rb:46:in `invoke'
  Oct 14 20:07:40 ubuntu1804-64.example.com foreman[37139]:         from /usr/share/foreman/vendor/ruby/2.5.0/gems/railties-6.0.3.4/lib/rails/commands.rb:18:in `<top (required)>'
  Oct 14 20:07:40 ubuntu1804-64.example.com foreman[37139]:         from /usr/share/foreman/vendor/ruby/2.5.0/gems/activesupport-6.0.3.4/lib/active_support/dependencies.rb:324:in `require'
  Oct 14 20:07:40 ubuntu1804-64.example.com foreman[37139]:         from /usr/share/foreman/vendor/ruby/2.5.0/gems/activesupport-6.0.3.4/lib/active_support/dependencies.rb:324:in `block in require'
  Oct 14 20:07:40 ubuntu1804-64.example.com foreman[37139]:         from /usr/share/foreman/vendor/ruby/2.5.0/gems/activesupport-6.0.3.4/lib/active_support/dependencies.rb:291:in `load_dependency'
  Oct 14 20:07:40 ubuntu1804-64.example.com foreman[37139]:         from /usr/share/foreman/vendor/ruby/2.5.0/gems/activesupport-6.0.3.4/lib/active_support/dependencies.rb:324:in `require'
  Oct 14 20:07:40 ubuntu1804-64.example.com foreman[37139]:         from /usr/share/foreman/bin/rails:9:in `<main>'
```

manifests/config/apache.pp
```diff
@@ -105,7 +105,7 @@
   Array[Stdlib::Fqdn] $serveraliases = [],
   Stdlib::Port $server_port = 80,
   Stdlib::Port $server_ssl_port = 443,
-  Stdlib::HTTPUrl $proxy_backend = 'http://localhost:3000/',
+  Pattern['^(https?|unix)://'] $proxy_backend = 'unix:///run/foreman.sock|http://localhost/',
```
ekohl, line 108: I was thinking about this version:
```suggestion
  Variant[Stdlib::HTTPUrl, Stdlib::Absolutepath] $proxy_backend = 'http://localhost:3000/',
```

Then later inside the class
```puppet
if $proxy_backend =~ Stdlib::Absolutepath {
  $_proxy_backend = "unix://${proxy_backend}|http://${servername}/"
} else {
  $_proxy_backend = $proxy_backend
}
```

However, that probably doesn't work because you still need the `unix://` version to configure Puma.

Perhaps we still need to roughly take that approach and convert it to the Apache or Puma format as close to the service as possible? Then you would also need to do a similar thing in `config.pp` for `$listen_stream`.

After I finish https://github.com/puma/puma/pull/2362 this will be easier.

ehelms: Let me know what you think about the updated design I pushed

manifests/service.pp
```diff
@@ -40,9 +40,15 @@
...
   if $use_foreman_service {
+    service { "${foreman_service}.socket":
+      ensure => $foreman_service_ensure,
+      enable => $foreman_service_enable,
+    }
+
     service { $foreman_service:
       ensure => $foreman_service_ensure,
       enable => $foreman_service_enable,
+      require => Service["${foreman_service}.socket"],
```
ehelms, line 51: This seems to have fixed the Ubuntu18 issue when running the acceptance test locally.

ehelms:
Oh good, fix Ubntu1804, break Debian 10.

ehelms:
The failure on debian10 started as soon as I added the requires on `foreman.service` for `foreman.socket`:

```
      Last 10 lines of output were:
       	Oct 15 15:38:31 debian10-64.example.com systemd[1]: foreman.socket: Socket service foreman.service already active, refusing.
       	Oct 15 15:38:31 debian10-64.example.com systemd[1]: Failed to listen on Foreman HTTP Server Accept Sockets.
       	Oct 15 15:39:39 debian10-64.example.com systemd[1]: foreman.socket: Socket service foreman.service already active, refusing.
       	Oct 15 15:39:39 debian10-64.example.com systemd[1]: Failed to listen on Foreman HTTP Server Accept Sockets.
       	Oct 15 15:41:03 debian10-64.example.com systemd[1]: foreman.socket: Socket service foreman.service already active, refusing.
       	Oct 15 15:41:03 debian10-64.example.com systemd[1]: Failed to listen on Foreman HTTP Server Accept Sockets.
```

Given this does not happen on the other OSes I am wondering if this is a difference in systemd versions causing this.

ekohl:
Perhaps it's a problem that the old `foreman.service` already bound to port 3000 and `foreman.socket` can't bind? Just throwing things out there.

ekohl:
> Given this does not happen on the other OSes I am wondering if this is a difference in systemd versions causing this.

Unintentionally I just reproduced this on Fedora 32. I had a `foreman.socket` and a `foreman.service` which were both running. Then a `systemctl restart foreman.socket` triggered it as well.

ehelms:
> restart

Was it running localhost:3000 and then switched to Unix Socket? Or it was simply running and you restart `foreman.socket` and had it blow up?

ekohl:
From my journal:
```
Oct 20 21:02:47 wisse systemd[1267]: Starting Foreman...
Oct 20 21:02:48 wisse foreman[746290]: => Booting Puma
Oct 20 21:02:48 wisse foreman[746290]: => Rails 6.0.3.3 application starting in production
Oct 20 21:02:48 wisse foreman[746290]: => Run `rails server --help` for more startup options
Oct 20 21:02:49 wisse foreman[746290]: API controllers newer than Apipie cache! Run apipie:cache rake task to regenerate cache.
Oct 20 21:02:52 wisse foreman[746290]: Failed to load systemd plugin
Oct 20 21:02:52 wisse foreman[746290]: ["tcp://0.0.0.0:5000"]
Oct 20 21:02:52 wisse foreman[746290]: [746290] Puma starting in cluster mode...
Oct 20 21:02:52 wisse foreman[746290]: [746290] * Version 5.0.2 (ruby 2.7.2-p137), codename: Spoony Bard
Oct 20 21:02:52 wisse foreman[746290]: [746290] * Min threads: 0, max threads: 16
Oct 20 21:02:52 wisse foreman[746290]: [746290] * Environment: production
Oct 20 21:02:52 wisse foreman[746290]: [746290] * Process workers: 2
Oct 20 21:02:52 wisse foreman[746290]: [746290] * Preloading application
Oct 20 21:02:52 wisse foreman[746290]: [746290] * Closing unused activated socket: tcp:0.0.0.0:5000
Oct 20 21:02:52 wisse foreman[746290]: [746290] Use Ctrl-C to stop
Oct 20 21:02:52 wisse foreman[746290]: [746290] - Worker 0 (pid: 746296) booted, phase: 0
Oct 20 21:02:52 wisse foreman[746290]: [746290] - Worker 1 (pid: 746300) booted, phase: 0
Oct 20 21:05:20 wisse systemd[1267]: foreman.socket: Succeeded.
Oct 20 21:05:20 wisse systemd[1267]: Closed Foreman HTTP Server Accept Sockets.
Oct 20 21:05:20 wisse systemd[1267]: Stopping Foreman HTTP Server Accept Sockets.
Oct 20 21:05:20 wisse systemd[1267]: foreman.socket: Socket service foreman.service already active, refusing.
Oct 20 21:05:20 wisse systemd[1267]: Failed to listen on Foreman HTTP Server Accept Sockets.
Oct 20 21:05:20 wisse systemd[1267]: Dependency failed for Foreman.
Oct 20 21:05:20 wisse systemd[1267]: foreman.service: Job foreman.service/start failed with result 'dependency'.
Oct 20 21:05:25 wisse foreman[746290]: [746290] - Gracefully shutting down workers...
Oct 20 21:05:26 wisse foreman[746290]: Exiting
Oct 20 21:05:26 wisse systemd[1267]: foreman.service: Succeeded.
```

I was working on a different patch, but I think what's similar is that a file descriptor was open for the correct bind (`0.0.0.0:5000`) but puma didn't consume it. Perhaps you're hitting a similar thing where one is restarted before the other was really reloaded/restarted.

ehelms:
I think, from some testing, the fundamental problem I am running into is that `foreman.service` is being activated by the packaging and this seems to be exacerbated on the Debian side. Given the `foreman.socket` has control of activating the service, I think we need to move away from the service being activated to focusing more on the socket. I did some searching I am not sure where this happens in Debian packaging. @ekohl @evgeni  can one of you point me in the direction of this?

evgeni:
I think this happens here:
https://github.com/theforeman/foreman-packaging/blob/deb/develop/debian/buster/foreman/rules#L36

docs: https://manpages.debian.org/testing/debhelper/dh_systemd_enable.1.en.html

I am not sure if passing `--no-enable` is sufficient. Probably.

evgeni:
However, I don't exactly understand what the problem with the service vs the socket is.
The service *can* bind the port itself, or get activated by `foreman.socket`, both are valid scenarios and starting the socket while the service is running should be effectively a noop?

evgeni:
Maybe we want `dh_systemd_start --no-start`? https://manpages.debian.org/testing/debhelper/dh_systemd_start.1.en.html

ehelms:
So, I think what happens is:

 * foreman package installs
 * triggers activation of foreman.service running with a bind to `host:port`
 * puppet runs configuring the foreman.socket
 * puppet attempts to start the foreman.socket which tries to activate the foreman.service which is already activated with a different bind
 * puppet fails

The annoying part is, on Debian 10 the above happens *if* the `foreman.service` Service requires `foreman.socket` Service. If I drop the `requires`, then this fails on Ubuntu 1804. This never fails on EL7.

ekohl:
I'm starting to wonder if we should wait for Puma 5.1.0 which should include https://github.com/puma/puma/pull/2362. That will simplify the configuration required. However, it may complicate support for older versions.

ehelms:
I think Puma 5.1.0 will clean things up but @ekohl you have expressed the hardening this provides and I think that's a good thing to have. I am less familiar in this area, but if I look at how EL and Debian states are post install, pre-installer run, we have variation there and I think we'd benefit from commonality:

On EL7, both `foreman.service` and `foreman.socket` come installed with `foreman-service` package. Both are installed from packaging as loaded and inactive:
```
[root@pipeline-foreman-server-nightly-centos7 vagrant]# yum install foreman-service -y
[root@pipeline-foreman-server-nightly-centos7 vagrant]# systemctl status foreman.service
● foreman.service - Foreman
   Loaded: loaded (/usr/lib/systemd/system/foreman.service; disabled; vendor preset: disabled)
   Active: inactive (dead)
     Docs: https://theforeman.org
[root@pipeline-foreman-server-nightly-centos7 vagrant]# systemctl status foreman.socket
● foreman.socket - Foreman HTTP Server Accept Sockets
   Loaded: loaded (/usr/lib/systemd/system/foreman.socket; disabled; vendor preset: disabled)
   Active: inactive (dead)
   Listen: 0.0.0.0:3000 (Stream)
```

On Debian 10 and Ubuntu 18.04, the `foreman.service` comes with the install of `foreman`, and the `foreman.socket` comes with the install of `foreman-service` package. The `foreman.service` is loaded and active, the `foreman.socket` is loaded and active :
```
root@pipeline-foreman-server-nightly-debian10:/home/vagrant# systemctl status foreman.service
● foreman.service - Foreman
   Loaded: loaded (/lib/systemd/system/foreman.service; enabled; vendor preset: enabled)
   Active: active (exited) since Thu 2020-10-22 12:28:17 UTC; 3min 21s ago
     Docs: https://theforeman.org
    Tasks: 0 (limit: 4915)
   Memory: 0B
   CGroup: /system.slice/foreman.service

Oct 22 12:28:17 pipeline-foreman-server-nightly-debian10 systemd[1]: Starting foreman.service...
Oct 22 12:28:17 pipeline-foreman-server-nightly-debian10 foreman[31396]: foreman not configured to start. Please edit /etc/default/foreman to enable.
Oct 22 12:28:17 pipeline-foreman-server-nightly-debian10 systemd[1]: Started foreman.service.
Oct 22 12:29:44 pipeline-foreman-server-nightly-debian10 systemd[1]: Dependency failed for Foreman.
Oct 22 12:29:44 pipeline-foreman-server-nightly-debian10 systemd[1]: foreman.service: Job foreman.service/start failed with result 'dependency'.
root@pipeline-foreman-server-nightly-debian10:/home/vagrant# systemctl status foreman.socket
● foreman.socket - Foreman HTTP Server Accept Sockets
   Loaded: loaded (/lib/systemd/system/foreman.socket; enabled; vendor preset: enabled)
   Active: inactive (dead)
   Listen: 0.0.0.0:3000 (Stream)

Oct 22 12:29:44 pipeline-foreman-server-nightly-debian10 systemd[1]: foreman.socket: Socket service foreman.service already active, refusing.
Oct 22 12:29:44 pipeline-foreman-server-nightly-debian10 systemd[1]: Failed to listen on Foreman HTTP Server Accept Sockets.
Oct 22 12:29:44 pipeline-foreman-server-nightly-debian10 systemd[1]: foreman.socket: Socket service foreman.service already active, refusing.
Oct 22 12:29:44 pipeline-foreman-server-nightly-debian10 systemd[1]: Failed to listen on Foreman HTTP Server Accept Sockets.
```

ekohl:
> I think Puma 5.1.0 will clean things up but @ekohl you have expressed the hardening this provides and I think that's a good thing to have.

Yes. The ever lasting struggle.

> Maybe we want `dh_systemd_start --no-start`? https://manpages.debian.org/testing/debhelper/dh_systemd_start.1.en.html

I think it makes sense not to start the service by default. We have an installer and there are no migration scripts to create the PostgreSQL DB. While it's normally the Debian policy to automatically start services, we're sufficiently different that it makes sense not to.

I would also suggest to cherry pick that to 2.2 if it works in nightly. Not starting a service by default is the default on Red Hat based systems so the installer should handle it perfectly well.

spec/spec_helper_acceptance.rb
```diff
@@ -14,6 +14,8 @@
     on host, 'yum clean expire-cache'
   end
 
+  on host, 'export LANG=en_us.UTF-8'
```
ekohl, line 17: This will be reverted by a modulesync. Is this really needed?

ehelms: That is what I am trying to figure out in part. I have seen errors related to locale when testing and I want to see if it makes a difference locally and on Travis.

ehelms:
This PR seems to be hitting this -- https://github.com/theforeman/puppet-foreman/pull/897

ehelms:
Woohoo, the Debian 10 tests passed with the packaging updates

ehelms:
I rebased this on latest to re-run the tests given it has sat for a bit.

wbclark (approved):
I installed nightly on centos7 with this PR using Forklift. Playing around with it a bit, I can see the socket at `/run/foreman.sock` and the drop in configuration with expected values at `/etc/systemd/system/foreman.socket.d/installer.conf`(I had to read some docs to understand the purpose of the blank `ListenStream=`). I tested basic functionality and everything seems working as expected.

I couldn't install nightly on ubuntu 1804 but the failure is with `redis.service` and seems wholly unrelated to changes here.

evgeni:
why is `/run/foreman.sock` `system_u:object_r:var_run_t:s0`? `foreman-selinux` has a file context that makes it `system_u:object_r:foreman_var_run_t,s0` (and that would be allowed to be written by httpd)

evgeni:
hah, `/run/foreman\.sock                                 regular file       system_u:object_r:foreman_var_run_t:s0 ` -- well, it's not a regular file.

PR #884, opened 2020-09-25
CLI: make refresh_cache and request_timeout params global

neomilium:
Considering the name of these parameters, they should be affect system-wide configuration but were set in root configuration file.
This commit fixes this unexpected behavior.

### DIFF
--- a/spec/classes/foreman_cli_spec.rb
+++ b/spec/classes/foreman_cli_spec.rb
@@ -22,7 +22,9 @@ describe 'foreman::cli' do
             verify_exact_contents(catalogue, '/etc/hammer/cli.modules.d/foreman.yml', [
                                     ':foreman:',
                                     '  :enable_module: true',
-                                    "  :host: 'http://example.com'"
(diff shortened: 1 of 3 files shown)


neomilium:
Spec tests failed due to two reasons:
* I need to test that these parameters affect the right file (WIP)
* Some tests strictly verify that there are only the tested parameters, no more

IMHO, two possible behaviors, the configuration file can be filled by
* specified parameters only (so, I will allow `undef` on `refresh_cache` and `refresh_timeout`)
* all parameters, ie. including `refresh_cache` and `refresh_timeout` (so, I will add them to attempted configuration entries)

The question is what the foreman team prefers?

Note: If someone have a better direction, show me the way ;-)

neomilium:
@ekohl Do you have any point of view on this? Who should I reach in Foreman about this topic?

ekohl:
I would accept this change (provided tests pass).

ekohl (approved):
Thanks!

PR #885, opened 2020-10-02
Organize Foreman parameters into logical groups

ehelms:
Adds additional groupings of parameters into logical groups and reduces
the scope of the "basic" parameters to the most pertinent for default
install of Foreman. This has the benefit to wrapper projects like
Kafo that parameters can be treated in groups for input and display
purposes.

This is my attempt to organize and feedback is welcome on categories. 

I debated adding an 'Initialization parameters:' grouping for all of the `initial_` parameters.

### DIFF
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -16,7 +16,34 @@
 #
 # $initial_admin_timezone::       Initial timezone of the admin user
 #
+# $initial_organization::         Name of an initial organization
+#
+# $initial_location::             Name of an initial location
+#
+# === Database parameters:
+#
 # $db_manage::                    If enabled, will install and configure the database server on this host
+#
+# $db_host::                      Database 'production' host
+#
+# $db_port::                      Database 'production' port
+#
+# $db_database::                  Database 'production' database (e.g. foreman)
+#
+# $db_username::                  Database 'production' user (e.g. foreman)
+#
+# $db_password::                  Database 'production' password, default is randomly generated
+#
+# $db_sslmode::                   Database 'production' ssl mode
+#
+# $db_root_cert::                 Root cert used to verify SSL connection to postgres
+#
+# $db_pool::                      Database 'production' size of connection pool. When running as a reverse proxy,
+#                                 the value of `$foreman_service_puma_threads_max` is used if it's higher than `$db_pool`.
+#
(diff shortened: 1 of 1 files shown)


ekohl:
I think this will suffer from https://projects.theforeman.org/issues/29788.

ehelms:
It will suffer from it, however, the more I look at how we think about Kafo options and presenting the user, I'm not sure that's a bad thing anymore. We present way too much to users due to not being grouped. This introduces a pattern of:

 1) If a parameter is important to a user to config, it should be not grouped and top level
 2) If a parameter has a sane default and only more adanced use cases may want to configure it, it should be in a group and therefore will only show up for kafo using `--full-help`

If we prefer to group *all* items, then in Kafo I think we have to go with a design model where we specify what groups are "basic" and which are "advanced".

ekohl:
How I used to think it worked as `$group` would show up in `--help` and `Advanced $group` would show up in `--full-help`. I think that still makes sense. So then we can say that `Database parameters` would be a common operation and show in in `--help` but `Deployment parameters` should be `Advanced deployment parameters` because they're not commonly changed. I think that gives a nice balance between the two.

ehelms:
Do you know if it is possible to do sub-groups within groups?

ekohl:
https://github.com/theforeman/kafo_parsers#rdoc-syntax doesn't mention it, but RDoc does have it built in. When thinking about the interactive mode, I wonder if you want to apply the same nesting then.

I know that in https://github.com/theforeman/puppet-puppet we are reasonably close to this kind of structure (common, agent, advanced agent, server, advanced server). Perhaps that's also a good module to play around with to see the differences.

ehelms:
Honestly, I'd almost rather a module, or class in a module, define and group the parameters however they want. Then in Kafo, treat every parameter was an advanced parameter that lands under `--full-help` with a "basic options" list that shows options defined within that list under `--help`. This removes the friction between the installer and modules and how parameters are defined and let's the installer, which is a different experience, customize what and how it presents parameters at a scenario level.

ekohl:
I have also thought about separating it. For what it's worth, I opened https://tickets.puppetlabs.com/browse/PDOC-267 so we could move away from RDoc to just puppet-strings syntax. An alternative is to have a separate kafo.yaml file that groups parameters. However, such a thing is a large change in how we parse it so we shouldn't take it lightly.

PR #886, opened 2020-10-05
Set compute resource version parameter to advanced

ehelms:


ekohl (commented):
This doesn't matter. If Kafo only detects a single group, it uses that. Regardless of advanced/non-advanced. I ran into this in the past.

ehelms:
I am thinking of changing Kafo to be a bit more definitive about the naming of the groups. However, from my testing this does work because these manifests have a top level heading, e.g. https://github.com/theforeman/puppet-foreman/pull/886/files#diff-536b9a0665a4e95c6507cf50454090a8R1

That means the parser cache ends up with two headings:

```
      :groups:
        version:
        - Foreman EC2 compute resource support
        - 'Parameters:'
```

manifests/compute/foreman_compute.pp
```diff
@@ -1,7 +1,7 @@
-# Provides a common class to install foreman-compute
-# used by multiple compute resources
+# = Provides a common class to install foreman-compute
+#   used by multiple compute resources
```
ekohl, line 2: Can you use multi line statements like this? I wonder if the various parsers we apply actually understand this.

manifests/compute/rackspace.pp
```diff
@@ -2,7 +2,7 @@
 #
 # Provides support for Rackspace compute resources
 #
-# === Parameters:
+# === Advanced parameters:
```
ekohl, line 5: We should kill this file. The package was dropped in 2.1.

ekohl (commented):
Could you make it `Refs #29829` (https://projects.theforeman.org/issues/29829).

manifests/compute/foreman_compute.pp
```diff
@@ -1,7 +1,8 @@
-# Provides a common class to install foreman-compute
-# used by multiple compute resources
+# = Provides a common class to install foreman-compute
```
ekohl, line 1: Actually, this class can also be killed. Since https://github.com/theforeman/foreman-packaging/commit/88e977a58d26350b613ec68b86c93eca17d8a941 (Foreman 1.22) `foreman-compute` no longer exists.

manifests/compute/rackspace.pp
ehelms: Happening here -- https://github.com/theforeman/puppet-foreman/pull/894

ehelms:
Rebased with drop of foreman-compute and rackspace

ekohl (approved):
CI is red, 79806f709ac53e16f6f3a86db9b40e92450cb1f0 fixed that. This should be ok (famous last words) so merging.

PR #888, opened 2020-10-10
Factor out Apache to its own class

ekohl:
Prior to this, when the Apache config was modified a full database refresh was triggered. There's no need for that and this makes applying those changes faster.

I found this while making the Pulpcore changes and this held me up. For now it's untested but pushing this allows me to test it out. Until then it's a draft.


manifests/config.pp
```diff
@@ -53,12 +53,10 @@
     ensure => absent,
   }
 
-  $listen_socket = $foreman::foreman_service_bind ? {
-    Stdlib::IP::Address::V6 => "[${foreman::foreman_service_bind}]:${foreman::foreman_service_port}",
-    default                 => "${foreman::foreman_service_bind}:${foreman::foreman_service_port}",
-  }
-
   if $foreman::use_foreman_service {
```
ehelms, line 56: Is there a best practice as to why this section does not live in `service.pp` since they are tied to the service itself?

manifests/config.pp
```diff
@@ -115,118 +113,62 @@
...
-      passenger_prestart      => $foreman::passenger_prestart,
-      passenger_min_instances => $foreman::passenger_min_instances,
-      passenger_start_timeout => $foreman::passenger_start_timeout,
-      foreman_url             => $foreman::foreman_url,
-      ipa_authentication      => $foreman::ipa_authentication,
-      keycloak                => $foreman::keycloak,
-      keycloak_app_name       => $foreman::keycloak_app_name,
-      keycloak_realm          => $foreman::keycloak_realm,
+  if $foreman::apache and $foreman::ipa_authentication {
+    unless fact('foreman_ipa.default_server') and fact('foreman_ipa.default_realm') {
```
ehelms, line 117: Not for this change, but I would have us consider putting external auth into it's own class for slimming down the config class and making it more obvious what aspects are tied to external auth.

manifests/apache.pp
```diff
@@ -0,0 +1,65 @@
+# @summary The apache configuration for Foreman
+# @api private
+class foreman::apache {
```
ehelms, line 3: :+1: I like this being a clear, dedicated class

manifests/config.pp
ekohl: I think initially when I wrote this I considered them config files. Now looking back (and with the experiences from Pulpcore as well) I think `service.pp` would be a better place. For example, there's no reason this needs to be done before the DB, apache etc.

manifests/config.pp
ekohl: Probably, there's more room for improvement and I think I scratched the surface here.

ekohl:
I'll revisit this once we've decided on dropping mod_passenger.

ekohl:
Since we've dropped mod_passenger for a while, I've rebased this. It includes #935

ehelms:
Can now be rebased again

PR #889, opened 2020-10-12
Switch plugin documentation to RDOC and mark advanced parameters

ehelms:


ehelms:
Friendly ping for any maintainers since its been a few weeks without any review.

ekohl (commented):
I don't see the benefit in this except for the cases where there are advanced parameters. puppet-strings is the official way of documenting and RDoc is one that only we use via Kafo. Ideally we'd find a way to get rid of RDoc completely and standardize on puppet-strings (to follow the ecosystem) but I'm not sure how to do that at this point. That's why I've chosen to use puppet-strings everywhere except when there are advanced parameters.

ares:
Hold on, why are we closing this? Does this do any harm for outside of foreman-installer use case? I totally see the value for foreman-installer. Unless we're close to correctly parse this from puppet-strings, why do we block foreman-installer improvements (supposing the answer for the first question is no).

ekohl:
I was arguing that we should prefer puppet-strings where it doesn't make a difference and RDoc where it does. Then it was closed.

ehelms:
This was discussed partially out of band of the PR. The disconnect between different document styles makes choosing an already disjoint path forward harder. Given I cannot see how to reconcile that systematically at the moment, I left things as is.

manifests/plugin/memcache.pp
```diff
@@ -6,6 +6,8 @@
 #
 # $hosts::      an array of hosts running memcache
 #
+# === Advanced parameters:
```
ekohl, line 9: This one still makes sense

PR #890, opened 2020-10-12
Fixes #30789 - Set DB pool size dynamically

ehelms:
Every thread in the Puma worker can open a database connection. This
means it needs to be taken into account to avoid exhasuting the pool.

(cherry picked from commit 026d47434316b8ae318c5e42936edc12859ab475)

### DIFF
--- a/manifests/config.pp
+++ b/manifests/config.pp
@@ -31,6 +31,12 @@ class foreman::config {
     mode  => '0640',
   }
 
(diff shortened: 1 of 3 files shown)


ehelms:
This is requested for Foreman 2.1.4

tbrisker (approved):
Thanks @ehelms ! @ekohl can we get a module release with this change please?

PR #892, opened 2020-10-13
Fixes #31215 - Generate DSL docs

ofedoren:
@ekohl, please, let me know if I should link it to the Redmine or add/change something else.

My intention here is to add generating of DSL docs similarly to API docs after Foreman or a plugin is installed.

### DIFF
--- a/manifests/database.pp
+++ b/manifests/database.pp
@@ -31,6 +31,6 @@ class foreman::database {
     ~> foreman::rake { 'db:seed':
       environment => delete_undef_values($seed_env),
     }
-    ~> Foreman::Rake['apipie:cache:index']
+    ~> Foreman::Rake['apipie:cache:index', 'apipie_dsl:cache']
   }
 }
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -319,6 +319,10 @@ class foreman (
     timeout => 0,
   }
 
+  foreman::rake { 'apipie_dsl:cache':
+    timeout => 0,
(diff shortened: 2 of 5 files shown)


manifests/database.pp
```diff
@@ -32,5 +32,6 @@
       environment => delete_undef_values($seed_env),
     }
     ~> Foreman::Rake['apipie:cache:index']
+    ~> Foreman::Rake['apipie_dsl:cache']
```
ekohl, line 35: This can be simplified to
```suggestion
    ~> Foreman::Rake['apipie:cache:index', 'apipie_dsl:cache']
```

This assumes that an `apipie:cache:index` doesn't need to trigger an `apipie_dsl:cache`.

(same in plugin.pp)

ofedoren: Thank you, done.

ofedoren:
@ekohl, what's the status here?..

wbclark (commented):
This change worked fine for me both in testing the individual puppet module and also in foreman-installer (tested via Forklift with the module_prs feature)

@ofedoren before we merge this could you please create a Redmine Issue and update the commit message and PR title to reference the Issue # ? For example, "Fixes #XXXXX - Generate DSL docs"

ofedoren:
Thanks, @wbclark! Done.

wbclark (approved):

wbclark:
Thanks @ofedoren ! It is merged.

PR #895, opened 2020-10-19
Drop foreman_compute that was removed in 1.22

ehelms:


ekohl (approved):
I'm not going to mark this as breaking since the package was removed in 1.22 and this module targets 2.0+.

PR #896, opened 2020-10-22
Refs #30535 - Correctly unset remote user groups

tbrisker:
This corrects a typo in the previous change, which unset
REMOTE_USER_USER_GROUPS instead of REMOTE_USER_GROUPS in the apache
configuration.


ekohl (approved):
Looks like I broke the tests when I merged db41ee82de0ad9feb5569491f6ce6fd1bb6ce522. I just pushed 79806f709ac53e16f6f3a86db9b40e92450cb1f0 and trust that I won't make the same mistake again by merging something that breaks CI.

PR #897, opened 2020-10-27
Refs #29938: Fix failing unit tests due to logging layout change

ehelms:


wbclark (commented):
I believe the double `##` in the commit message is a typo

wbclark (approved):

PR #902, opened 2020-11-09
Cherry-picks for foreman 2.2.1

tbrisker:
Changes include https://projects.theforeman.org/issues/29938 and a followup to https://projects.theforeman.org/issues/30535 fixing a typo

### DIFF
--- a/manifests/params.pp
+++ b/manifests/params.pp
@@ -189,7 +189,7 @@ class foreman::params {
   # Application logging
   $logging_level = 'info'
   $logging_type = 'file'
-  $logging_layout = 'pattern'
+  $logging_layout = 'multiline_request_pattern'
   $loggers = {}
 
   # Rails Cache Store
--- a/README.md
+++ b/README.md
@@ -76,8 +76,8 @@ previous stable release.
 
 This module targets Foreman 2.0+.
 
-If you're running Foreman instance older than Foreman 2.2 you need to change logging layout parameter from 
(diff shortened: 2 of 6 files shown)


ekohl (approved):

wbclark (approved):
https://github.com/theforeman/puppet-foreman/pull/902/commits/cd1378e0603f30983c13fc6ed071316fcab941e2 is backwards incompatible because that logging layout was added in Foreman 2.2

however the purpose of these cherry picks seems to be to support Foreman 2.2.1 so that should be fine

normally a change marked as backwards incompatible should be accompanied by a bump in the major release, but in this case i could continue to list it as backwards incompatible in the changelog while only bumping the minor release

after merging the cherry picks I'll prepare a release PR

manifests/params.pp
```diff
@@ -189,7 +189,7 @@
   # Application logging
   $logging_level = 'info'
   $logging_type = 'file'
-  $logging_layout = 'pattern'
+  $logging_layout = 'multiline_request_pattern'
```
ekohl, line 192: Thinking more about it, it wouldn't be backwards incompatible if we kept this the same. There's a migration in the installer anyway so it isn't needed there.

tbrisker: makes sense

wbclark: done with https://github.com/theforeman/puppet-foreman/pull/904

PR #903, opened 2020-11-11
Release 15.1.2

wbclark:
I bumped only the patch version despite https://github.com/theforeman/puppet-foreman/commit/cf178b3b65b7d2a7c4f7d13e24391b0ba75b35c8 being a breaking change.

I reason that this is OK because that commit does add notes to the README about overriding the new default parameter in order to maintain compatibility with Foreman older than 2.2, so this should be noted as a breaking change in the CHANGELOG and users can review that commit and find the necessary info if they are trying to use an older Foreman version.

CC: @tbrisker @ehelms @ekohl

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,17 @@
 # Changelog
 
+## [15.1.2](https://github.com/theforeman/puppet-foreman/tree/15.1.2) (2020-11-11)
+
+[Full Changelog](https://github.com/theforeman/puppet-foreman/compare/15.1.1...15.1.2)
+
+**Implemented enhancements:**
+
+ - [\#29938](https://projects.theforeman.org/issues/29938) - allow 'multiline_request_pattern' logging layout [\#902](https://github.com/theforeman/puppet-foreman/pull/902) ([domitea](https://github.com/domitea)) + [\#904](https://github.com/theforeman/puppet-foreman/pull/904) ([wbclark](https://github.com/wbclark))
+
+**Fixed bugs:**
+
+- Refs [\#30535](https://projects.theforeman.org/issues/30535) - Correctly unset remote user groups [\#896](https://github.com/theforeman/puppet-foreman/pull/896) ([tbrisker](https://github.com/tbrisker))
+
 ## [15.1.1](https://github.com/theforeman/puppet-foreman/tree/15.1.1) (2020-10-14)
 
 [Full Changelog](https://github.com/theforeman/puppet-foreman/compare/15.1.0...15.1.1)
--- a/metadata.json
+++ b/metadata.json
@@ -1,6 +1,6 @@
 {
   "name": "theforeman-foreman",
-  "version": "15.1.1",
+  "version": "15.1.2",
   "author": "theforeman",
   "summary": "Foreman server configuration",
   "license": "GPL-3.0+",


CHANGELOG.md
```diff
@@ -1,5 +1,17 @@
 # Changelog
 
+## [15.1.2](https://github.com/theforeman/puppet-foreman/tree/15.1.2) (2020-11-11)
+
+[Full Changelog](https://github.com/theforeman/puppet-foreman/compare/15.1.1...15.1.2)
+
+**Breaking changes:**
```
ekohl, line 7: Perhaps we can make it non-breaking: https://github.com/theforeman/puppet-foreman/pull/902/files#r522016518

HISTORY.md
```diff
@@ -1,3 +1,24 @@
+## [15.1.2](https://github.com/theforeman/puppet-foreman/tree/15.1.2) (2020-11-11)
```
ekohl, line 1: I'd stop doing this. In stable branches we typically write them by hand anyway and this only creates noise in the git log. They make reviews harder.

ekohl (changes requested):

HISTORY.md
wbclark: Done.

CHANGELOG.md
wbclark: Rebased.

CHANGELOG.md
```diff
@@ -1,5 +1,18 @@
 # Changelog
 
+## [15.1.2](https://github.com/theforeman/puppet-foreman/tree/15.1.2) (2020-11-11)
+
+[Full Changelog](https://github.com/theforeman/puppet-foreman/compare/15.1.1...15.1.2)
+
+**Implemented enhancements:**
+
+ - fixes [\#29938](https://projects.theforeman.org/issues/29938) - change default logging layout [\#847](https://github.com/theforeman/puppet-foreman/pull/847) ([domitea](https://github.com/domitea))
```
ekohl, line 9: How about:
```suggestion
 - [\#29938](https://projects.theforeman.org/issues/29938) allow `multiline_request_pattern` logging layout [\#902](https://github.com/theforeman/puppet-foreman/pull/902) + [\#904](https://github.com/theforeman/puppet-foreman/pull/904)
```

CHANGELOG.md
```diff
@@ -1,5 +1,18 @@
...
+
+**Fixed bugs:**
+
+- Use backwards compatible 'pattern' logging_layout by default [\#904](https://github.com/theforeman/puppet-foreman/pull/904) ([wbclark](https://github.com/wbclark))
```
ekohl, line 13: Then you can drop this one

ekohl (approved):

ekohl (approved):

tbrisker (approved):

tbrisker:
Pretty sure this failure is unrelated?
```
  Notice: /Stage[main]/Apt::Update/Exec[apt_update]/returns: W: GPG error: https://deb.theforeman.org plugins Release: The following signatures were invalid: BADSIG 6F8600B9563278F6 Foreman Automatic Signing Key (2016) <packages@theforeman.org>

  Notice: /Stage[main]/Apt::Update/Exec[apt_update]/returns: E: The repository 'https://deb.theforeman.org plugins Release' is not signed.

tbrisker:
I pulled in the github action change to the stable branch as well, so hopefully this won't fail if we need to try again (or at least we won't need to wait a whole day for the travis queue to clear)

tbrisker:
All :green_apple: now, for the record acceptance tests took about 30 minutes and unit tests about 12. Looking forward to when we'll be able to drop puppet 5 and cut the matrix in half.

PR #909, opened 2020-12-07
Set the reverse proxy host to the name of the service

ehelms:
This change stems from the breakage and investigation here:
https://community.theforeman.org/t/katello-nightly-rpm-pipeline-769-failed/21468/4?u=ehelms

The shadowing of the /v2/ in Pulpcore API by Katello to provide auth causes
it to clash with the reverse proxying within the same Vhost to Foreman.
This appears to be tied to how the unix socket based reverse proxy handles things.

### DIFF
--- a/manifests/config/apache.pp
+++ b/manifests/config/apache.pp
@@ -135,7 +135,7 @@ class foreman::config::apache(
   $docroot = "${app_root}/public"
 
   if $proxy_backend =~ 'unix://' {
-    $_proxy_backend = "${proxy_backend}|http://${servername}/"
+    $_proxy_backend = "${proxy_backend}|http://foreman/"
   } else {
     $_proxy_backend = regsubst($proxy_backend, 'tcp://', 'http://')
   }
--- a/spec/classes/foreman_config_apache_spec.rb
+++ b/spec/classes/foreman_config_apache_spec.rb
@@ -228,14 +228,14 @@ describe 'foreman::config::apache' do
                 .with_proxy_pass(
                   "no_proxy_uris" => ['/pulp', '/pulp2', '/streamer', '/pub', '/icons'],
                   "path"          => '/',
-                  "url"           => 'unix:///run/foreman.sock|http://foo.example.com/',
+                  "url"           => 'unix:///run/foreman.sock|http://foreman/',
                   "params"        => { "retry" => '0' },
                 )
                 .with_rewrites([
                   {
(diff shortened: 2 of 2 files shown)


ehelms:
Related change -- https://github.com/theforeman/puppet-pulpcore/pull/153

evgeni:
So I *think* I understand the current breakage and what the fix does and it makes sense to me. There is one concern though (not strictly introduced by this PR, but maybe further exposed):

For the sake of the example: Apache has `foreman.example.com` as the vhost (and that's the name clients talk to).

Before this change, the `Host` header that was passed to Rails was also `foreman.example.com`, so any `Location` (or similar) redirect produced by Rails was already "correct". With this change, Rails will see `foreman` in the `Host` header and use that for redirects. Thankfully, Apache has [`ProxyPassReverse`](https://httpd.apache.org/docs/trunk/mod/mod_proxy.html#proxypassreverse) which will magically translate this back to `foreman.example.com` for us when it sees it in a `Location` header. It seems we do set `ProxyPassReverse` to the same value (but I couldn't find a test that ensures that), so that should be fine (but we really should have that in tests).

There are however other places that *can* contain the `Host` header as seen by the backend:
* `Set-Cookie`, which can be adjusted using [`ProxyPassReverseCookieDomain`](https://httpd.apache.org/docs/trunk/mod/mod_proxy.html#proxypassreversecookiedomain) (today, our cookies don't contain domains, so it's not important that we don't set this parameter)
* [`RewriteRule`s](https://httpd.apache.org/docs/2.4/rewrite/flags.html#flag_p) that can trigger proxying, and I don't see if they also enable `ProxyPassReverse`
https://github.com/theforeman/puppet-foreman/blob/42165264305d26a8767ae1253864c4f981339fb9/spec/classes/foreman_config_apache_spec.rb#L238

evgeni:
Hah, https://httpd.apache.org/docs/2.4/rewrite/proxy.html says that one has to set `ProxyPassReverse` explicitly for `RewriteRule` proxies too.

evgeni:
FWIW, I don't think that `Host` is later used *on* the WS connection at all, so maybe we're good anyways.

evgeni (approved):
after talking with Eric, the `ProxyPassReverse` is always added by the apache module:
https://github.com/puppetlabs/puppetlabs-apache/blob/624330496a431dad03e430b8f2c9ac1914c3b8de/templates/vhost/_proxy.erb#L49-L55

ACK

PR #912, opened 2021-01-06
Fixes #30284 - Improve smartproxy registration failure error messages

wbclark:
draft PR

### DIFF
--- a/lib/puppet/provider/foreman_resource/rest_v3.rb
+++ b/lib/puppet/provider/foreman_resource/rest_v3.rb
@@ -55,11 +55,14 @@ Puppet::Type.type(:foreman_resource).provide(:rest_v3) do
     OAuth::AccessToken.new(oauth_consumer)
   end
 
-  def request(method, path, params = {}, data = nil, headers = {})
+  def request_uri(path)
     base_url = resource[:base_url]
     base_url += '/' unless base_url.end_with?('/')
+    URI.join(base_url, path)
+  end
 
-    uri = URI.join(base_url, path)
+  def request(method, path, params = {}, data = nil, headers = {})
+    uri = request_uri(path)
     uri.query = params.map { |p,v| "#{CGI.escape(p.to_s)}=#{CGI.escape(v.to_s)}" }.join('&') unless params.empty?
 
     headers = {
--- a/lib/puppet/provider/foreman_resource/rest_v3.rb
+++ b/lib/puppet/provider/foreman_resource/rest_v3.rb
@@ -99,6 +99,20 @@ Puppet::Type.type(:foreman_resource).provide(:rest_v3) do
   end
 
   def error_message(response)
-    JSON.parse(response.body)['error']['full_messages'].join(' ') rescue "unknown error (response #{response.code})"
+    explanations = {
+      '400' => 'Something is wrong with the data sent to Foreman server',
+      '401' => 'Often this is caused by invalid Oauth credentials',
+      '404' => 'The requested resource was not found',
+      '500' => 'Check /var/log/foreman/production.log on Foreman server for detailed information',
+      '502' => 'The webserver received an invalid response from the backend service. Was Foreman unable to handle the request?',
+      '503' => 'The webserver was unable to reach the backend service. Is foreman.service running?',
+      '504' => 'The webserver timed out waiting for a response from the backend service. Is Foreman under unusually heavy load?'
+    }
+
+    if (explanation = explanations[response.code.to_str])
+      "Response: #{response.code} #{response.message}: #{explanation}"
+    else
+      JSON.parse(response.body)['error']['full_messages'].join(' ') rescue "Response: #{response.code} #{response.message}"
+    end
   end
 end
--- a/lib/puppet/provider/foreman_smartproxy/rest_v3.rb
+++ b/lib/puppet/provider/foreman_smartproxy/rest_v3.rb
@@ -3,8 +3,14 @@ Puppet::Type.type(:foreman_smartproxy).provide(:rest_v3, :parent => Puppet::Type
 
   def proxy
     @proxy ||= begin
-      r = request(:get, 'api/v2/smart_proxies', :search => %{name="#{resource[:name]}"})
-      raise Puppet::Error.new("Proxy #{resource[:name]} cannot be retrieved: #{error_message(r)}") unless success?(r)
+      path = 'api/v2/smart_proxies'
+      r = request(:get, path, :search => %{name="#{resource[:name]}"})
+
+      unless success?(r)
+        error_string = "Error making GET request to Foreman at #{request_uri(path)}: #{error_message(r)}"
+        raise Puppet::Error.new(error_string)
+      end
+
       JSON.load(r.body)['results'][0]
     end
   end
@@ -19,14 +25,26 @@ Puppet::Type.type(:foreman_smartproxy).provide(:rest_v3, :parent => Puppet::Type
 
   def create
     post_data = {:smart_proxy => {:name => resource[:name], :url => resource[:url]}}.to_json
-    r = request(:post, 'api/v2/smart_proxies', {}, post_data)
-    raise Puppet::Error.new("Proxy #{resource[:name]} cannot be registered: #{error_message(r)}") unless success?(r)
+    path = 'api/v2/smart_proxies'
+    r = request(:post, path, {}, post_data)
+
+    unless success?(r)
+      error_string = "Error making POST request to Foreman at #{request_uri(path)}: #{error_message(r)}"
+      raise Puppet::Error.new(error_string)
+    end
+
     validate_features!(resource[:features], features_list(JSON.load(r.body)))
   end
 
   def destroy
-    r = request(:delete, "api/v2/smart_proxies/#{id}")
-    raise Puppet::Error.new("Proxy #{resource[:name]} cannot be removed: #{error_message(r)}") unless success?(r)
+    path = "api/v2/smart_proxies/#{id}"
+    r = request(:delete, path)
+
+    unless success?(r)
+      error_string = "Error making DELETE request to Foreman at #{request_uri(path)}: #{error_message(r)}"
+      raise Puppet::Error.new(error_string)
+    end
+
     @proxy = nil
   end
 
@@ -36,8 +54,13 @@ Puppet::Type.type(:foreman_smartproxy).provide(:rest_v3, :parent => Puppet::Type
 
   def url=(value)
     post_data = {:smart_proxy => {:url => value}}.to_json
-    r = request(:put, "api/v2/smart_proxies/#{id}", {}, post_data)
-    raise Puppet::Error.new("Proxy #{resource[:name]} cannot be updated: #{error_message(r)}") unless success?(r)
+    path = "api/v2/smart_proxies/#{id}"
+    r = request(:put, path, {}, post_data)
+
+    unless success?(r)
+      error_string = "Error making PUT request to Foreman at #{request_uri(path)}: #{error_message(r)}"
+      raise Puppet::Error.new(error_string)
+    end
   end
 
   def features
@@ -49,8 +72,13 @@ Puppet::Type.type(:foreman_smartproxy).provide(:rest_v3, :parent => Puppet::Type
   end
 
   def refresh_features!
-    r = request(:put, "api/v2/smart_proxies/#{id}/refresh")
-    raise Puppet::Error.new("Proxy #{resource[:name]} cannot be refreshed: #{error_message(r)}") unless success?(r)
+    path = "api/v2/smart_proxies/#{id}/refresh"
+    r = request(:put, path)
+
+    unless success?(r)
+      error_string = "Error making PUT request to #{request_uri(path)}: #{error_message(r)}"
+      raise Puppet::Error.new(error_string)
+    end
 
     body = JSON.load(r.body)
     # Replace proxy/feature list cache: pre-#19476 versions have limited responses, clear cache and re-fetch for them
--- a/spec/unit/foreman_resource_rest_v3_spec.rb
+++ b/spec/unit/foreman_resource_rest_v3_spec.rb
@@ -142,11 +142,39 @@ describe provider_class do
 
   describe '#error_message(response)' do
     it 'returns array of errors from JSON' do
-      expect(provider.error_message(double(:body => '{"error":{"full_messages":["error1","error2"]}}'))).to eq('error1 error2')
+      expect(provider.error_message(double(:body => '{"error":{"full_messages":["error1","error2"]}}', :code => 'dummycode'))).to eq('error1 error2')
     end
 
-    it 'returns message for missing error messages' do
-      expect(provider.error_message(double(:body => '{}', :code => 404))).to eq('unknown error (response 404)')
+    it 'returns message for unrecognized response code' do
+      expect(provider.error_message(double(:body => '{}', :code => '418', :message => "I'm a teapot"))).to eq("Response: 418 I'm a teapot")
+    end
+
+    it 'returns message for 400 response' do
+      expect(provider.error_message(double(:body => '{}', :code => '400', :message => 'Bad Request'))).to eq('Response: 400 Bad Request: Something is wrong with the data sent to Foreman server')
+    end
+
+    it 'returns message for 401 response' do
+      expect(provider.error_message(double(:body => '{}', :code => '401', :message => 'Unauthorized Request'))).to eq('Response: 401 Unauthorized Request: Often this is caused by invalid Oauth credentials')
+    end
+
+    it 'returns message for 404 response' do
+      expect(provider.error_message(double(:body => '{}', :code => '404', :message => 'Not Found'))).to eq('Response: 404 Not Found: The requested resource was not found')
+    end
+
+    it 'returns message for 500 response' do
+      expect(provider.error_message(double(:body => '{}', :code => '500', :message => 'Internal Server Error'))).to eq('Response: 500 Internal Server Error: Check /var/log/foreman/production.log on Foreman server for detailed information')
+    end
+
+    it 'returns message for 502 response' do
+      expect(provider.error_message(double(:body => '{}', :code => '502', :message => 'Bad Gateway'))).to eq('Response: 502 Bad Gateway: The webserver received an invalid response from the backend service. Was Foreman unable to handle the request?')
+    end
+
+    it 'returns message for 503 response' do
+      expect(provider.error_message(double(:body => '{}', :code => '503', :message => 'Service Unavailable'))).to eq('Response: 503 Service Unavailable: The webserver was unable to reach the backend service. Is foreman.service running?')
+    end
+
+    it 'returns message for 504 response' do
+      expect(provider.error_message(double(:body => '{}', :code => '504', :message => 'Gateway Timeout'))).to eq('Response: 504 Gateway Timeout: The webserver timed out waiting for a response from the backend service. Is Foreman under unusually heavy load?')
     end
   end
 end


wbclark:
Current output when httpd is down:

```
    2021-01-06 17:17:26 [NOTICE] [configure] 1600 out of 1684 done.
    2021-01-06 17:17:30 [ERROR ] [configure] /Stage[main]/Foreman_proxy::Register/Foreman_smartproxy[centos7-foreman-proxy-nightly.mercury.example.com]: Could not evaluate: Exception Failed to open TCP connection to centos7-katello-3-18.mercury.example.com:443 (Connection refused - connect(2) for "centos7-katello-3-18.mercury.example.com" port 443) in get request to: https://centos7-katello-3-18.mercury.example.com/api/v2/smart_proxies?search=name=%22centos7-foreman-proxy-nightly.mercury.example.com%22
    2021-01-06 17:17:30 [ERROR ] [configure] Wrapped exception:
    2021-01-06 17:17:30 [ERROR ] [configure] Failed to open TCP connection to centos7-katello-3-18.mercury.example.com:443 (Connection refused - connect(2) for "centos7-katello-3-18.mercury.example.com" port 443)
    2021-01-06 17:17:30 [ERROR ] [configure] /Stage[main]/Foreman_proxy::Register/Foreman_smartproxy[centos7-foreman-proxy-nightly.mercury.example.com]: Failed to call refresh: Exception Failed to open TCP connection to centos7-katello-3-18.mercury.example.com:443 (Connection refused - connect(2) for "centos7-katello-3-18.mercury.example.com" port 443) in get request to: https://centos7-katello-3-18.mercury.example.com/api/v2/smart_proxies?search=name=%22centos7-foreman-proxy-nightly.mercury.example.com%22
    2021-01-06 17:17:30 [ERROR ] [configure] /Stage[main]/Foreman_proxy::Register/Foreman_smartproxy[centos7-foreman-proxy-nightly.mercury.example.com]: Exception Failed to open TCP connection to centos7-katello-3-18.mercury.example.com:443 (Connection refused - connect(2) for "centos7-katello-3-18.mercury.example.com" port 443) in get request to: https://centos7-katello-3-18.mercury.example.com/api/v2/smart_proxies?search=name=%22centos7-foreman-proxy-nightly.mercury.example.com%22
    2021-01-06 17:17:30 [ERROR ] [configure] Wrapped exception:
    2021-01-06 17:17:30 [ERROR ] [configure] Failed to open TCP connection to centos7-katello-3-18.mercury.example.com:443 (Connection refused - connect(2) for "centos7-katello-3-18.mercury.example.com" port 443)
    2021-01-06 17:17:32 [NOTICE] [configure] System configuration has finished.
    2021-01-06 17:17:32 [NOTICE] [post] Executing hooks in group post
    2021-01-06 17:17:32 [NOTICE] [post] All hooks in group post finished
  
      There were errors detected during install.
      Please address the errors and re-run the installer to ensure the system is properly configured.
      Failing to do so is likely to result in broken functionality.
  
      The full log is at /var/log/foreman-installer/foreman-proxy-content.log
```

wbclark:
When foreman.service is down, there is no issue because foreman.socket will bring it back up.

OTOH, when foreman.socket itself is down:

```
    2021-01-06 19:14:49 [NOTICE] [configure] 1600 out of 1684 done.
    2021-01-06 19:14:53 [ERROR ] [configure] /Stage[main]/Foreman_proxy::Register/Foreman_smartproxy[centos7-foreman-proxy-nightly.mercury.example.com]: Could not evaluate: Proxy centos7-foreman-proxy-nightly.mercury.example.com cannot be retrieved: unknown error (response 503)
    2021-01-06 19:14:53 [ERROR ] [configure] /Stage[main]/Foreman_proxy::Register/Foreman_smartproxy[centos7-foreman-proxy-nightly.mercury.example.com]: Failed to call refresh: Proxy centos7-foreman-proxy-nightly.mercury.example.com cannot be retrieved: unknown error (response 503)
    2021-01-06 19:14:53 [ERROR ] [configure] /Stage[main]/Foreman_proxy::Register/Foreman_smartproxy[centos7-foreman-proxy-nightly.mercury.example.com]: Proxy centos7-foreman-proxy-nightly.mercury.example.com cannot be retrieved: unknown error (response 503)
    2021-01-06 19:14:55 [NOTICE] [configure] System configuration has finished.
    2021-01-06 19:14:55 [NOTICE] [post] Executing hooks in group post
    2021-01-06 19:14:55 [NOTICE] [post] All hooks in group post finished
  
      There were errors detected during install.
      Please address the errors and re-run the installer to ensure the system is properly configured.
      Failing to do so is likely to result in broken functionality.
  
      The full log is at /var/log/foreman-installer/foreman-proxy-content.log
```

wbclark:
When a 401 response is received (for example by using incorrect `--foreman-proxy-oauth-consumer-key`):

```
    2021-01-06 20:48:53 [NOTICE] [configure] 1600 out of 1684 done.
    2021-01-06 20:48:57 [ERROR ] [configure] /Stage[main]/Foreman_proxy::Register/Foreman_smartproxy[centos7-foreman-proxy-nightly.mercury.example.com]: Could not evaluate: Proxy centos7-foreman-proxy-nightly.mercury.example.com cannot be retrieved: unknown error (response 401)
    2021-01-06 20:48:57 [ERROR ] [configure] /Stage[main]/Foreman_proxy::Register/Foreman_smartproxy[centos7-foreman-proxy-nightly.mercury.example.com]: Failed to call refresh: Proxy centos7-foreman-proxy-nightly.mercury.example.com cannot be retrieved: unknown error (response 401)
    2021-01-06 20:48:57 [ERROR ] [configure] /Stage[main]/Foreman_proxy::Register/Foreman_smartproxy[centos7-foreman-proxy-nightly.mercury.example.com]: Proxy centos7-foreman-proxy-nightly.mercury.example.com cannot be retrieved: unknown error (response 401)
    2021-01-06 20:48:59 [NOTICE] [configure] System configuration has finished.
    2021-01-06 20:48:59 [NOTICE] [post] Executing hooks in group post
    2021-01-06 20:48:59 [NOTICE] [post] All hooks in group post finished
  
      There were errors detected during install.
      Please address the errors and re-run the installer to ensure the system is properly configured.
      Failing to do so is likely to result in broken functionality.
  
      The full log is at /var/log/foreman-installer/foreman-proxy-content.log
```

lib/puppet/provider/foreman_smartproxy/rest_v3.rb
```diff
@@ -4,7 +4,15 @@
...
     @proxy ||= begin
       r = request(:get, 'api/v2/smart_proxies', :search => %{name="#{resource[:name]}"})
-      raise Puppet::Error.new("Proxy #{resource[:name]} cannot be retrieved: #{error_message(r)}") unless success?(r)
+      #raise Puppet::Error.new("Proxy #{resource[:name]} cannot be retrieved: #{error_message(r)}") unless success?(r)
+      unless success?(r)
+        case error_message(r)
+        when /response 401/
+          raise Puppet::Error.new("401 RESPONSE PLACEHOLDER ERROR MESSAGE")
+        when /response 503/
+          raise Puppet::Error.new("503 RESPONSE PLACEHOLDER ERROR MESSAGE")
```
ekohl, line 13: Please see the API for https://www.rubydoc.info/stdlib/net/Net/HTTPResponse. There's no need to parse the error message if you have the code.

You can also take a look at [error_message(response)](https://github.com/theforeman/puppet-foreman/blob/master/lib/puppet/provider/foreman_resource/rest_v3.rb#L97-L99) which may be a much better place to handle it.

lib/puppet/provider/foreman_smartproxy/rest_v3.rb
```diff
@@ -4,7 +4,7 @@
   def proxy
     @proxy ||= begin
       r = request(:get, 'api/v2/smart_proxies', :search => %{name="#{resource[:name]}"})
-      raise Puppet::Error.new("Proxy #{resource[:name]} cannot be retrieved: #{error_message(r)}") unless success?(r)
```
ekohl, line 7: I think this is a degradation in usefulness. Now you no longer know if it was GET or an update action. See `create` and `destroy`.

lib/puppet/provider/foreman_resource/rest_v3.rb
```diff
@@ -95,6 +95,13 @@ def success?(response)
...
 
   def error_message(response)
-    JSON.parse(response.body)['error']['full_messages'].join(' ') rescue "unknown error (response #{response.code})"
+    case response.code
+    when '401'
+      '401 RESPONSE PLACEHOLDER ERROR MESSAGE'
+    when '503'
+      '503 RESPONSE PLACEHOLDER ERROR MESSAGE'
+    else
+      JSON.parse(response.body)['error']['full_messages'].join(' ') rescue "unknown error (response #{response.code})"
```
ekohl, line 104: It probably makes sense to always attempt to parse JSON and if see if it contains a full error message. This allows the Foreman application to give the most precise error in some scenarios. Alternatively, you can peek at the content type header to see if it actually is JSON.

lib/puppet/provider/foreman_smartproxy/rest_v3.rb
wbclark: I want to make some kind of change here because the current wording `Proxy #{resource[:name]} cannot be retrieved:` is often misunderstood by users to mean there is an issue with their Foreman proxy infrastructure. I'll try to do this in a way that remains sensitive to the HTTP verb and conveys the same information in a way that is less open to misinterpretation.

wbclark: Perhaps `Proxy #{resource[:name]} cannot be retrieved from Foreman server` is sufficient.

ekohl: The important part is that it's retrieving data (as opposed to creating, updating or removing).

Perhaps it should be qualified as `Smart Proxy` instead of `Proxy`. It would introduce branding problems, but I think that's acceptable. `Proxy` is ambiguous since we also have HTTP proxies and AFAIK we don't brand the API endpoints either.

ehelms: Does the response in this case contain the full problem? e.g. to know if it was an auth error, a cannot find that proxy error, other

ekohl: Looks like the full error is only returned here:
https://github.com/theforeman/foreman/blob/b6c40daf3a4b589630e06d409697a62b664c27f0/app/views/api/v2/errors/unprocessable_entity.json.rabl#L6
So only on HTTP 422.

However, that discussion really belongs in https://github.com/theforeman/puppet-foreman/pull/912/files#r553307082. I insist that it still logs the action it was trying to perform.

lib/puppet/provider/foreman_resource/rest_v3.rb
```diff
@@ -95,6 +95,15 @@ def success?(response)
   end
 
   def error_message(response)
-    JSON.parse(response.body)['error']['full_messages'].join(' ') rescue "unknown error (response #{response.code})"
+    messages = {
+      401: '401 Unauthorized response from Foreman server'
+      503: '503 Server Unavailable when trying to reach Foreman server: Apache was unable to reach the backend service. Is foreman.service running?'
```
ekohl, line 100: The server doesn't have to be Apache. We have community members who use nginx.

lib/puppet/provider/foreman_smartproxy/rest_v3.rb
wbclark: I've updated such that it logs more clearly what is the exact action it tried to perform. The messages now include the http verb,  uri, a simple explanation of the issue, and the ability to log an additional custom message per http error code.

wbclark:
I've marked this as ready for review.

Here is an example of the output when a 401 response is received from Foreman:

```
    2021-01-15 16:49:43 [NOTICE] [configure] 1600 out of 1684 done.
    2021-01-15 16:49:48 [ERROR ] [configure] /Stage[main]/Foreman_proxy::Register/Foreman_smartproxy[centos7-foreman-proxy-nightly.mercury.example.com]: Could not evaluate: Error making GET request to https://centos7-katello-3-18.mercury.example.com/api/v2/smart_proxies
    2021-01-15 16:49:48 [ERROR ] [configure] SmartProxy centos7-foreman-proxy-nightly.mercury.example.com cannot be retrieved from Foreman:
    2021-01-15 16:49:48 [ERROR ] [configure] 401 Unauthorized Request: Check credentials for validity
```

lib/puppet/provider/foreman_resource/rest_v3.rb
```diff
@@ -95,6 +98,21 @@ def success?(response)
...
+      '403' => '403 Forbidden: The credentials were valid but do not grant access to the requested resource',
+      '404' => '404 Not Found: The requested resource was not found',
+      '500' => '500 Internal Server Error: Check /var/log/foreman/production.log on Foreman server for detailed information',
+      '502' => '502 Bad Gateway: The webserver received an invalid response from the application server. Was Foreman unable to handle the request?',
+      '503' => '503 Service Unavailable: The webserver was unable to reach the backend service. Is foreman.service running?',
+      '504' => '504 Gateway Timeout: The webserver timed out waiting for a response from the application. Is Foreman under unusually heavy load?'
+    }
+
+    if messages.keys.include?(response.code.to_str)
+      JSON.parse(response.body)['error']['full_messages'].join(' ') rescue messages[response.code.to_str]
```
ekohl, line 113: As noted (https://github.com/theforeman/puppet-foreman/pull/912#discussion_r553888668), the errors only show up for HTTP 422 so I think it's actually good enough to simplify this:
```suggestion
      messages[response.code.to_str]
```

lib/puppet/provider/foreman_resource/rest_v3.rb
```diff
@@ -95,6 +98,21 @@ def success?(response)
...
+      '401' => '401 Unauthorized Request: Check credentials for validity',
+      '403' => '403 Forbidden: The credentials were valid but do not grant access to the requested resource',
+      '404' => '404 Not Found: The requested resource was not found',
+      '500' => '500 Internal Server Error: Check /var/log/foreman/production.log on Foreman server for detailed information',
+      '502' => '502 Bad Gateway: The webserver received an invalid response from the application server. Was Foreman unable to handle the request?',
+      '503' => '503 Service Unavailable: The webserver was unable to reach the backend service. Is foreman.service running?',
+      '504' => '504 Gateway Timeout: The webserver timed out waiting for a response from the application. Is Foreman under unusually heavy load?'
+    }
+
+    if messages.keys.include?(response.code.to_str)
```
ekohl, line 112:
```suggestion
    if messages.include?(response.code.to_str)
```

lib/puppet/provider/foreman_smartproxy/rest_v3.rb
```diff
@@ -3,8 +3,18 @@
...
     @proxy ||= begin
-      r = request(:get, 'api/v2/smart_proxies', :search => %{name="#{resource[:name]}"})
-      raise Puppet::Error.new("Proxy #{resource[:name]} cannot be retrieved: #{error_message(r)}") unless success?(r)
+      path = 'api/v2/smart_proxies'
+      r = request(:get, path, :search => %{name="#{resource[:name]}"})
+
+      unless success?(r)
+        error_string = <<~ERROR.strip
+          Error making GET request to #{request_uri(path)}
+          SmartProxy #{resource[:name]} cannot be retrieved from Foreman:
```
ekohl, line 12: I'd try to avoid multiple lines as that's hard to read in the output.

lib/puppet/provider/foreman_smartproxy/rest_v3.rb
```diff
@@ -19,14 +29,34 @@ def exists?
...
   def create
     post_data = {:smart_proxy => {:name => resource[:name], :url => resource[:url]}}.to_json
-    r = request(:post, 'api/v2/smart_proxies', {}, post_data)
-    raise Puppet::Error.new("Proxy #{resource[:name]} cannot be registered: #{error_message(r)}") unless success?(r)
+    path = 'api/v2/smart_proxies'
+    r = request(:post, path, {}, post_data)
+
+    unless success?(r)
+      error_string = <<~ERROR.strip
+        Error making POST request to #{request_uri(path)}
```
ekohl, line 37: Is this really adding anything for the user? IMHO it certainly shouldn't be the first thing. First thing should be the action that failed (the line below this). Then technical details.

ekohl (changes requested):
Does the response know the verb as well? I think `error_message` could give the technical details as well if it has the request object (`r`) and you don't need to repeat it 4 times.

lib/puppet/provider/foreman_smartproxy/rest_v3.rb
wbclark: Just as you were posting this, I was thinking we need one more line to improve the readability.

The current design is

```
    2021-01-15 16:49:43 [NOTICE] [configure] 1600 out of 1684 done.
    2021-01-15 16:49:48 [ERROR ] [configure] /Stage[main]/Foreman_proxy::Register/Foreman_smartproxy[centos7-foreman-proxy-nightly.mercury.example.com]: Could not evaluate: Error making GET request to https://centos7-katello-3-18.mercury.example.com/api/v2/smart_proxies
    2021-01-15 16:49:48 [ERROR ] [configure] SmartProxy centos7-foreman-proxy-nightly.mercury.example.com cannot be retrieved from Foreman:
    2021-01-15 16:49:48 [ERROR ] [configure] 401 Unauthorized Request: Check credentials for validity
```

I think it should be (taking your other feedback about ordering into consideration) more like:

```

    2021-01-15 16:49:43 [NOTICE] [configure] 1600 out of 1684 done.
    2021-01-15 16:49:48 [ERROR ] [configure] /Stage[main]/Foreman_proxy::Register/Foreman_smartproxy[centos7-foreman-proxy-nightly.mercury.example.com]: Could not evaluate:
    2021-01-15 16:49:48 [ERROR ] [configure] SmartProxy centos7-foreman-proxy-nightly.mercury.example.com cannot be retrieved from Foreman:
    2021-01-15 16:49:48 [ERROR ] [configure] Error making GET request to https://centos7-katello-3-18.mercury.example.com/api/v2/smart_proxies
    2021-01-15 16:49:48 [ERROR ] [configure] 401 Unauthorized Request: Check credentials for validity
```

I think that provides the best clarity, precision, and readability. Let me know if it's agreeable to you

lib/puppet/provider/foreman_smartproxy/rest_v3.rb
wbclark: Agreed. I'll change the ordering in my next round of changes.

lib/puppet/provider/foreman_resource/rest_v3.rb
```diff
@@ -95,6 +98,21 @@ def success?(response)
...
-    JSON.parse(response.body)['error']['full_messages'].join(' ') rescue "unknown error (response #{response.code})"
+    messages = {
+      '400' => '400 Bad Request: Something is wrong with the data we sent to Foreman server',
+      '401' => '401 Unauthorized Request: Check credentials for validity',
+      '403' => '403 Forbidden: The credentials were valid but do not grant access to the requested resource',
+      '404' => '404 Not Found: The requested resource was not found',
+      '500' => '500 Internal Server Error: Check /var/log/foreman/production.log on Foreman server for detailed information',
+      '502' => '502 Bad Gateway: The webserver received an invalid response from the application server. Was Foreman unable to handle the request?',
+      '503' => '503 Service Unavailable: The webserver was unable to reach the backend service. Is foreman.service running?',
+      '504' => '504 Gateway Timeout: The webserver timed out waiting for a response from the application. Is Foreman under unusually heavy load?'
```
ehelms, line 109: I wonder if we could say `Apache` instead of `webserver` here. If I am user seeing these, they sound generic enough that I would have to think about what the webserver vs the application was.

lib/puppet/provider/foreman_resource/rest_v3.rb
```diff
@@ -95,6 +98,21 @@ def success?(response)
...
   def error_message(response)
-    JSON.parse(response.body)['error']['full_messages'].join(' ') rescue "unknown error (response #{response.code})"
+    messages = {
+      '400' => '400 Bad Request: Something is wrong with the data we sent to Foreman server',
+      '401' => '401 Unauthorized Request: Check credentials for validity',
+      '403' => '403 Forbidden: The credentials were valid but do not grant access to the requested resource',
+      '404' => '404 Not Found: The requested resource was not found',
+      '500' => '500 Internal Server Error: Check /var/log/foreman/production.log on Foreman server for detailed information',
+      '502' => '502 Bad Gateway: The webserver received an invalid response from the application server. Was Foreman unable to handle the request?',
+      '503' => '503 Service Unavailable: The webserver was unable to reach the backend service. Is foreman.service running?',
```
ehelms, line 108: This line uses `backend service`, others use `application`. As a user seeing these, I'd like consistency of terms and as specific as possible to help me identify what might be having problems.

ehelms:
If you have handy examples of the output for 3-4 of these conditions as printed by the installer, that would help a lot for me to think about them as a user encountering them.

lib/puppet/provider/foreman_resource/rest_v3.rb
wbclark: I think branding would be extremely helpful here. I can see at least 3 uses that substantively improve this feature:

1. Being properly specific with regards to 'SmartProxy' vs. 'Capsule' and similar brand specific terms
2. Linking to relevant KCS in downstream branded error messages
3. Being more specific in message language ('Apache' vs. 'Webserver') due to less variety in deployments

ehelms: For the 504, as an example, could it read "504 Gateway Timeout: The webserver timed out waiting for a response from Foreman" or `foreman.service` as you used in the 503.

I think 'Smart Proxy' is the right term regardless of branding as that is the software. Though, arguably, its 'foreman-proxy' and thus rightfully confusing :)

wbclark:
Sure @ehelms 

Here's an example 401 Response:

```
    2021-01-15 17:44:00 [NOTICE] [configure] Starting system configuration.
      The total number of configuration tasks may increase during the run.
      Observe logs or specify --verbose-log-level to see individual configuration tasks.
    2021-01-15 17:44:24 [NOTICE] [configure] 100 out of 1657 done.
    2021-01-15 17:44:37 [NOTICE] [configure] 200 out of 1657 done.
    2021-01-15 17:45:27 [NOTICE] [configure] 300 out of 1657 done.
    2021-01-15 17:46:07 [NOTICE] [configure] 400 out of 1657 done.
    2021-01-15 17:46:10 [NOTICE] [configure] 500 out of 1658 done.
    2021-01-15 17:46:11 [NOTICE] [configure] 600 out of 1658 done.
    2021-01-15 17:46:20 [NOTICE] [configure] 700 out of 1658 done.
    2021-01-15 17:46:22 [NOTICE] [configure] 800 out of 1676 done.
    2021-01-15 17:46:31 [NOTICE] [configure] 900 out of 1679 done.
    2021-01-15 17:46:32 [NOTICE] [configure] 1000 out of 1682 done.
    2021-01-15 17:46:49 [NOTICE] [configure] 1100 out of 1682 done.
    2021-01-15 17:46:51 [NOTICE] [configure] 1200 out of 1684 done.
    2021-01-15 17:46:52 [NOTICE] [configure] 1300 out of 1684 done.
    2021-01-15 17:46:52 [NOTICE] [configure] 1400 out of 1684 done.
    2021-01-15 17:47:15 [NOTICE] [configure] 1500 out of 1684 done.
    2021-01-15 17:47:16 [NOTICE] [configure] 1600 out of 1684 done.
    2021-01-15 17:47:20 [ERROR ] [configure] /Stage[main]/Foreman_proxy::Register/Foreman_smartproxy[centos7-foreman-proxy-nightly.mercury.example.com]: Could not evaluate:
    2021-01-15 17:47:20 [ERROR ] [configure] SmartProxy centos7-foreman-proxy-nightly.mercury.example.com cannot be retrieved from Foreman:
    2021-01-15 17:47:20 [ERROR ] [configure] Error making GET request to https://centos7-katello-3-18.mercury.example.com/api/v2/smart_proxies
    2021-01-15 17:47:20 [ERROR ] [configure] 401 Unauthorized Request: Check credentials for validity
    2021-01-15 17:47:20 [ERROR ] [configure] /Stage[main]/Foreman_proxy::Register/Foreman_smartproxy[centos7-foreman-proxy-nightly.mercury.example.com]: Failed to call refresh:
    2021-01-15 17:47:20 [ERROR ] [configure] SmartProxy centos7-foreman-proxy-nightly.mercury.example.com cannot be retrieved from Foreman:
    2021-01-15 17:47:20 [ERROR ] [configure] Error making GET request to https://centos7-katello-3-18.mercury.example.com/api/v2/smart_proxies
    2021-01-15 17:47:20 [ERROR ] [configure] 401 Unauthorized Request: Check credentials for validity
    2021-01-15 17:47:20 [ERROR ] [configure] /Stage[main]/Foreman_proxy::Register/Foreman_smartproxy[centos7-foreman-proxy-nightly.mercury.example.com]:
    2021-01-15 17:47:20 [ERROR ] [configure] SmartProxy centos7-foreman-proxy-nightly.mercury.example.com cannot be retrieved from Foreman:
    2021-01-15 17:47:20 [ERROR ] [configure] Error making GET request to https://centos7-katello-3-18.mercury.example.com/api/v2/smart_proxies
    2021-01-15 17:47:20 [ERROR ] [configure] 401 Unauthorized Request: Check credentials for validity
    2021-01-15 17:47:22 [NOTICE] [configure] System configuration has finished.
    2021-01-15 17:47:22 [NOTICE] [post] Executing hooks in group post
    2021-01-15 17:47:22 [NOTICE] [post] All hooks in group post finished
```

Here's an example with 503 response:

```
    2021-01-15 21:13:51 [NOTICE] [configure] Starting system configuration.
      The total number of configuration tasks may increase during the run.
      Observe logs or specify --verbose-log-level to see individual configuration tasks.
    2021-01-15 21:14:15 [NOTICE] [configure] 100 out of 1657 done.
    2021-01-15 21:14:28 [NOTICE] [configure] 200 out of 1657 done.
    2021-01-15 21:15:17 [NOTICE] [configure] 300 out of 1657 done.
    2021-01-15 21:15:57 [NOTICE] [configure] 400 out of 1657 done.
    2021-01-15 21:16:00 [NOTICE] [configure] 500 out of 1658 done.
    2021-01-15 21:16:01 [NOTICE] [configure] 600 out of 1658 done.
    2021-01-15 21:16:10 [NOTICE] [configure] 700 out of 1658 done.
    2021-01-15 21:16:12 [NOTICE] [configure] 800 out of 1676 done.
    2021-01-15 21:16:23 [NOTICE] [configure] 900 out of 1679 done.
    2021-01-15 21:16:23 [NOTICE] [configure] 1000 out of 1682 done.
    2021-01-15 21:16:42 [NOTICE] [configure] 1100 out of 1682 done.
    2021-01-15 21:16:44 [NOTICE] [configure] 1200 out of 1684 done.
    2021-01-15 21:16:44 [NOTICE] [configure] 1300 out of 1684 done.
    2021-01-15 21:16:44 [NOTICE] [configure] 1400 out of 1684 done.
    2021-01-15 21:17:07 [NOTICE] [configure] 1500 out of 1684 done.
    2021-01-15 21:17:09 [NOTICE] [configure] 1600 out of 1684 done.
    2021-01-15 21:17:13 [ERROR ] [configure] /Stage[main]/Foreman_proxy::Register/Foreman_smartproxy[centos7-foreman-proxy-nightly.mercury.example.com]: Could not evaluate:
    2021-01-15 21:17:13 [ERROR ] [configure] SmartProxy centos7-foreman-proxy-nightly.mercury.example.com cannot be retrieved from Foreman:
    2021-01-15 21:17:13 [ERROR ] [configure] Error making GET request to https://centos7-katello-3-18.mercury.example.com/api/v2/smart_proxies
    2021-01-15 21:17:13 [ERROR ] [configure] 503 Service Unavailable: The webserver was unable to reach the backend service. Is foreman.service running?
    2021-01-15 21:17:13 [ERROR ] [configure] /Stage[main]/Foreman_proxy::Register/Foreman_smartproxy[centos7-foreman-proxy-nightly.mercury.example.com]: Failed to call refresh:
    2021-01-15 21:17:13 [ERROR ] [configure] SmartProxy centos7-foreman-proxy-nightly.mercury.example.com cannot be retrieved from Foreman:
    2021-01-15 21:17:13 [ERROR ] [configure] Error making GET request to https://centos7-katello-3-18.mercury.example.com/api/v2/smart_proxies
    2021-01-15 21:17:13 [ERROR ] [configure] 503 Service Unavailable: The webserver was unable to reach the backend service. Is foreman.service running?
    2021-01-15 21:17:13 [ERROR ] [configure] /Stage[main]/Foreman_proxy::Register/Foreman_smartproxy[centos7-foreman-proxy-nightly.mercury.example.com]:
    2021-01-15 21:17:13 [ERROR ] [configure] SmartProxy centos7-foreman-proxy-nightly.mercury.example.com cannot be retrieved from Foreman:
    2021-01-15 21:17:13 [ERROR ] [configure] Error making GET request to https://centos7-katello-3-18.mercury.example.com/api/v2/smart_proxies
    2021-01-15 21:17:13 [ERROR ] [configure] 503 Service Unavailable: The webserver was unable to reach the backend service. Is foreman.service running?
    2021-01-15 21:17:15 [NOTICE] [configure] System configuration has finished.
    2021-01-15 21:17:15 [NOTICE] [post] Executing hooks in group post
    2021-01-15 21:17:15 [NOTICE] [post] All hooks in group post finished
```

ehelms:
The broken up error lines is a little hard to grok, especially when a line ends with `:` but then nothing directly follows. I'd have to see it on one line wrapping text to compare which look sbetter though. 

Any idea why it repeats the error message 3 times? Is that the provider code or Kafo?

wbclark:
> The broken up error lines is a little hard to grok, especially when a line ends with `:` but then nothing directly follows. I'd have to see it on one line wrapping text to compare which look sbetter though.
> 
> Any idea why it repeats the error message 3 times? Is that the provider code or Kafo?

My understanding is that it is repeated due to the provider being used multiple times. Originally we would see something like:

```
    2021-01-06 19:14:53 [ERROR ] [configure] /Stage[main]/Foreman_proxy::Register/Foreman_smartproxy[centos7-foreman-proxy-nightly.mercury.example.com]: Could not evaluate: Proxy centos7-foreman-proxy-nightly.mercury.example.com cannot be retrieved: unknown error (response 503)
    2021-01-06 19:14:53 [ERROR ] [configure] /Stage[main]/Foreman_proxy::Register/Foreman_smartproxy[centos7-foreman-proxy-nightly.mercury.example.com]: Failed to call refresh: Proxy centos7-foreman-proxy-nightly.mercury.example.com cannot be retrieved: unknown error (response 503)
    2021-01-06 19:14:53 [ERROR ] [configure] /Stage[main]/Foreman_proxy::Register/Foreman_smartproxy[centos7-foreman-proxy-nightly.mercury.example.com]: Proxy centos7-foreman-proxy-nightly.mercury.example.com cannot be retrieved: unknown error (response 503)
    ```
    
  And you can notice the slight variation in language like `Could not evaluate` and `Failed to call refresh`. With that said I'd have to look in more detail at how the provider is used to understand it better, which I have not done yet.
  
  Regarding your other question, without splitting the error across multiple lines, it prints output like:
  
  ```
      2021-01-15 22:00:30 [NOTICE] [configure] Starting system configuration.
      The total number of configuration tasks may increase during the run.
      Observe logs or specify --verbose-log-level to see individual configuration tasks.
    2021-01-15 22:00:58 [NOTICE] [configure] 100 out of 1657 done.
    2021-01-15 22:01:14 [NOTICE] [configure] 200 out of 1657 done.
    2021-01-15 22:02:17 [NOTICE] [configure] 300 out of 1657 done.
    2021-01-15 22:02:56 [NOTICE] [configure] 400 out of 1657 done.
    2021-01-15 22:02:59 [NOTICE] [configure] 500 out of 1658 done.
    2021-01-15 22:03:00 [NOTICE] [configure] 600 out of 1658 done.
    2021-01-15 22:03:09 [NOTICE] [configure] 700 out of 1658 done.
    2021-01-15 22:03:11 [NOTICE] [configure] 800 out of 1676 done.
    2021-01-15 22:03:21 [NOTICE] [configure] 900 out of 1679 done.
    2021-01-15 22:03:22 [NOTICE] [configure] 1000 out of 1682 done.
    2021-01-15 22:03:39 [NOTICE] [configure] 1100 out of 1682 done.
    2021-01-15 22:03:42 [NOTICE] [configure] 1200 out of 1684 done.
    2021-01-15 22:03:43 [NOTICE] [configure] 1300 out of 1684 done.
    2021-01-15 22:03:43 [NOTICE] [configure] 1400 out of 1684 done.
    2021-01-15 22:04:04 [NOTICE] [configure] 1500 out of 1684 done.
    2021-01-15 22:04:06 [NOTICE] [configure] 1600 out of 1684 done.
    2021-01-15 22:04:10 [ERROR ] [configure] /Stage[main]/Foreman_proxy::Register/Foreman_smartproxy[centos7-foreman-proxy-nightly.mercury.example.com]: Could not evaluate: SmartProxy centos7-foreman-proxy-nightly.mercury.example.com cannot be retrieved from Foreman: Error making GET request to https://centos7-katello-3-18.mercury.example.com/api/v2/smart_proxies: 503 Service Unavailable: The webserver was unable to reach the backend service. Is foreman.service running?
    2021-01-15 22:04:10 [ERROR ] [configure] /Stage[main]/Foreman_proxy::Register/Foreman_smartproxy[centos7-foreman-proxy-nightly.mercury.example.com]: Failed to call refresh: SmartProxy centos7-foreman-proxy-nightly.mercury.example.com cannot be retrieved from Foreman: Error making GET request to https://centos7-katello-3-18.mercury.example.com/api/v2/smart_proxies: 503 Service Unavailable: The webserver was unable to reach the backend service. Is foreman.service running?
    2021-01-15 22:04:10 [ERROR ] [configure] /Stage[main]/Foreman_proxy::Register/Foreman_smartproxy[centos7-foreman-proxy-nightly.mercury.example.com]: SmartProxy centos7-foreman-proxy-nightly.mercury.example.com cannot be retrieved from Foreman: Error making GET request to https://centos7-katello-3-18.mercury.example.com/api/v2/smart_proxies: 503 Service Unavailable: The webserver was unable to reach the backend service. Is foreman.service running?
    2021-01-15 22:04:12 [NOTICE] [configure] System configuration has finished.
    2021-01-15 22:04:12 [NOTICE] [post] Executing hooks in group post
    2021-01-15 22:04:12 [NOTICE] [post] All hooks in group post finished
```

ekohl:
I agree that multiple lines is harder to correlate. In general the error messages are on a single line which makes it easier to correlate. I think there's too much information in the line now.

Since the resource name is already in there, perhaps it could be reduced to `SmartProxy cannot be retrieved from Foreman` instead of including the name.

Then the part "Error making GET request to $URL" is also a bit redundant since we already know there was an error.

I'm wondering if we really need to explain HTTP errors completely. Perhaps that part can be on a new line?

So I'm thinking for unknown errors: "Smart Proxy cannot be retrieved from Foreman: GET $URL returned HTTP $CODE\nEXPLANATION FOR HTTP CODE".

wbclark:
The reason I want to include something like `Error making GET request to $URL` is that it's a lot less ambiguous than `SmartProxy cannot be retrieved from Foreman`.

Specifically including the $VERB and $URL has value for less technically savvy users - particular for downstream users that might have no idea at all what Foreman is or what a SmartProxy is or what it means to retrieve it - by informing what is it exactly that we tried to do (make a $VERB request), and where exactly they need to look (in the downstream case, satellite.example.com) to better understand what went wrong.

ehelms:
For me, if I try to look at the messages here is what I feel I need:

Current:
```

  2021-01-15 22:04:10 [ERROR ] [configure] /Stage[main]/Foreman_proxy::Register/Foreman_smartproxy[centos7-foreman-proxy-nightly.mercury.example.com]: Could not evaluate: SmartProxy centos7-foreman-proxy-nightly.mercury.example.com cannot be retrieved from Foreman: Error making GET request to https://centos7-katello-3-18.mercury.example.com/api/v2/smart_proxies: 503 Service Unavailable: The webserver was unable to reach the backend service. Is foreman.service running?
```

New:
```
  2021-01-15 22:04:10 [ERROR ] [configure] /Stage[main]/Foreman_proxy::Register/Foreman_smartproxy[centos7-foreman-proxy-nightly.mercury.example.com]: Error making GET request to Foreman at https://centos7-katello-3-18.mercury.example.com/api/v2/smart_proxies: 503 Service Unavailable: The webserver was unable to reach the backend service. Is foreman.service running?
``

From this I know the status code, I know the URL having the problem, the target hostname, and a message that points me to what I should look at as a likely culprit for the problem.

File: lib/puppet/provider/foreman_resource/rest_v3.rb (hunk 9 above)
Comment (ekohl) [reply lib/puppet/provider/foreman_resource/rest_v3.rb:109]: In vanilla Foreman you see `Smart Proxies` in the menu so there it's all consistent. The API endpoint is also `/smart_proxies`.

Arguably you could say that Foreman Proxy is just the default implementation but in theory any service could implement it. For example, a DHCP appliance can implement it and be registered directly. For Pulp you can also write a plugin `pulpcore_smart_proxy` to replace `smart_proxy_pulp`.

**Comment by ekohl** [comment]
> The reason I want to include something like `Error making GET request to $URL` is that it's a lot less ambiguous than `SmartProxy cannot be retrieved from Foreman`.
> 
> Specifically including the $VERB and $URL has value for less technically savvy users - particular for downstream users that might have no idea at all what Foreman is or what a SmartProxy is or what it means to retrieve it - by informing what is it exactly that we tried to do (make a $VERB request), and where exactly they need to look (in the downstream case, satellite.example.com) to better understand what went wrong.

In my suggestion I didn't leave VERB and URL out. I just shortened it. First a human readable part, then `:` and the detailed error.

File: lib/puppet/provider/foreman_smartproxy/rest_v3.rb [hunk 11]
```diff
@@ -3,8 +3,19 @@
...
-      r = request(:get, 'api/v2/smart_proxies', :search => %{name="#{resource[:name]}"})
-      raise Puppet::Error.new("Proxy #{resource[:name]} cannot be retrieved: #{error_message(r)}") unless success?(r)
+      path = 'api/v2/smart_proxies'
+      r = request(:get, path, :search => %{name="#{resource[:name]}"})
+
+      unless success?(r)
+        error_string = <<~ERROR.chomp
+
+          SmartProxy #{resource[:name]} cannot be retrieved from Foreman:
+          Error making GET request to #{request_uri(path)}
```
ekohl, line 13:
```suggestion
          Error making GET request to #{r.uri)}
```

lib/puppet/provider/foreman_resource/rest_v3.rb
```diff
@@ -95,6 +98,21 @@ def success?(response)
   end
 
   def error_message(response)
-    JSON.parse(response.body)['error']['full_messages'].join(' ') rescue "unknown error (response #{response.code})"
+    messages = {
+      '400' => '400 Bad Request: Something is wrong with the data we sent to Foreman server',
```
ekohl, line 102: Probably better to extract it a bit, but you can do:

```suggestion
      '400' => "#{response.code} #{response.message}: Something is wrong with the data we sent to Foreman server",
```

That way you can provide it for every error code even if there is no detailed explanation.

lib/puppet/provider/foreman_resource/rest_v3.rb
```diff
@@ -95,6 +98,21 @@ def success?(response)
...
+      '401' => '401 Unauthorized Request: Check credentials for validity',
+      '403' => '403 Forbidden: The credentials were valid but do not grant access to the requested resource',
```
ekohl, line 104: This is highly unlikely. oauth has full access and there are no permission checks. I don't think we should document this. Just keep it generic:
```suggestion
      '403' => "#{response.code} #{response.message}"
```

Probably even:
```suggestion
      default => "#{response.code} #{response.message}"
```

lib/puppet/provider/foreman_resource/rest_v3.rb
```diff
@@ -95,6 +98,21 @@ def success?(response)
...
+      '500' => '500 Internal Server Error: Check /var/log/foreman/production.log on Foreman server for detailed information',
+      '502' => '502 Bad Gateway: The webserver received an invalid response from the application server. Was Foreman unable to handle the request?',
+      '503' => '503 Service Unavailable: The webserver was unable to reach the backend service. Is foreman.service running?',
+      '504' => '504 Gateway Timeout: The webserver timed out waiting for a response from the application. Is Foreman under unusually heavy load?'
+    }
+
+    if messages.include?(response.code.to_str)
+      messages[response.code.to_str]
+    else
+      JSON.parse(response.body)['error']['full_messages'].join(' ') rescue "unknown error (response #{response.code})"
```
ekohl, line 115:
```suggestion
      JSON.parse(response.body)['error']['full_messages'].join(' ') rescue "unknown error (response #{response.code} #{response.message})"
```

lib/puppet/provider/foreman_smartproxy/rest_v3.rb
```diff
@@ -3,8 +3,19 @@
...
   def proxy
     @proxy ||= begin
-      r = request(:get, 'api/v2/smart_proxies', :search => %{name="#{resource[:name]}"})
-      raise Puppet::Error.new("Proxy #{resource[:name]} cannot be retrieved: #{error_message(r)}") unless success?(r)
+      path = 'api/v2/smart_proxies'
+      r = request(:get, path, :search => %{name="#{resource[:name]}"})
+
+      unless success?(r)
+        error_string = <<~ERROR.chomp
+
```
ekohl, line 11: Why this empty line?

lib/puppet/provider/foreman_resource/rest_v3.rb
```diff
@@ -54,11 +54,14 @@ def generate_token
     OAuth::AccessToken.new(oauth_consumer)
   end
 
-  def request(method, path, params = {}, data = nil, headers = {})
+  def request_uri(path)
```
ekohl, line 57: You can use `response.uri` so this isn't needed.

ekohl (changes requested):

lib/puppet/provider/foreman_smartproxy/rest_v3.rb
wbclark: I already tried this previously and it did not work. `#{r.uri}` would evaluate to an empty string in my tests. This is the reason I introduced the `request_uri` method.

lib/puppet/provider/foreman_smartproxy/rest_v3.rb
wbclark: The effect is to split this into two lines, which I personally found to be more readable:

```
    2021-01-15 21:17:13 [ERROR ] [configure] /Stage[main]/Foreman_proxy::Register/Foreman_smartproxy[centos7-foreman-proxy-nightly.mercury.example.com]: Could not evaluate:
    2021-01-15 21:17:13 [ERROR ] [configure] SmartProxy centos7-foreman-proxy-nightly.mercury.example.com cannot be retrieved from Foreman:
```

lib/puppet/provider/foreman_resource/rest_v3.rb
wbclark: The reason I introduced this method in the first place is that `#{r.uri}` would evaluate to an empty string in my tests. Let me know if you have a suggestion to fix that issue and if so I'll revert it.

Specifically, the `request` method returns something like `response = oauth_consumer.request(...)` where the oath_consumer is `OAuth::Consumer.new(...)`

So ultimately `r.uri` should be `OAuth::Consumer.new(...).request(...).uri` and what's notable to me is that when I tried it, it didn't throw any `NoMethodError` here, it simply returned empty.

wbclark:
> For me, if I try to look at the messages here is what I feel I need:
> 
> Current:
> 
> ```
> 
>   2021-01-15 22:04:10 [ERROR ] [configure] /Stage[main]/Foreman_proxy::Register/Foreman_smartproxy[centos7-foreman-proxy-nightly.mercury.example.com]: Could not evaluate: SmartProxy centos7-foreman-proxy-nightly.mercury.example.com cannot be retrieved from Foreman: Error making GET request to https://centos7-katello-3-18.mercury.example.com/api/v2/smart_proxies: 503 Service Unavailable: The webserver was unable to reach the backend service. Is foreman.service running?
> ```
> 
> New:
> 
> ```
>   2021-01-15 22:04:10 [ERROR ] [configure] /Stage[main]/Foreman_proxy::Register/Foreman_smartproxy[centos7-foreman-proxy-nightly.mercury.example.com]: Error making GET request to Foreman at https://centos7-katello-3-18.mercury.example.com/api/v2/smart_proxies: 503 Service Unavailable: The webserver was unable to reach the backend service. Is foreman.service running?
> ```
> 
> From this I know the status code, I know the URL having the problem, the target hostname, and a message that points me to what I should look at as a likely culprit for the problem.

wbclark:
The close and re-open was a mis-click on quote-reply.... -_-

ekohl:
Could you link it to https://projects.theforeman.org/issues/30284?

wbclark:
> New:
> 
> ```
>   2021-01-15 22:04:10 [ERROR ] [configure] /Stage[main]/Foreman_proxy::Register/Foreman_smartproxy[centos7-foreman-proxy-nightly.mercury.example.com]: Error making GET request to Foreman at https://centos7-katello-3-18.mercury.example.com/api/v2/smart_proxies: 503 Service Unavailable: The webserver was unable to reach the backend service. Is foreman.service running?
> ```
> 
> From this I know the status code, I know the URL having the problem, the target hostname, and a message that points me to what I should look at as a likely culprit for the problem.

I think eliminating the `Could not evaluate:` requires a change in Kafo. The rest is doable here.

ekohl:
FYI: I still think that https://github.com/theforeman/kafo/pull/290 is the way to go for real error reporting. I suspect you have more context available to give a precise message.

lib/puppet/provider/foreman_resource/rest_v3.rb
wbclark: Thanks. I updated the general case to include the #{response.message} as well. (I also removed the `unknown error` bit).`

lib/puppet/provider/foreman_resource/rest_v3.rb
wbclark: I incorporated this idea outside of `messages` and instead at `JSON.parse(response.body)['error']['full_messages'].join(' ') rescue "Response: #{response.code} #{response.message})"`

Let me know please if that's suitable to you

wbclark:
Error output with the latest version is as below:

```
    2021-01-18 23:46:13 [NOTICE] [configure] 1600 out of 1684 done.
    2021-01-18 23:46:17 [ERROR ] [configure] /Stage[main]/Foreman_proxy::Register/Foreman_smartproxy[centos7-foreman-proxy-nightly.mercury.example.com]: Could not evaluate: Error making GET request to Foreman at https://centos7-katello-3-18.mercury.example.com/api/v2/smart_proxies: 503 Service Unavailable: The webserver was unable to reach the backend service. Is foreman.service running?
    2021-01-18 23:46:17 [ERROR ] [configure] /Stage[main]/Foreman_proxy::Register/Foreman_smartproxy[centos7-foreman-proxy-nightly.mercury.example.com]: Failed to call refresh: Error making GET request to Foreman at https://centos7-katello-3-18.mercury.example.com/api/v2/smart_proxies: 503 Service Unavailable: The webserver was unable to reach the backend service. Is foreman.service running?
    2021-01-18 23:46:17 [ERROR ] [configure] /Stage[main]/Foreman_proxy::Register/Foreman_smartproxy[centos7-foreman-proxy-nightly.mercury.example.com]: Error making GET request to Foreman at https://centos7-katello-3-18.mercury.example.com/api/v2/smart_proxies: 503 Service Unavailable: The webserver was unable to reach the backend service. Is foreman.service running?
    2021-01-18 23:46:20 [NOTICE] [configure] System configuration has finished.
```

Removing additional unnecessary language and de-duplicating error messages in foreman-installer output will require an additional change in Kafo, independent of the implementation here.

wbclark:
> FYI: I still think that [theforeman/kafo#290](https://github.com/theforeman/kafo/pull/290) is the way to go for real error reporting. I suspect you have more context available to give a precise message.

It looks like a powerful solution, but it also requires a lot more build out, so it looks like there isn't an easy path to get that done for the next release?

lib/puppet/provider/foreman_smartproxy/rest_v3.rb
wbclark: I've converted the output back to a single line

lib/puppet/provider/foreman_resource/rest_v3.rb
```diff
@@ -95,6 +98,20 @@ def success?(response)
   end
 
   def error_message(response)
-    JSON.parse(response.body)['error']['full_messages'].join(' ') rescue "unknown error (response #{response.code})"
+    messages = {
+      '400' => '400 Bad Request: Something is wrong with the data we sent to Foreman server',
```
ehelms, line 102: 'we' can be ambiguous when used in these kinds of contexts, could the message say "Something is wrong with the data sent to the Foreman server"

lib/puppet/provider/foreman_resource/rest_v3.rb
```diff
@@ -95,6 +98,20 @@ def success?(response)
...
+      '401' => '401 Unauthorized Request: Check credentials for validity',
```
ehelms, line 103: Can we call out what credentials to check? The oauth credentials? Or the client ceritificates?

ehelms:
A couple of minor comments from me. I think this is a good update to the output messages for an area that is often unclear as to what might be the underlying cause when users hit it to help users and developers more easily diagnosis and fix the problem. I think there have been some further updates outlined we could investigate for general error reporting the installer, and for me those should not hold up this improvement.

lib/puppet/provider/foreman_resource/rest_v3.rb
wbclark: Yes, I agree this is a bit of a copout.

I was able to reproduce the 401 response by introducing a typo in the `--foreman-proxy-oauth-consumer-key`. Another method to reproduce it is to introduce a typo in the `--foreman-proxy-foreman-base-url`. Using incorrect certificates should result in the same response (though I have not personally tested it).

While I don't insist that we have to implement it right this minute, I think this illustrates the use case for branded error messages. We could link here to a community wiki with detailed information on how to troubleshoot the error, while downstream could link to a KCS describing the error and troubleshooting steps.

ekohl: The only authentication that the Foreman server does is oauth credentials. AFAIK no client certificates are used so the only certificate is the server CA that's verified. If that's invalid, no HTTP(S) connection can be established in the first place. However, if they point it at a different server/service then who knows. There may also be some HTTP proxy that does odd things. Because of that last point I'm hesitant in trying to explain exactly what's going on.

lib/puppet/provider/foreman_resource/rest_v3.rb
ekohl: I don't consider this point resolved. What I meant is that on every single line you shouldn't repeat the code and response message. You can even simplify this (as I hinted at with 'extract it a bit').
```ruby
explanations = {
  '400' => 'Something is wrong with the data we sent to Foreman server',
  ...
}

if (explanation = explanations[response.code])
  "#{response.code} #{response.message}: #{explanation}"
else
  ...
end
```

lib/puppet/provider/foreman_resource/rest_v3.rb
```diff
@@ -95,6 +98,20 @@ def success?(response)
...
+      '404' => '404 Not Found: The requested resource was not found',
```
ekohl, line 104: This feels like an edge case to me. The only thing I can think of is that they entered the wrong server URL and end up on a different service.

lib/puppet/provider/foreman_resource/rest_v3.rb
wbclark: Thanks for the suggestion. I'm going to push an update once tests are passing for me that incorporates this approach

lib/puppet/provider/foreman_resource/rest_v3.rb
wbclark: The next version I push will contain the explanation `'Usually this indicates something was wrong with the Oauth credentials'`

lib/puppet/provider/foreman_resource/rest_v3.rb
wbclark: The next version I push will remove the 'we'

lib/puppet/provider/foreman_resource/rest_v3.rb
wbclark: It's definitely an edge case but I had a few more in mind when I wrote it, based on situations I've seen previously:

1. Attempted improper use of load balancing

2. "I forgot I cloned this VM and left it running on the same network"

3. Automation/scripts run amok

lib/puppet/provider/foreman_resource/rest_v3.rb
```diff
@@ -95,6 +98,20 @@ def success?(response)
...
+      '400' => 'Something is wrong with the data sent to Foreman server',
+      '401' => 'Often this is caused by invalid Oauth credentials',
+      '404' => 'The requested resource was not found',
+      '500' => 'Check /var/log/foreman/production.log on Foreman server for detailed information',
+      '502' => 'The webserver received an invalid response from the backend service. Was Foreman unable to handle the request?',
+      '503' => 'The webserver was unable to reach the backend service. Is foreman.service running?',
+      '504' => 'The webserver timed out waiting for a response from the backend service. Is Foreman under unusually heavy load?'
+    }
+
+    if explanation = explanations[response.code.to_str]
```
ekohl, line 111: The convention is to use braces when there is an assignment as a visual cue. Otherwise it's too easy to confuse with `==` and it may just be a bug/typo.
```suggestion
    if (explanation = explanations[response.code.to_str])
```

The alternative is:
```suggestion
    explanation = explanations[response.code.to_str]
    if explanation
```

ekohl (commented):
Why is it `Refs` instead of `Fixes`? What do you think is needed to close the issue?

Code wise I think this is ok.

ehelms:
Same, I do feel this 'Fixes' the issue. And other enhancements we may do can be captured by independent issues.

wbclark:
> Why is it `Refs` instead of `Fixes`? What do you think is needed to close the issue?
> 
> Code wise I think this is ok.

I'd like to also remove near duplicate lines like `Unable to Evaluate: #{some_message}` vs. `Failed to call Refresh: #{the_same_message}` from installer output

But I'm happy to treat that as a separate issue.... I'll change this one to `Fixes`

lib/puppet/provider/foreman_resource/rest_v3.rb
wbclark: Thanks, I wasn't aware of this convention.

Thinking about it some more, would this be acceptable?

```
if !(explanation = explanations[response.code.to_str]).nil?
```

The `()`s are only useful if the reader knows the convention.

wbclark: Nevermind, it's not even semantically equivalent that way...

ehelms (approved):

lib/puppet/provider/foreman_resource/rest_v3.rb
ekohl: My plan is to add Rubocop to the repo which will enforce the convention. If you don't do it and meant `if x == y`, Rubocop will ask you to change it to `if (x = y)`. That should tell you enough.

On a side note, Python thought it was a bad idea to allows assignment statements in if conditions to prevent those logic bugs. Recently it did add `if x := y` as a compromise.

That's the theory behind it.

wbclark:
Forgot to comment on Friday that I did in fact update to use the parenthesis convention for variable assignment within conditional.

@ekohl could you take a look and let me know if you are satisfied?

wbclark:
I have rebased to solve the merge conflict

ehelms:
I'd like to get this change in before I do a release of the module.

lib/puppet/provider/foreman_resource/rest_v3.rb
```diff
@@ -96,6 +99,20 @@ def success?(response)
   end
 
   def error_message(response)
-    JSON.parse(response.body)['error']['full_messages'].join(' ') rescue "unknown error (response #{response.code})"
+    explanations = {
+      '400' => 'Something is wrong with the data sent to Foreman server',
```
ekohl, line 103: Thinking out loud because it's a bit vague. How about
```suggestion
      '400' => 'Receiving server (Foreman) did not accept our data',
```

The reason I put Foreman between braces is that if it actually hits some proxy or something else in between then it may be clearer. Of course there shouldn't be anything in between but if everything went well we wouldn't be writing this code :) I believe 400 is more likely to be returned by some non-Foreman server in between. Like if a user pointed it at `https://internal-service.example.com`

lib/puppet/provider/foreman_resource/rest_v3.rb
```diff
@@ -96,6 +99,20 @@ def success?(response)
...
+      '401' => 'Often this is caused by invalid Oauth credentials',
+      '404' => 'The requested resource was not found',
+      '500' => 'Check /var/log/foreman/production.log on Foreman server for detailed information',
```
ekohl, line 106: Since [#28384](https://projects.theforeman.org/issues/28384) (Foreman 2.2) there is a rake task. If we have the request ID (available as the `X-Request-Id` header) we can recommend `REQUEST_ID=... foreman-rake errors:fetch_log` instead of the precise error. I don't have experience with this myself, but if we want to make it easy to debug we can be much more precise nowadays. This task has to run on the Foreman server but the registration can happen from any server so we can't directly run it.

Note that Apache can also return a HTTP 500 in which case you need to look in the Apache logs. That was more common with Passenger (for example, when Rails failed to even boot) but may still happen.

ekohl (approved):
Inline there are some recommendations for more actionable error messages but I'm ok merging this and improving it later on.

lib/puppet/provider/foreman_resource/rest_v3.rb
wbclark: Thanks. I'm not fully satisfied with that wording either (and I'm not sure that I'll ever be).

Thinking longer term, I think the best approach is to have links in these explanation messages to community wikis that can be much more comprehensive in covering various cases, with the option of downstream releases replacing those links with more specifically applicable ones for their community. Often the workflow for an user encountering one of these errors will be to copy and paste it into a search engine, so we might as well remove that step for them and just send them directly to the relevant webpage where we can really cover all possible causes in order of likelihood.

I'd like to consider this as a parking lot item to revisit in the future.

wbclark:
Thanks @ekohl . I'm not fully satisfied with that wording either (and I'm not sure that I'll ever be).

Thinking longer term, I think the best approach is to have links in these explanation messages to community wikis that can be much more comprehensive in covering various cases, with the option of downstream releases replacing those links with more specifically applicable ones for their community. Often the workflow for an user encountering one of these errors will be to copy and paste it into a search engine, so we might as well remove that step for them and just send them directly to the relevant webpage where we can really cover all possible causes in order of likelihood.

I'd like to consider this as a parking lot item to revisit in the future. For now, I'll merge it considering it's still an improvement over the current state.

PR #914, opened 2021-01-15
Fixes #32325 - Deploy facts needed for determining if a host is a foreman

adamruzicka:
Part of https://community.theforeman.org/t/infrastructure-roles/22001

### DIFF
--- a/lib/puppet/provider/foreman_smartproxy/rest_v3.rb
+++ b/lib/puppet/provider/foreman_smartproxy/rest_v3.rb
@@ -24,7 +24,7 @@ Puppet::Type.type(:foreman_smartproxy).provide(:rest_v3, :parent => Puppet::Type
   end
 
   def create
-    post_data = {:smart_proxy => {:name => resource[:name], :url => resource[:url]}}.to_json
+    post_data = {:smart_proxy => {:name => resource[:name], :url => resource[:url], :instance_id => resource[:instance_id]}}.to_json
     path = 'api/v2/smart_proxies'
     r = request(:post, path, {}, post_data)
 
--- a/lib/puppet/provider/foreman_smartproxy/rest_v3.rb
+++ b/lib/puppet/provider/foreman_smartproxy/rest_v3.rb
@@ -24,7 +24,9 @@ Puppet::Type.type(:foreman_smartproxy).provide(:rest_v3, :parent => Puppet::Type
   end
 
   def create
-    post_data = {:smart_proxy => {:name => resource[:name], :url => resource[:url], :instance_id => resource[:instance_id]}}.to_json
+    proxy_params = {:name => resource[:name], :url => resource[:url]}
+    proxy_params[:instance_id] = resource[:instance_id] unless resource[:instance_id].nil?
+    post_data = {:smart_proxy => proxy_params}.to_json
     path = 'api/v2/smart_proxies'
     r = request(:post, path, {}, post_data)
 
--- a/lib/puppet/type/foreman_smartproxy.rb
+++ b/lib/puppet/type/foreman_smartproxy.rb
@@ -52,6 +52,10 @@ Puppet::Type.newtype(:foreman_smartproxy) do
     newvalues(URI.regexp)
   end
 
+  newparam(:instance_id) do
+    desc 'Instance id of the smartproxy'
+  end
+
   newparam(:timeout) do
     desc "Timeout for HTTP(s) requests"
 
--- a/manifests/config.pp
+++ b/manifests/config.pp
@@ -35,6 +35,7 @@ class foreman::config {
     mode  => '0640',
   }
 
+  $uuid_fact_signature = foreman::hmac_signature($foreman::oauth_consumer_secret, $foreman::instance_id)
   file { 'foreman_uuid.json':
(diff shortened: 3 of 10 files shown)


ehelms:
I ended up putting some comments in https://github.com/theforeman/puppet-foreman_proxy/pull/641 that I think apply here as well around idempotency.

manifests/config.pp
```diff
@@ -35,6 +35,15 @@
     mode  => '0640',
   }
 
+  $uuid_fact_signature = foreman::hmac_signature($foreman::oauth_consumer_secret, $foreman::instance_id)
+  file { 'foreman_uuid.json':
+    path    => '/opt/puppetlabs/facter/facts.d/foreman_uuid.json',
```
ehelms, line 40: Is this a safe path to assume @ekohl ? Or does this need a switcher on AIO similar to puppet_ssldir? (https://github.com/theforeman/puppet-foreman/pull/914/files#diff-e78db79abb0b543310767fb9d2814cc4e2408f613de4985186251c70bdd4edbbR123)

lib/puppet/type/foreman_smartproxy.rb
```diff
@@ -52,6 +52,10 @@ def is_to_s(value)
     newvalues(URI.regexp)
   end
 
+  newparam(:instance_id) do
+    desc 'Instance id of the smartproxy'
+  end
```
ekohl, line 57: I don't see a provider part to this. Where does this actually have an effect?

adamruzicka: It is used here https://github.com/theforeman/puppet-foreman_proxy/pull/641/files#diff-f9877574c121f639214a4b7e1b0d6ecf4f988fa27f57ba33cf75717c561f54e7R13

adamruzicka: Right, changed.

lib/puppet/provider/foreman_smartproxy/rest_v3.rb
```diff
@@ -24,7 +24,7 @@ def exists?
   end
 
   def create
-    post_data = {:smart_proxy => {:name => resource[:name], :url => resource[:url]}}.to_json
+    post_data = {:smart_proxy => {:name => resource[:name], :url => resource[:url], :instance_id => resource[:instance_id]}}.to_json
```
ekohl, line 27: I guess it will fail and break compatibility with Foreman which don't accept this parameter. We do keep compatibility with older versions so that users can first upgrade their Puppet modules and then the actual instance. I also use it to run git master on my Foreman 2.4 server to find problems.

How about we only pass the instance ID if it's not nil? That allows users to not pass an instance ID on older versions.

ekohl (changes requested):
You also need to implement `instance_id` and `instance_id=` methods to allow updating the attribute.

manifests/config.pp
adamruzicka: According to this there are three paths where facts can be placed by default and those are 
```
/opt/puppetlabs/facter/facts.d/
/etc/puppetlabs/facter/facts.d/
/etc/facter/facts.d/
```

Out of those, only the one in `/opt` is fully present on a foreman machine, for the other two we'd have to manage last two levels of directories.

[1] - https://puppet.com/docs/puppet/7.5/external_facts.html#executable-fact-locations

manifests/init.pp
```diff
@@ -178,6 +178,8 @@
 #
 # $instance_id::                  Foreman instance id
 #
+# $expose_instance_id::           Expose Foreman instance id
```
ekohl, line 181:
```suggestion
# $expose_instance_id::           Deploy a custom fact to expose the Foreman instance id
```

manifests/config.pp
```diff
@@ -35,13 +35,15 @@
...
-  $uuid_fact_signature = foreman::hmac_signature($foreman::oauth_consumer_secret, $foreman::instance_id)
-  file { 'foreman_uuid.json':
-    path    => '/opt/puppetlabs/facter/facts.d/foreman_uuid.json',
-    owner   => 'root',
-    group   => 'root',
-    mode    => '0440',
-    content => to_json({'foreman_uuid' => $foreman::instance_id, 'foreman_uuid_signature' => $uuid_fact_signature}),
+  if $foreman::expose_instance_id {
+    $uuid_fact_signature = foreman::hmac_signature($foreman::oauth_consumer_secret, $foreman::instance_id)
+    file { 'foreman_uuid.json':
```
ekohl, line 40: Instead of an `if`, I prefer to explicitly manage it:
```suggestion
    file { 'foreman_uuid.json':
      ensure => bool2str($foreman::expose_instance_id, 'file', 'absent'),
```

That allows users to clean it up again.

ekohl (changes requested):

adamruzicka:
Is there anythng I can do to make this move?

ekohl:
I need to find some time to properly review this. I have some doubts about the security aspect. However, right now I'm very busy with the release itself and making sure it works.

ares:
Any update here as the release is getting to the final phase?

adamruzicka:
Closing in favor of #965

PR #915, opened 2021-01-18
Fixes #31670 - don't timeout when running db:migrate

evgeni:


manifests/database.pp
```diff
@@ -22,7 +22,8 @@
     }
 
     foreman::rake { 'db:migrate':
-      unless => '/usr/sbin/foreman-rake db:abort_if_pending_migrations',
+      timeout => 0,
```
evgeni, line 25: I am not 100% sure 0 is a good value here, but couldn't think of anything else.

ekohl: You could make it a parameter on the (private) class so you at least have a way to change it as a local workaround.

evgeni: You mean a param of the `foreman::database` class?

ekohl: Exactly.

evgeni: So, like this?

ehelms (approved):

manifests/database.pp
```diff
@@ -1,5 +1,8 @@
-# Set up the foreman database
-class foreman::database {
+# @summary Set up the foreman database
+# @api private
+class foreman::database(
+  Integer $timeout = 0,
```
ekohl, line 4: You can set a lower bound (and upper bound, but I don't think that makes sense here)
```suggestion
  Integer[0] $timeout = 0,
```

ekohl (approved):

manifests/database.pp
evgeni: done

evgeni:
I've also bumped the timeout to 3600 seconds (1h), as discussed earlier.

@ehelms @ekohl ACk still holds?

ekohl (approved):

PR #918, opened 2021-02-23
Update Github Actions workflow

ekohl:
This applies the lessons learned while applying this to Voxpupuli and The Foreman orgs.


ekohl:
Looks like this is broken on Puppet 5 because Redis uses camptocamp/systemd where master dropped Puppet 5. Another reason is that I dropped beaker-hiera from the Gemfile which breaks things. I'll try to get a gem release out now that it's transferred to Voxpupuli.

PR #922, opened 2021-03-23
Fixes #32175: Allow toggling task backup when cleaning them up

ehelms:
Changes the default behavior to disable task backup when enabling
the task cleanup cron job. Users who still want backups kept
on disk will need to set the parameter to true.

### DIFF
--- a/manifests/plugin/tasks.pp
+++ b/manifests/plugin/tasks.pp
@@ -6,9 +6,13 @@
 # @param cron_line
 #   Cron line defining when the cleanup cron job should run
 #
+# @param backup
+#   Enable creating a backup of cleaned up tasks in CSV format when automatic_cleanup is enabled
+#
 class foreman::plugin::tasks(
   Boolean $automatic_cleanup = false,
   String $cron_line = '45 19 * * *',
+  Boolean $backup = false,
 ) {
   foreman::plugin { 'tasks':
     package => $foreman::plugin_prefix.regsubst(/foreman[_-]/, 'foreman-tasks'),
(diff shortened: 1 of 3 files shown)


ehelms:
Configuration for task-backups are also guided by https://github.com/theforeman/foreman-tasks/blob/master/config/foreman-tasks.yaml.example which is laid down by the RPM and not under Puppets control. This focuses on making it so that the cron job does not by default create backups as the environment variables override the base config. If a user chooses to run the cleanup by hand they have to adjust any specific configuration differences they would want.

Putting https://github.com/theforeman/foreman-tasks/blob/master/config/foreman-tasks.yaml.example under puppet control seemed like a much bigger change and handling the `:rules:` section as tricky given hashes are never fun and are further complicated when translated to installer parameters.

ehelms:
Needs https://github.com/theforeman/puppet-foreman/pull/923

manifests/plugin/tasks.pp
```diff
@@ -6,9 +6,13 @@
 # @param cron_line
 #   Cron line defining when the cleanup cron job should run
 #
+# @param backup
+#   Enable creating a backup of cleaned up tasks in CSV format
```
ekohl, line 10: Perhaps it should be noted that this is only relevant if `automatic_cleanup` is true? It's sort of there, but being explicit doesn't hurt.

ekohl (commented):
Small nit, otherwise :+1:

ehelms:
@adamruzicka Take a look if you have time, in case you have concerns around this approach and reasoning

adamruzicka:
I'm fine with this approach, +1 from me

ekohl (approved):

ekohl:
I'm debating whether this is an enhancement or breaking change. It does change behavior so technically it would be. On the other hand, it is a plugin. I also think the next release should be major anyway since I want to finish https://github.com/theforeman/puppet-foreman/pull/862.

ehelms:
> I'm debating whether this is an enhancement

:+1:

PR #924, opened 2021-03-23
use deb gpg key from our server, not the gpg network

evgeni:
also ensure the key is refreshed if expired

### DIFF
--- a/manifests/repos/apt.pp
+++ b/manifests/repos/apt.pp
@@ -2,27 +2,34 @@
 # @api private
 define foreman::repos::apt (
   Variant[Enum['nightly'], Pattern['^\d+\.\d+$']] $repo,
-  Variant[String, Hash] $key = 'AE0AF310E2EA96B6B6F4BD726F8600B9563278F6',
+  String $key = 'AE0AF310E2EA96B6B6F4BD726F8600B9563278F6',
+  Stdlib::HTTPUrl $key_location = 'https://deb.theforeman.org/foreman.asc',
   Stdlib::HTTPUrl $location = 'https://deb.theforeman.org/',
 ) {
   include apt
 
-  ::apt::source { $name:
+  apt::key { $name:
+    ensure => refreshed,
+    id     => $key,
+    source => $key_location,
+  }
+
+  apt::source { $name:
     repos    => $repo,
     location => $location,
-    key      => $key,
     include  => {
       src => false,
     },
+    require  => Apt::Key['foreman'],
   }
 
-  ::apt::source { "${name}-plugins":
+  apt::source { "${name}-plugins":
     release  => 'plugins',
     repos    => $repo,
     location => $location,
-    key      => $key,
     include  => {
       src => false,
     },
+    require  => Apt::Key['foreman'],
   }
 }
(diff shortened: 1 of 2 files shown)


evgeni:
I do wonder if we should just use `apt::key` directly once, instead of twice in the definition of the repos.

manifests/repos/apt.pp
```diff
@@ -10,7 +10,11 @@
   ::apt::source { $name:
     repos    => $repo,
     location => $location,
-    key      => $key,
+    key      => {
+      id     => $key,
+      source => 'http://deb.theforeman.org/foreman.asc',
+      ensure => 'refreshed',
```
ekohl, line 16: Does this mean it fetches the key on every run?

manifests/repos/apt.pp
```diff
@@ -10,7 +10,11 @@
   ::apt::source { $name:
     repos    => $repo,
     location => $location,
-    key      => $key,
+    key      => {
+      id     => $key,
+      source => 'http://deb.theforeman.org/foreman.asc',
```
ekohl, line 15: Could you make the key also a parameter? Also wondering if we can fetch it over https so there is at least integrity.

ekohl:
> I do wonder if we should just use `apt::key` directly once, instead of twice in the definition of the repos.

That probably makes sense.

manifests/repos/apt.pp
evgeni: according to https://forge.puppet.com/modules/puppetlabs/apt/reference#ensure-1:

Using 'refreshed' will make keys auto update when they have expired (assuming a new key exists on the key server).

manifests/repos/apt.pp
```diff
@@ -4,25 +4,32 @@
   Variant[Enum['nightly'], Pattern['^\d+\.\d+$']] $repo,
   Variant[String, Hash] $key = 'AE0AF310E2EA96B6B6F4BD726F8600B9563278F6',
   Stdlib::HTTPUrl $location = 'https://deb.theforeman.org/',
+  Stdlib::HTTPUrl $key_location = 'https://deb.theforeman.org/foreman.asc',
 ) {
   include apt
 
+  ::apt::key { $name:
```
ekohl, line 11: We didn't fix this in linting, but `::` is no longer needed
```suggestion
  apt::key { $name:
```

manifests/repos/apt.pp
```diff
@@ -4,25 +4,32 @@
   Variant[Enum['nightly'], Pattern['^\d+\.\d+$']] $repo,
   Variant[String, Hash] $key = 'AE0AF310E2EA96B6B6F4BD726F8600B9563278F6',
```
ekohl, line 5: Due to the new structure, I think this is now just a `String`:
```suggestion
  String $key = 'AE0AF310E2EA96B6B6F4BD726F8600B9563278F6',
```

manifests/repos/apt.pp
```diff
@@ -4,25 +4,32 @@
...
   Stdlib::HTTPUrl $location = 'https://deb.theforeman.org/',
+  Stdlib::HTTPUrl $key_location = 'https://deb.theforeman.org/foreman.asc',
```
ekohl, line 7: Mind keeping this with `$key` in the argument sorting?

ekohl (commented):
I merged the systemd pin PR so if you update, please rebase.

ehelms:
Pin camptocamp has been merged, so now you can drop it :)

ekohl (approved):

ekohl:
Technically this changes a parameter but since it's a private API I don't consider it breaking, hence an enhancement.

PR #926, opened 2021-03-25
Fixes #32208 - accept trailing slash in Krb auth url

ezr-ondrej:
The Kerberos url is quite strict, but accepting trailing `/` is probably not bad and can avoid a lot of confusion

### DIFF
--- a/templates/lookup_identity.conf.erb
+++ b/templates/lookup_identity.conf.erb
@@ -1,5 +1,5 @@
 
-<LocationMatch ^/users/(ext)?login$>
+<LocationMatch ^/users/(ext)?login/?$>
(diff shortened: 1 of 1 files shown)


ekohl (commented):
Does Rails accept a trailing slash for this route?

ezr-ondrej:
> Does Rails accept a trailing slash for this route?

Yes, Rails doesn't care about those, it silently builds the unique URL. So it is very confusing why it's not working, because Rails are landing on a correct action, just without the correct headers (even negotiate happens, just not correctly, because it is forced by our `curl --negotiate` test command.

ekohl (approved):

PR #919, opened 2021-02-25
Fixes #31964 - Assign equal weight to sidekiq queues

adamruzicka:
The previous approach worker, but it made sidekiq default worker process all
items from the default queue before moving to remote execution queue. This
change adds equal weight to the queues, making sidekiq pick jobs from both.

PR #927, opened 2021-03-25
Fixes #31964 - Assign equal weight to sidekiq queues

ekohl:
The previous approach worker, but it made sidekiq default worker process all items from the default queue before moving to remote execution queue.  This change adds equal weight to the queues, making sidekiq pick jobs from both.

To achieve this a function is introduced to convert data to YAML. Foreman typically uses symbolized keys in config files. This is hard to achieve with pure Puppet so a Ruby function is used.

Replaces https://github.com/theforeman/puppet-foreman/pull/919

### DIFF
--- a/lib/puppet/functions/foreman/to_symbolized_yaml.rb
+++ b/lib/puppet/functions/foreman/to_symbolized_yaml.rb
@@ -1,0 +1,38 @@
+# frozen_string_literal: true
+
+require 'yaml'
+# @summary
+#   Convert a data structure and output it as YAML while symbolizing keys
+#
+# In Foreman often YAML files have symbols as keys. Since it's hard to do that
+# from Puppet, this function does it for you.
+#
+# @example How to output YAML
+#   # output yaml to a file
+#     file { '/tmp/my.yaml':
+#       ensure  => file,
+#       content => foreman::to_symbolized_yaml($myhash),
+#     }
+# @example Use options control the output format
+#   file { '/tmp/my.yaml':
+#     ensure  => file,
+#     content => foreman::to_symbolized_yaml($myhash, {indentation: 4})
+#   }
+Puppet::Functions.create_function(:'foreman::to_symbolized_yaml') do
+  # @param data
+  # @param options
+  #
+  # @return [String]
+  dispatch :to_symbolized_yaml do
+    param 'Any', :data
+    optional_param 'Hash', :options
+  end
+
+  def to_symbolized_yaml(data, options = {})
+    if data.is_a?(Hash)
+      data = Hash[data.map { |k, v| [k.to_sym, v] }]
+    end
+
+    data.to_yaml(options)
+  end
+end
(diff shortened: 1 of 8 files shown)


lib/puppet/functions/foreman/to_symbolized_yaml.rb
```diff
@@ -0,0 +1,38 @@
...
+#   Convert a data structure and output it as YAML while symbolizing keys
+#
+# In Foreman often YAML files have symbols as keys. Since it's hard to do that
+# from Puppet, this function does it for you.
+#
+# @example How to output YAML
+#   # output yaml to a file
+#     file { '/tmp/my.yaml':
+#       ensure  => file,
+#       content => to_yaml($myhash),
```
ehelms, line 14:
```suggestion
#       content => foreman::to_symbolized_yaml($myhash),
```

lib/puppet/functions/foreman/to_symbolized_yaml.rb
```diff
@@ -0,0 +1,38 @@
...
+#     }
+# @example Use options control the output format
+#   file { '/tmp/my.yaml':
+#     ensure  => file,
+#     content => to_yaml($myhash, {indentation: 4})
```
ehelms, line 19:
```suggestion
#     content => foreman::to_symbolized_yaml($myhash, {indentation: 4})
```

ehelms:
Will this handle:

 * Nested hashes
 * Hashes inside arrays

lib/puppet/functions/foreman/to_symbolized_yaml.rb
ekohl: I think you caught my copy-paste

ekohl:
> * Nested hashes
> 
> * Hashes inside arrays

No. I chose not to implement that. Partly because I was lazy, partly because I don't know if there is any demand for it. It may even break in some cases.

ekohl:
Rebased to resolve conflicts.

spec/defines/foreman_dynflow_worker_spec.rb
```diff
@@ -42,21 +42,58 @@ class { 'foreman':
...
+          context 'and priorities' do
+            let(:params) do
+              {queues: [['default', 1], ['remote_execution', 1]]}
+            end
+
+            let(:expected_content) do
+              <<~CONTENT
+                ---
+                :concurrency: 1
+                :queues:
```
ehelms, line 79: `queues` can either be a list or a list of lists? Thats right ugly but I realize not our "concern" per this feature.

ehelms (approved):

spec/defines/foreman_dynflow_worker_spec.rb
ekohl: Yes, that is indeed my impression based on @adamruzicka's patch. I think we're just exposing what Dynflow accepts.

@adamruzicka can you confirm?

adamruzicka: It is a sidekiq-specific thing[1], Dynflow has no role in this. It can be either a list or a list of two element lists, where first element is name of the queue and the other is the priority.

[1] - https://github.com/mperham/sidekiq/wiki/Advanced-Options#queues

adamruzicka (approved):

PR #928, opened 2021-03-31
Fixes #29780 - Drop Passenger support and target Foreman 2.4+

ekohl:
With Foreman 2.5, support to run with Passenger (technically mod_passenger) is dropped in favor of using Puma with a reverse proxy.

This also moves the request headers part of settings.yaml to a concat fragment to simplify some code.

It keeps in code to remove tfm-rubygem-passenger-native on EL7 which it previously didn't do, but is nice to do. It doesn't remove mod_passenger itself since that's not managed in this module.

### DIFF
--- a/README.md
+++ b/README.md
@@ -74,10 +74,7 @@ previous stable release.
 
 ### Foreman version compatibility notes
 
-This module targets Foreman 2.0+.
-
-If you're running Foreman instance older than Foreman 2.2 you need to change logging layout parameter from
-`multiline_request_pattern` to `multiline_pattern` due to compatibility.
+This module targets Foreman 2.4+.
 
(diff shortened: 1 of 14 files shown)


ekohl:
Converting to draft since I want to test this on my own server.

One open question: should we clean up the tfm passenger native package or should that be `Obsoletes: ` in the `tfm` package? https://github.com/theforeman/foreman-packaging/pull/6494 doesn't do that now but should it?

ekohl:
Rebased, but no actual change. I've deployed this on my own Foreman 2.4 server and it works well there.

ehelms (approved):
Passed my pipeline testing and the code looks sane

PR #929, opened 2021-04-01
Only run apipie:cache:index for plugin with an API

ekohl:
This is the result of a discussion on IRC with @tbrisker and @ofedoren. Right now it's not clear if this is the right thing, but this captures the general thought.

The idea is that the apipie cache index only needs to be regenerated if a plugin with an API is installed. This saves time on installation.

It also turns out that the DB does not need to be present and will actually prevent warnings about this during db:migrate. Not triggering it after that also makes it much faster.

One thing where this will now introduce a regression is that it no longer (reliably) regenerates the cache on upgrades since no refresh events are triggered. The ideal solution is to stop relying on refreshes and instead have an unless statement that determines if it's really needed.

### DIFF
--- a/manifests/plugin/discovery.pp
+++ b/manifests/plugin/discovery.pp
@@ -4,5 +4,6 @@
 #
 class foreman::plugin::discovery {
   foreman::plugin {'discovery':
+    has_api => true,
   }
 }
--- a/manifests/rake/apipie.pp
+++ b/manifests/rake/apipie.pp
@@ -1,0 +1,9 @@
+# @summary a rake task to ensure a correct apipie cache
+# @api private
+class foreman::rake::apipie {
+  foreman::rake { 'apipie:cache:index':
+    timeout => 0,
+    require => Class['foreman::config'],
+    before  => Class['foreman::database'],
+  }
+}
--- a/manifests/database.pp
+++ b/manifests/database.pp
@@ -35,6 +35,6 @@ class foreman::database(
     ~> foreman::rake { 'db:seed':
       environment => delete_undef_values($seed_env),
     }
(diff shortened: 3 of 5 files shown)


manifests/plugin/discovery.pp
```diff
@@ -4,5 +4,6 @@
 #
 class foreman::plugin::discovery {
   foreman::plugin {'discovery':
+    has_api => true,
```
ekohl, line 7: We would need to go over all plugins and determine this. This is just an example. Packaging should tell us easily what the value is.

manifests/rake/apipie.pp
```diff
@@ -0,0 +1,9 @@
+# @summary a rake task to ensure a correct apipie cache
+# @api private
+class foreman::rake::apipie {
+  foreman::rake { 'apipie:cache:index':
+    timeout => 0,
```
ekohl, line 5: This would need an `unless` here to make it reliable.

ehelms:
To help me think about this, is this fixing a bug or just a performance improvement?

ekohl:
Performance improvement. Today we spend a lot of time regenerating the index and in many cases it's not needed. It's only really needed if files change on disk (which happens after a package install/update). @ofedoren also said it's only needed if a plugin has an API. I haven't verified that yet.

ekohl:
Experience now tells me this isn't really true. The DB must be present and if you don't run the index (even without plugins) it will continue to complain about a missing cache. I don't want to close this just yet, but there's a reason it's a draft.

ekohl:
Since this still needs some serious work in Foreman itself if we ever want this, I'm closing this.

PR #935, opened 2021-04-06
Fixes #29649 - Drop default_server argument in IPA

ekohl:
ipa-getkeytab can figure out the default server on its own[1]. There is no need to specify it and can even break things. For example, DNS can be used to detect servers. Then the fact is empty and it fails while the command would actually pass.

The fact is simplified since it's a major version bump anyway and nothing else should use our foreman_ipa fact.

[1] https://github.com/theforeman/puppet-foreman/pull/880#issuecomment-683902223

Replaces https://github.com/theforeman/puppet-foreman/pull/880

### DIFF
--- a/lib/facter/sssd.rb
+++ b/lib/facter/sssd.rb
@@ -1,19 +1,6 @@
 require 'facter/util/sssd'
 
 if defined? Facter::Util::Sssd
-  # == Fact: foreman_ipa
-  Facter.add(:foreman_ipa, :type => :aggregate) do
-    {
-      :default_realm => 'global/realm',
-      :default_server => 'global/server',
-    }.each do |key, path|
-      chunk(key) do
-        val = Facter::Util::Sssd.ipa_value(path)
-        {key => val} if val
-      end
-    end
-  end
-
   # == Fact: foreman_sssd
   Facter.add(:foreman_sssd, :type => :aggregate) do
     {
--- a/manifests/config.pp
+++ b/manifests/config.pp
@@ -126,10 +126,6 @@ class foreman::config {
     $foreman_socket_override = template('foreman/foreman.socket-overrides.erb')
 
     if $foreman::ipa_authentication {
-      unless fact('foreman_ipa.default_server') {
-        fail("${facts['networking']['hostname']}: The system does not seem to be IPA-enrolled")
-      }
-
       if $facts['os']['selinux']['enabled'] {
         selboolean { ['allow_httpd_mod_auth_pam', 'httpd_dbus_sssd']:
           persistent => true,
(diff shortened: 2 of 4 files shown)


ehelms (approved):

ehelms:
@ekohl Is it ok to merge this?

ekohl:
I've got most of my lab set up again on my new desktop, I'll see about verifying this with a real IPA sever.

manifests/config.pp
```diff
@@ -120,7 +120,7 @@
     $foreman_socket_override = template('foreman/foreman.socket-overrides.erb')
 
     if $foreman::ipa_authentication {
-      unless fact('foreman_ipa.default_server') and fact('foreman_ipa.default_realm') {
+      unless fact('foreman_ipa.default_realm') {
```
evgeni, line 123: given we don't need it anymore when we merge #967 -- what should we do here? ;-)

ekohl: Drop the whole `unless` block.

lib/facter/sssd.rb
```diff
@@ -1,19 +1,6 @@
 require 'facter/util/sssd'
 
 if defined? Facter::Util::Sssd
-  # == Fact: foreman_ipa
-  Facter.add(:foreman_ipa, :type => :aggregate) do
-    {
-      :default_realm => 'global/realm',
```
evgeni, line 7: I might end up needing the realm for Debian support 🙈

ekohl: With https://github.com/theforeman/foreman-documentation/pull/594 I think we can safely remove it? If so, I'll rebase this.

evgeni: Yes, that was my intention!

ekohl:
Rebased

manifests/config.pp
```diff
@@ -126,10 +126,6 @@
     $foreman_socket_override = template('foreman/foreman.socket-overrides.erb')
 
     if $foreman::ipa_authentication {
-      unless fact('foreman_ipa.default_server') {
-        fail("${facts['networking']['hostname']}: The system does not seem to be IPA-enrolled")
```
evgeni, line 130: I wonder what this guard originally was about?

Worst that can happen without it is that `ipa-getkeytab` below fails?

evgeni (approved):
Overall I am fine with the change.

The only thing that left me wondering is the "not enrolled system" guard that we're removing, but I also don't *really* see any harm in doing so?

It has been there since 378d602424da89f4493767949d534b66942db8e2 introduced IPA support in the installer

manifests/config.pp
ekohl: Previously it would guaranteed generate an incorrect command if that fact was missing. However, we can't really reliably detect it (DNS records can be used for example). So you're right that ipa-getkeytab could fail during runtime and exec resources don't log well in our installer (happens at the info level and to make the installer quieter, info is only sent to `/var/log`). However, I do think that fully supporting all possible registrations is more important than perfect error handling in this case.

PR #936, opened 2021-04-08
Refs #32276: Add Katello hammer plugin

ehelms:
Adding this here will allow adding to the foreman-installer similar to all other plugin hammer packages and then allows being able to drop these requirements from the katello meta RPM (https://github.com/theforeman/foreman-packaging/blob/rpm/develop/packages/katello/katello/katello.spec#L34-L36).

### DIFF
--- a/spec/acceptance/foreman_cli_plugins_spec.rb
+++ b/spec/acceptance/foreman_cli_plugins_spec.rb
@@ -3,41 +3,77 @@ require 'spec_helper_acceptance'
 describe 'Scenario: install foreman-cli + plugins without foreman' do
   before(:context) { purge_foreman }
 
-  let(:pp) do
-    <<-PUPPET
-    class { 'foreman::cli':
-      foreman_url => 'https://foreman.example.com',
-      username    => 'admin',
-      password    => 'changeme',
-    }
+  package_prefix = fact('os.release.major') == '7' ? "tfm-" : ""
 
-    if $facts['os']['family'] == 'RedHat' {
-      include foreman::cli::ansible
-      include foreman::cli::azure
-    }
-    include foreman::cli::discovery
-    include foreman::cli::remote_execution
-    include foreman::cli::tasks
-    include foreman::cli::templates
-    PUPPET
+  context 'for standard plugins' do
+    let(:pp) do
+      <<-PUPPET
+      class { 'foreman::cli':
+        foreman_url => 'https://foreman.example.com',
+        username    => 'admin',
+        password    => 'changeme',
+      }
+
+      if $facts['os']['family'] == 'RedHat' {
+        include foreman::cli::ansible
+        include foreman::cli::azure
+      }
+      include foreman::cli::discovery
+      include foreman::cli::remote_execution
+      include foreman::cli::tasks
+      include foreman::cli::templates
+      PUPPET
+    end
+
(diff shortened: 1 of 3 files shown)


spec/acceptance/foreman_cli_plugins_spec.rb
```diff
@@ -26,7 +26,7 @@ class { 'foreman::cli':
 
   it_behaves_like 'hammer'
 
-  ['discovery', 'remote_execution', 'tasks', 'templates'].each do |plugin|
+  ['discovery', 'remote_execution', 'tasks', 'templates', 'katello'].each do |plugin|
```
evgeni, line 29:
```
expected Package "tfm-rubygem-hammer_cli_foreman_katello" to be installed
```

you'll need to fixup the logic below to not add the `foreman_` prefix and add it to the plugins here, thanks katello.

ekohl: You're missing the actual `include foreman::cli::katello`. I would expect that it's not packaged for Debian so add it like ansible/azure. Right now we also don't have any explicit checks for those and assume that if Puppet worked correctly, it's good enough.

spec/classes/cli_plugins_spec.rb
```diff
@@ -2,13 +2,21 @@
...
         let(:pre_condition) { 'include foreman::cli' }
-        let(:plugin_name) { plugin == 'azure' ? 'foreman_azure_rm' : "foreman_#{plugin}" }
+        let(:plugin_name) do
+          if plugin == 'azure'
+            'foreman_azure_rm'
+          elsif plugin == 'katello'
+            'katello'
+          else
+            "foreman_#{plugin}"
+          end
```
ekohl, line 18: Feels like a case statement is more natural
```suggestion
          case plugin
          when 'azure'
            'foreman_azure_rm'
          when 'katello'
            'katello'
          else
            "foreman_#{plugin}"
          end
```

spec/acceptance/foreman_cli_plugins_spec.rb
```diff
@@ -3,41 +3,75 @@
...
-                   when 'Debian'
-                     "ruby-hammer-cli-foreman-#{plugin.tr('_', '-')}"
-                   else
-                     plugin
-                   end
+      it_behaves_like 'hammer'
 
-    describe package(package_name) do
-      it { is_expected.to be_installed }
+      package_name = fact('os.release.major') == '7' ? "tfm-rubygem-hammer_cli_katello" : "rubygem-hammer_cli_katello"
```
ekohl, line 71:
```suggestion
      package_name = (fact('os.release.major') == '7' ? "tfm-" : "") + "rubygem-hammer_cli_katello"
```

Or

```suggestion
      package_name = "#{fact('os.release.major') == '7' ? "tfm-" : ""}rubygem-hammer_cli_katello"
```

Or, since it's already used elsewhere, define a regular variable that can also be used in the regular plugin test:
```ruby
scl_prefix = fact('os.release.major') == '7' ? "tfm-" : ""
```

ekohl (commented):
Small coding style suggestions, but otherwise :+1:

ehelms:
Some of the failures here are related to https://github.com/theforeman/puppet-foreman/pull/937

ekohl (approved):

PR #937, opened 2021-04-09
Enable Ruby 2.7 module for EL8 on Foreman 2.5+

ehelms:


ekohl (changes requested):
I don't think this is right. This feels like something that the repo class should do. https://github.com/theforeman/puppet-foreman/blob/master/manifests/repos/yum.pp to be exact.

manifests/repo.pp
```diff
@@ -31,6 +31,14 @@
     }
   }
 
+  if $facts['os']['release']['major'] == '8' {
```
ekohl, line 34: I think this should take the version into account. The question is what it should do if `$repo` is undef. Do we assume nightly or assume nothing. I'm leaning to the latter.

ekohl (approved):

PR #938, opened 2021-04-15
Refs #32885: Add puppet user to user_groups only if server or client certificate contains puppet path

ehelms:
Looking for feedback here as I don't know what the exact correct answer is to prevent the following error when setting `puppet::server` to false:

```
2021-04-15 20:08:58 [ERROR ] [configure] Could not set groups on user[foreman]: Execution of '/sbin/usermod -G puppet foreman' returned 6: usermod: group 'puppet' does not exist
2021-04-15 20:08:58 [ERROR ] [configure] /Stage[main]/Foreman::Config/User[foreman]/groups: change from  to 'puppet' failed: Could not set groups on user[foreman]: Execution of '/sbin/usermod -G puppet foreman' returned 6: usermod: group 'puppet' does not exist
```

It feels like a user of the module should not have to explicitly configure the `user_groups` if they are setting `puppet::server` to false and that the module should adjust appropriately.

The puppet-foreman_proxy seems to have a similar failure but different approach: https://github.com/theforeman/puppet-foreman_proxy/blob/941e934a4a27a40e7963bc59f6ab5c14fafb82dc/manifests/config.pp#L33

### DIFF
--- a/manifests/config.pp
+++ b/manifests/config.pp
@@ -74,6 +74,12 @@ class foreman::config {
   }
 
   if $foreman::manage_user {
+    if $foreman::puppet_ssldir in $foreman::server_ssl_key or $foreman::puppet_ssldir in $foreman::client_ssl_key {
+      $_user_groups = $foreman::user_groups + ['puppet']
+    } else {
+      $_user_groups = $foreman::user_groups
+    }
+
     group { $foreman::group:
       ensure => 'present',
     }
@@ -83,7 +89,7 @@ class foreman::config {
       comment => 'Foreman',
       home    => $foreman::app_root,
       gid     => $foreman::group,
-      groups  => $foreman::user_groups,
+      groups  => unique($_user_groups),
     }
   }
 
--- a/spec/acceptance/hieradata/common.yaml
+++ b/spec/acceptance/hieradata/common.yaml
@@ -5,4 +5,6 @@ foreman::server_ssl_cert: /etc/foreman-certs/certificate.pem
 foreman::server_ssl_chain: /etc/foreman-certs/certificate.pem
 foreman::server_ssl_crl: ""
 foreman::server_ssl_key: /etc/foreman-certs/key.pem
-foreman::user_groups: []
+foreman::client_ssl_ca: /etc/foreman-certs/certificate.pem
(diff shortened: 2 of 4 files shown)


manifests/config.pp
```diff
@@ -83,7 +89,7 @@
       comment => 'Foreman',
       home    => $foreman::app_root,
       gid     => $foreman::group,
-      groups  => $foreman::user_groups,
+      groups  => $_user_groups,
```
ekohl, line 92: I think this works, but there's a risk someone added `puppet` already and it ends up being added twice. We may also need an installer migration for this.
```suggestion
      groups  => unique($_user_groups),
```

spec/acceptance/hieradata/common.yaml
```diff
@@ -5,4 +5,6 @@ foreman::server_ssl_cert: /etc/foreman-certs/certificate.pem
 foreman::server_ssl_chain: /etc/foreman-certs/certificate.pem
 foreman::server_ssl_crl: ""
 foreman::server_ssl_key: /etc/foreman-certs/key.pem
-foreman::user_groups: []
+foreman::client_ssl_ca: /etc/foreman-certs/certificate.pem
+foreman::client_ssl_cert: /etc/foreman-certs/certificate.pem
+foreman::client_ssl_key: /etc/foreman-certs/key.pem
```
ekohl, line 10: For this I thought about making these defaults undef and then `pick($server, $client)` to determine the real value. That way, you don't need to specify them twice if they match.

ehelms: I think I could tackle that over here -- https://github.com/theforeman/puppet-foreman/pull/951

ekohl (commented):
I'd like to see some spec test for this. Also, I think the word "only" is missing in the title to indicate the change. Other than that :+1:

ehelms:
I added a spec test, let me know if you were also wanting an acceptance test.

ekohl (approved):
We should do an installer migration for this as well to remove `'puppet'` from the answers file to really benefit from this.

ehelms:
Here is the installer PR -- https://github.com/theforeman/foreman-installer/pull/690

PR #943, opened 2021-04-27
Customisable Yum repository base URL and GPG key path

### LINKED ISSUE #942: Customisable Yum repository base URL
We've got our own mirror of the Foreman Yum repositories so it'd be nice if the module allowed to set the base URL of the Yum repository which is currently hard-coded to `https://yum.theforeman.org/`:

https://github.com/theforeman/puppet-foreman/blob/master/manifests/repos/yum.pp#L18

As a bonus it'd be nice as well if the path to the GPG public key could be customised as well to accommodate different repository mirroring approaches.

We have a local patch with a proposed fix that we'd have to improve before submitting but I don't know if I'll have to time to work on in the near future so if somebody wants to provide one just go ahead :)

nbarrientos:
This patch adds two parameters:
1) `foreman::repo::yum_repo_baseurl` allowing to configure the prefix of the Yum repositories URL. This permits users having mirrors of `http://yum.theforeman.org` to use this module, for instance by setting the variable to `https://enterprise/internal/mirrors/yum.theforeman.org`.

The parameter's interface is pulled up to the public class and passed down to `foreman::repos::yum` via `foreman::repos`. The name might be ambiguous as the value is ultimately used as a component of a `Yumrepo`'s `baseurl`. Happy to give it a better name.

2) `foreman::repos::yum::keypath` allowing to completely customise the location of the GPG public key used to verify the signatures of the packages in the Yum repositories configured by the module. If not set the value before this patch is respected.

In this case I decided not to expose the parameter in the public interface to avoid cluttering the patch and also because perhaps this is something that few people will use. I'd be happy though to move it to `foreman::repo`, the same way as 1) if necessary.

Fixes #942

### DIFF
--- a/manifests/repos/yum.pp
+++ b/manifests/repos/yum.pp
@@ -5,7 +5,9 @@ define foreman::repos::yum (
   String $yumcode,
   Boolean $gpgcheck,
   Stdlib::HTTPUrl $baseurl,
+  Optional[String] $keypath = undef,
 ) {
+  $_keypath = pick($keypath, "${baseurl}/releases/${repo}/RPM-GPG-KEY-foreman")
   $gpgcheck_enabled_default = $gpgcheck ? {
     false   => '0',
     default => '1',
@@ -18,14 +20,14 @@ define foreman::repos::yum (
     descr    => "Foreman ${repo}",
     baseurl  => "${baseurl}/releases/${repo}/${yumcode}/\$basearch",
     gpgcheck => $gpgcheck_enabled,
-    gpgkey   => "${baseurl}/releases/${repo}/RPM-GPG-KEY-foreman",
+    gpgkey   => $_keypath,
     enabled  => '1',
   }
   yumrepo { "${name}-source":
     descr    => "Foreman ${repo} - source",
     baseurl  => "${baseurl}/releases/${repo}/${yumcode}/source",
     gpgcheck => $gpgcheck_enabled,
-    gpgkey   => "${baseurl}/releases/${repo}/RPM-GPG-KEY-foreman",
+    gpgkey   => $_keypath,
     enabled  => '0',
   }
   yumrepo { "${name}-plugins":
--- a/manifests/repos/yum.pp
+++ b/manifests/repos/yum.pp
@@ -4,6 +4,7 @@ define foreman::repos::yum (
   Variant[Enum['nightly'], Pattern['^\d+\.\d+$']] $repo,
   String $yumcode,
   Boolean $gpgcheck,
+  Stdlib::HTTPUrl $baseurl,
 ) {
   $gpgcheck_enabled_default = $gpgcheck ? {
     false   => '0',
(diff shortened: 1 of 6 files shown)


manifests/repos/yum.pp
```diff
@@ -4,7 +4,13 @@
   Variant[Enum['nightly'], Pattern['^\d+\.\d+$']] $repo,
   String $yumcode,
   Boolean $gpgcheck,
+  Stdlib::HTTPUrl $baseurl,
+  Optional[String] $keypath = undef,
 ) {
+  $_keypath = $keypath ? {
+    undef   => "${baseurl}/releases/${repo}/RPM-GPG-KEY-foreman",
+    default => $keypath,
+  }
```
ekohl, line 13: I always prefer [pick()](https://forge.puppet.com/modules/puppetlabs/stdlib/reference#pick)
```suggestion
  $_keypath = pick($keypath, "${baseurl}/releases/${repo}/RPM-GPG-KEY-foreman")
```

spec/classes/foreman_repo_spec.rb
```diff
@@ -46,7 +50,8 @@
         let :params do
           {
             repo: '1.19',
-            configure_scl_repo: false
+            configure_scl_repo: false,
+            yum_repo_baseurl: 'https://baz.fi'
```
ekohl, line 54: I'm a big fan of using `example.com` for these cases. Also, trailing comma's are great
```suggestion
            yum_repo_baseurl: 'https://example.com',
```

ekohl (changes requested):
Small details inline but overall :+1:

spec/classes/foreman_repo_spec.rb
nbarrientos: Sure, I'll patch the other cases, too.

ekohl (approved):
Note to self: squash on merge.

spec/classes/foreman_repo_spec.rb
nbarrientos: It's fixed (sorry, my sub-conscience typed `example.org` instead of `example.com`; both are valid IANA-approved example domains, though). 

In one case I'm using `examplekeys.org` to verify that the code is actually taking the `keypath` literally and not mixing it with `baseurl`.

nbarrientos:
I'd be happy to reorganise commits or even squash everything, just ping me. There are two features being added so keeping the patch down to two commits makes sense to me. Let me know :) Thanks.

ekohl:
I'd prefer things squashed. If they're 2 features, I'm happy with 2 commits but 4 is a bit too much.

spec/classes/foreman_repo_spec.rb
ekohl: My point was indeed that an IANA example domain should be used.

nbarrientos:
In two beautiful commits now. Thanks for your time!

ekohl (approved):
Waiting for CI to complete, otherwise :+1:

PR #944, opened 2021-04-29
Allow customising the list of HTTP headers to unset

### LINKED ISSUE #941: Allow customising request header filtering in vhost internal options
https://github.com/theforeman/puppet-foreman/commit/6f5b99bb08f19ff61410dc6208346b9c8c333221 wipes out several HTTP headers as a protective measure. If I understand correctly, this change was introduced after Foreman was patched to consume the `REMOTE_USER` HTTP *header* instead of the environment variable due to selecting Puma as preferred deployment application server. References:

* https://projects.theforeman.org/issues/30535
* https://github.com/theforeman/puppet-foreman/pull/872

While this is a sensible default, it prevents users from delegating the authentication to a trusted front-end. There might be other layers in front of the Apache web server configured by this module so it'd be nice if this header filter could be configurable via a class parameter. In our particular case, we need this because:

1) As mentioned,  the `REMOTE_USER` header is set by a third-party frontend to be consumed directly by Foreman so it cannot be wiped out by the Apache instance that sits in the middle (configured by this module).
2) We need to be able to unset other headers ourselves, like `OIDC_ACCESS_TOKEN`. This is because our authentication front-end already deals with OpenID connect authentication and sets `REMOTE_USER` itself. We don't want Foreman to see `OIDC_ACCESS_TOKEN` because if it does, as it seems not to be possible to disable OIDC, Foreman will try to use it and fail as in our case OIDC is not (and won't be) configured in the backend. Unfortunately, `REMOTE_USER` is totally ignored in the described scenario and there's no fall back.

We have a local patch with a proposed fix that we'd have to improve before submitting but I don't know if I'll have to time to work on in the near future so if somebody wants to provide one just go ahead :)

nbarrientos:
This patchset adds a new parameter to `foreman::config::apache` called `request_headers_to_unset` containing a list of HTTP headers to be unset before proxy passing the request to the application. See the
associated issue for more information.

Please let me know if you'd rather to have the parameter exposed via `init.pp` instead.

Fixes #941

### DIFF
--- a/manifests/config/apache.pp
+++ b/manifests/config/apache.pp
@@ -82,6 +82,9 @@
 # @param keycloak_realm
 #   The realm as passed to keycloak-httpd-client-install
 #
+# @param request_headers_to_unset A list of HTTP headers coming from
+#   the client that will be unset and hence not passed to the
+#   application.
 class foreman::config::apache(
   Stdlib::Absolutepath $app_root = '/usr/share/foreman',
   String $priority = '05',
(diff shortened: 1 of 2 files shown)


ekohl (approved):
> Please let me know if you'd rather to have the parameter exposed via `init.pp` instead.

I actually prefer it here. This can be considered advanced and only allowing that setting that via Hiera is good. This also makes it easy to set additional headers if we need to in the future. :+1: 

I've just started the tests and they should pass before merging. IMHO only waiting on that.

PR #945, opened 2021-04-29
Make database.yml and settings.yaml have consistent headers

gcoxmoz:
This brings the two files in sync with how they include the
_header.erb subtemplate.

Now, you might look at this and think, "what a silly commit."  Possibly, but, I've been having this issue where `database.yml` will have two _header.erb instances if `puppet agent` runs things, and only one _header.erb if I do a `puppet apply`.  `settings.yaml` doesn't do this, it always has one copy of the header.  Of course, if `database.yml` changes, that triggers all sorts of rakes and refreshes, which makes it so I can't tell if the prospective changes I'm testing with 'agent' are noisy or fine.

I then added the triple dash to open the yaml file, since it wasn't there on database.yml.

And I also added a `-` to remove a newline in `settings.yaml`, because otherwise it has a double newline after its header, and if I'm in here talking about consistency I may as well do that too.

Thanks for looking.

### DIFF
--- a/templates/database.yml.erb
+++ b/templates/database.yml.erb
@@ -1,4 +1,7 @@
-<%= ERB.new(File.read(File.expand_path("_header.erb",File.dirname(file)))).result(binding) -%>
+<% @header = ''
+   ERB.new(File.read(File.expand_path("_header.erb",File.dirname(file))), nil, nil, "@header").result(binding) -%>
+---
+<%= @header -%>
 
 <%
   host     = scope.lookupvar("::foreman::db_host")
--- a/templates/settings.yaml.erb
+++ b/templates/settings.yaml.erb
@@ -1,7 +1,7 @@
 <% @header = ''
    ERB.new(File.read(File.expand_path("_header.erb",File.dirname(file))), nil, nil, "@header").result(binding) -%>
 ---
-<%= @header %>
+<%= @header -%>
 
 :unattended: <%= scope.lookupvar("foreman::unattended") %>
 <% unless [nil, :undefined, :undef].include?(scope.lookupvar("foreman::unattended_url")) -%>


wbclark (commented):
Thanks @gcoxmoz - I appreciate you looking into this inconsistency and I definitely don't consider it a silly commit

Checking the `git blame` on these files and the commit where `templates/_header.erb` was first introduced, I noticed that these are much older commits (8-9 years), and furthermore that `database.yml` and `settings.yml` were authored by different people and pushed directly without a PR, meaning there is no discussion to reference today about what was the intention back then. I think the most likely explanation is simply different authors / different styles / simply an oversight.

With https://github.com/theforeman/puppet-foreman/commit/eae698b9069d24f7a5857b22ac70ebd7b5c48eec we can see at least a few different patterns depending on the file extension (`---` for `.yml`, `#!/usr/bin/env ruby` for `.rb`, and so on) which was useful to understand the intention around headers as they were introduced.

Checking over the other yaml templates present today, some like `templates/foreman_memcache.yaml.erb` are missing headers entirely. I think it's always nice to let users know where these files are coming from. Since we are are already talking about 'consistency' here, would you be willing to take a look at ensuring consistent headers across other templated yaml files as well? :)

Let me know either way. Thanks!

wbclark:
> I'm somewhat concerned that I don't how this has escaped folks for 8+ years, and that I don't know how to write a test that would prove it's doing the wrong thing given how it's weird how the issue doesn't show up in puppet apply.

My best guess is that while your workflow is a totally normal and valid way to use this (or any) puppet module, it wouldn't likely be found by Foreman users who are not installing with Puppet directly but using Puppet to install only indirectly via foreman-installer. I don't have any specific numbers on this, but since that's the installation method suggested by the documentation, I assume it is what most people would use.

> I also am concerned that, if I blow this PR out into a full-consistency-effort, that the change I feel is needed (this one) could drown in nitpicking on things that are nice but I don't care about.

That's a totally fine and valid concern also. Thanks for bringing this to our attention and we can create an issue to check over headers for other templates later.
 
> That said, I can certainly go OCD and apply this same level of change to the rest of the yamls with the same change. I have opened #946 with those changes. (in draft until tests complete and I see if they're green)

Or if you want to go above and beyond in trying it yourself, that's fine too :)

I'll prioritize taking a look at the other PR, and in case I feel like there is risk of getting bogged down and taking more of your time, we can merge this one for the value it adds now and I can happily pick up remaining work on the other if needed.

Thanks very much for your efforts,

wbclark (approved):
Although there is more that can be done to improve in this area, this PR represents an immediate improvement that improves quality / adds value now. For that reason I will ACK it.

Thanks @gcoxmoz !

PR #946, opened 2021-04-30
Add consistent headers to YAML templates

gcoxmoz:
Per a request in #945, update existing yaml template files to add headers indicating their origin.


gcoxmoz:
OK, since I'm kinda flaily in what the tests should look like (they themselves being inconsistent and probably better handled as structural matches than string matches), I yield to better coders than myself.  Thanks.

(credit doesn't matter to me, but thank you nonetheless)

wbclark:
> I yield to better coders than myself. Thanks.

I'm not so sure about the accuracy of that part, but the rest at least sounds good :)

Cheers,

gcoxmoz:
Out of date and not seeming relevant now.

PR #951, opened 2021-05-11
POC: Require user to define server and client certificates

ehelms:
The title here is perhaps a bit misleading, the goal of this draft POC is to allow design discussion about how we might design the following question:

How do we remove Puppet as the default certificates in the module but allow users to easily define the use of the Puppet certificates for use with puppet-foreman?

Today these paths are hard-coded, but also (as I understand it) exist in a more authoritative location within `puppet::server`: https://github.com/theforeman/puppet-puppet/blob/master/manifests/server.pp#L453-L463  Thus we ought to be able to provide a way (e.g wrapper/profile module? heira?) to use those paths as inputs to these Foreman parameters without hard coding.

This is a similar problem for how Katello's certificate setup currently works. Today the paths Katello uses for certificates to Foreman are hard coded in the installer's answer file (https://github.com/theforeman/foreman-installer/blob/develop/config/katello-answers.yaml#L14-L28) but are all available in variables. For example, https://github.com/theforeman/puppet-certs/blob/master/manifests/apache.pp#L24-L26


ehelms:
@ekohl I recall you once had some thoughts on how we could both support enforcing a user define these and perform some slightly automatic detection for the standard Puppet certificates use case

PR #953, opened 2021-05-12
Allow customising ProxyAddHeaders

### LINKED ISSUE #952: Allow customising ProxyAddHeaders
When the Apache instance configured by this module is not dealing with user requests directly because there's another proxy in front, it's undesirable that the "local" Apache resets headers like `X-Forwarded-Host`. When it does, it could happen that links created using [Rails `_url` helpers](https://github.com/rails/rails/blob/ef97441036e0ebbe1aa2108d59c408707f998ffd/actionpack/lib/action_dispatch/http/url.rb#L210-L226) do not contain the `Host` initially sent by the client leading to broken links (as they point to backend nodes not reachable by the user).

The task is to make [this](https://github.com/theforeman/puppet-foreman/blob/master/manifests/config/apache.pp#L177) configurable by adding a parameter to `foreman::config::apache`.

nbarrientos:
When the Apache instance configured by this module is not dealing with user requests directly because there's another proxy in front, it's undesirable that the "local" Apache resets headers like X-Forwarded-Host. When it does, it could happen that links created using Rails _url helpers do not contain the Host initially sent by the client leading to broken links (as they point to backend nodes not reachable by the user).

This patch makes this setting customisable by the administrator by exposing it a parameter.

Fixes #952

### DIFF
--- a/manifests/config/apache.pp
+++ b/manifests/config/apache.pp
@@ -48,6 +48,10 @@
 # @param user
 #   The user under which the application runs.
 #
+# @param proxy_add_headers
+#   Do not set X-Forwarded-* headers. Useful when having another
+#   balancing layer in front of this Apache instance.
+#
 # @param proxy_backend
 #   The backend service to proxy to
 #
@@ -93,6 +97,7 @@ class foreman::config::apache(
   Stdlib::Port $server_port = 80,
   Stdlib::Port $server_ssl_port = 443,
   Pattern['^(https?|unix)://'] $proxy_backend = 'unix:///run/foreman.sock',
+  Boolean $proxy_add_headers = true,
   Hash $proxy_params = {'retry' => '0'},
   Array[String] $proxy_no_proxy_uris = ['/pulp', '/pulp2', '/streamer', '/pub', '/icons'],
   Boolean $ssl = false,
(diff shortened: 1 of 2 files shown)


ekohl (approved):
For what it's worth, you don't need to open an issue for a PR. A PR with a good description is sufficient.

ekohl:
I was reading more documentation and wanted to capture it. There's [ProxyPreserveHost](https://httpd.apache.org/docs/2.4/mod/mod_proxy.html#proxypreservehost):
> When enabled, this option will pass the Host: line from the incoming request to the proxied host, instead of the hostname specified in the ProxyPass line.
>
> This option should normally be turned Off. It is mostly useful in special configurations like proxied mass name-based virtual hosting, where the original Host header needs to be evaluated by the backend server.

Perhaps this is what you're looking for?

Another is [mod_remoteip](https://httpd.apache.org/docs/2.4/mod/mod_remoteip.html) which may also be useful.

nbarrientos:
Thanks! Not (re-)setting any `X-Forwarded-`* header at all (by setting `ProxyAddHeaders` to `Off`) on the _second_ Apache is just enough for us, this way Foreman also sees for instance `X-Forwarded-For` which is I believe consumed by the audit logs.

PR #958, opened 2021-06-01
Drop Puppet 5 support

ehelms:


ekohl (changes requested):
This bit was only needed until SERVER-2597 was fixed. That was part of Puppetserver 6.10.0 which we can pretty much assume if 6.15.0 is the minimal version. In theory the server could be older, but I don't think we need to support that.

https://github.com/theforeman/puppet-foreman/blob/6c902a4f1f484137ebccafe4d3390881fd0b3d61/lib/puppet/functions/foreman/enc.rb#L48-L54

It can be dropped since the basic auth option is passed correctly in the request.

ekohl (approved):

PR #960, opened 2021-06-21
Accept Datatype Sensitive for all Secrets

cocker-cc:


manifests/cli.pp
```diff
@@ -28,16 +28,16 @@
...
-  String $version = $foreman::cli::params::version,
-  Boolean $manage_root_config = $foreman::cli::params::manage_root_config,
-  Optional[String] $username = $foreman::cli::params::username,
-  Optional[String] $password = $foreman::cli::params::password,
-  Boolean $use_sessions = $foreman::cli::params::use_sessions,
-  Boolean $refresh_cache = $foreman::cli::params::refresh_cache,
-  Integer[-1] $request_timeout = $foreman::cli::params::request_timeout,
-  Optional[Stdlib::Absolutepath] $ssl_ca_file = $foreman::cli::params::ssl_ca_file,
-  String $hammer_plugin_prefix = $foreman::cli::params::hammer_plugin_prefix,
+  Optional[Stdlib::HTTPUrl]                    $foreman_url          = $foreman::cli::params::foreman_url,
```
ekohl, line 31: I'd prefer not to use alignment. It really creates more problems than it solves.

ekohl (changes requested):
I haven't really reviewed it in depth. Please don't align parameters.

Other than that I am open to it. We'll need to verify if our installer handles it properly and otherwise patch that.

I do feel like a Puppet function to unwrap if needed makes sense.

manifests/cli.pp
cocker-cc: I un-aligned the Parameters again.

manifests/init.pp
```diff
@@ -199,99 +199,119 @@
...
-  Boolean $hsts_enabled = $foreman::params::hsts_enabled,
-  Array[Stdlib::HTTPUrl] $cors_domains = $foreman::params::cors_domains,
-  Integer[0] $foreman_service_puma_threads_min = $foreman::params::foreman_service_puma_threads_min,
-  Integer[0] $foreman_service_puma_threads_max = $foreman::params::foreman_service_puma_threads_max,
-  Integer[0] $foreman_service_puma_workers = $foreman::params::foreman_service_puma_workers,
-  Hash[String, Any] $rails_cache_store = $foreman::params::rails_cache_store,
-  Boolean $keycloak = $foreman::params::keycloak,
-  String[1] $keycloak_app_name = $foreman::params::keycloak_app_name,
-  String[1] $keycloak_realm = $foreman::params::keycloak_realm,
+  Stdlib::HTTPUrl                                                           $foreman_url                      = $foreman::params::foreman_url,
```
ekohl, line 202: Could you un-align it in all files again?

manifests/database/postgresql.pp
```diff
@@ -7,7 +7,9 @@
 
   $password = $foreman::db_password ? {
     'UNSET' => false,
-    default => postgresql::postgresql_password($foreman::db_username, $foreman::db_password),
+    # as soon, as the Function accepts Sensitive (Pull-Request pending), we can remove the Workaround
```
ekohl, line 10: https://github.com/puppetlabs/puppetlabs-postgresql/pull/1279

manifests/init.pp
cocker-cc: Took me some Time with `git rebase -i` while keeping the Commits separate.  But now I unaligned all Files

manifests/database.pp
```diff
@@ -14,7 +14,7 @@
   if $foreman::db_manage_rake {
     $seed_env = {
       'SEED_ADMIN_USER'       => $foreman::initial_admin_username,
-      'SEED_ADMIN_PASSWORD'   => $foreman::initial_admin_password,
+      'SEED_ADMIN_PASSWORD'   => if $foreman::initial_admin_password =~ Sensitive {$foreman::initial_admin_password.unwrap} else {$foreman::initial_admin_password},
```
ekohl, line 17: How about a private `foreman::unwrap_sensitive` function?
```puppet
# @api private
function foreman::unwrap_sensitive ($value) {
  if $value =~ Sensitive {
    $value.unwrap
  } else {
    $value
  }
}
```

manifests/init.pp
```diff
@@ -292,6 +292,26 @@
   String[1] $keycloak_app_name = $foreman::params::keycloak_app_name,
   String[1] $keycloak_realm = $foreman::params::keycloak_realm,
 ) inherits foreman::params {
+
+  # this Workaround is necessary, because
+  # 1. the ERB-Template cannot handle Sensitive Data (whereas EPP would do so natively), and
+  # 2. the Function postgresql::postgresql_password is not ready for Sensitive yet (Pull-Request pending)
+  $db_password_unsensitive = if $db_password =~ Sensitive {
```
ekohl, line 299: This could use the same function to unwrap

spec/classes/foreman_spec.rb
```diff
@@ -252,6 +252,95 @@
         end
       end
 
+      describe 'with all parameters and Sensitive for Secrets' do
+        let :params do
+          {
+            foreman_url: 'http://localhost',
```
ekohl, line 258: You don't need to pass in all values. Just passing in those you care about should be sufficient. Then it'll use the defaults from params for the other ones.

spec/classes/plugin/supervisory_authority_spec.rb
```diff
@@ -1,12 +1,24 @@
...
+  context 'with Standard-Parameters' do
+    let(:params) do
+      {
+        'server_url'   => 'https://example.com',
+        'secret_token' => 'secret_example',
+        'service_name' => 'foreman prod',
+      }
+    end
+    include_examples 'basic foreman plugin tests', 'supervisory_authority'
+  end
```
ekohl, line 13: Not: I like separating contexts visually
```suggestion
  end

```

templates/foreman_supervisory_authority.yaml.epp
```diff
@@ -0,0 +1,25 @@
+<%- |
```
ekohl, line 1: Just thinking out loud: should we instead use a `to_yaml` method for this? We have https://github.com/theforeman/puppet-foreman/blob/master/lib/puppet/functions/foreman/to_symbolized_yaml.rb and it looks like that will give the same result.

templates/hammer_etc.yml.epp
```diff
@@ -0,0 +1,30 @@
...
+  # and asks for them interactively at the begining of each session.
+  :use_sessions: <%= $epp_use_sessions %>
+
+  # Check API documentation cache status on each request
+  :refresh_cache: <%= $epp_refresh_cache %>
+
+  # API request timeout in seconds, set -1 for infinity
+  :request_timeout: <%= $epp_request_timeout %>
+
+<% if $epp_ssl_ca_file { -%>
```
ekohl, line 26:
```suggestion
<% if $epp_ssl_ca_file { -%>
```

templates/hammer_root.yml.epp
```diff
@@ -0,0 +1,8 @@
+<%- |
+  Optional[String]                             $epp_username,
+  Optional[Variant[String, Sensitive[String]]] $epp_password,
+| -%>
+:foreman:
+  # Credentials. You'll be asked for the interactively if you leave them blank here
+  :username: '<%= if $epp_username { $epp_username } else { 'undef'} %>'
+  :password: '<%= if $epp_password { $epp_password } else { 'undef'} %>'
```
ekohl, line 8: I'm not sure `undef` is accepted as a valid value.

manifests/plugin/supervisory_authority.pp
```diff
@@ -29,19 +29,34 @@
...
+  Integer[0] $api_buffer_size = 256,
+  String $api_request_size = '750kb',
+  String $api_request_time = '10s',
+  Integer[0] $transaction_max_spans = 500,
+  Boolean $http_compression = false,
+  String $metrics_interval = '30s',
 ) {
   foreman::plugin { 'supervisory_authority':
-    config => template('foreman/foreman_supervisory_authority.yaml.erb'),
+    config => epp(
```
ekohl, line 45: Given there's a secret in here, should this itself be marked as Sensitive?
```suggestion
    config => Sensitive(epp(
```

manifests/plugin.pp
```diff
@@ -27,7 +27,7 @@
   String[1] $config_file_owner = 'root',
   String[1] $config_file_group = $foreman::group,
   Stdlib::Filemode $config_file_mode = '0640',
-  Optional[String] $config = undef,
+  Optional[Variant[String, Sensitive[String]]] $config = undef,
```
ekohl, line 30: Should the `config_file` block below be changed to add `show_diff => !($config =~ Sensitive)` so it doesn't show up in logs anymore?

ekohl (commented):
I'm inclined to prefer this in separate PRs. The reason is that I don't know if our installer handles Sensitive well. Another is that we have some tools that read the sensitive values from our answers. For example, they read a DB password. I'm not saying this is good, but it's a concern that we need to address. Consider it a heads up that this is not a trivial merge for us.

So I'd say that the first one is converting templates from ERB to EPP. This is IMHO an easy PR to merge since it remains compatible.

Then dealing with Sensitive. Perhaps a way to mitigate our installer concerns is to first make sure the module handles Sensitive everywhere but keep defaulting to regular strings. This would allow users of the module to pass in sensitive values while not breaking our installer.

Then a third could be defaulting to Sensitive values.

I also have no experience with Sensitive myself. Whenever I tried it, I found it be quite useless given that you still need to unwrap in many places. I also don't know how well it works with Hiera.

As a last one - should we also change files with sensitive values to `show_diff => false`?

manifests/cli.pp
```diff
@@ -60,7 +60,16 @@
     owner   => 'root',
     group   => 'root',
     mode    => '0644',
-    content => template('foreman/hammer_etc.yml.erb'),
+    content => epp(
+      'foreman/hammer_etc.yml.epp',
+      {
+        epp_host            => $foreman_url_real,
```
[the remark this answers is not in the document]
cocker-cc: `epp_*` clearly expresses, that these Variables are solely meant for this EPP.  Using Variables like `epp_*` the next Developer is less tempted to just throw in plain Variables from out of the running Puppetcode.  Using Variables with `epp_*` there is a clear Border-Interface between running Puppetcode and the Template.  
Especially, when it comes to rename a Variable in the running Puppetcode, this comes handy, because you do not accidently rename the Templatevariable.

manifests/database.pp
cocker-cc: Not necessary anymore with [this Pullrequest](https://github.com/puppetlabs/puppet/pull/8651/files).

manifests/init.pp
cocker-cc: Not necessary anymore with [this Pullrequest](https://github.com/puppetlabs/puppet/pull/8651/files).

templates/hammer_root.yml.epp
cocker-cc: Of course this is not `undef`, but the String `'undef'`.  And of course with this String the Application will not work.  In the [originating Template](https://github.com/theforeman/puppet-foreman/blob/master/templates/hammer_root.yml.erb) there was no Handling of `undef` at all.   Do you have a better Solution?

manifests/plugin/supervisory_authority.pp
cocker-cc: You are right.

manifests/plugin.pp
cocker-cc: `file`-Resource does not show sensitive Content.  No Need for `show_diff => false`.

cocker-cc:
> I'm inclined to prefer this in separate PRs. 
> So I'd say that the first one is converting templates from ERB to EPP. 
> Then dealing with Sensitive. 
> Then a third could be defaulting to Sensitive values.

That sounds like a Plan.  So, we close this PR now, and I'll create separate PRs then.  Okay?

> I also have no experience with Sensitive myself. Whenever I tried it, I found it be quite useless given that **you still need to unwrap in many places**. 

Exactly **this** is the main Driver for my Effort.  Code with or without `Sensitive` should "just work".  The `unwrap` should happen at the least possible Location – or not at all, when it comes to EPP-Files.

> I also don't know how well it works with Hiera.

Not sure, if I understand You right.  With `Variant[String, Sensitive[String]]` the Code will (and should!) run with both Datatypes.  But if one insists, that Secrets coming from Hiera have to be of Type `Sensitive`, you can have something like this in your Module's `data/common.yaml`:
```
lookup_options:
  # Hiera delivers String.  To match the Classparameter's Datatype, we have to convert to Sensitive[String]
  '^my_cool_module::.+::secret_\w+$':
    convert_to: Sensitive
``` 

In my Environment the Secrets come from Hashicorp Vault using a Function like `vault_secrets::lookup`, which returns `Sensitive` natively.  My Goal is, to pass these Secrets down to the Component-Modules without any `unwrap`.  So I create Pullrequrests for all these Modules in Question, and `foreman` is one of them.

manifests/cli.pp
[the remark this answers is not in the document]
alexjfisher: epp already protects you against this?
https://puppet.com/docs/puppet/7/lang_template_epp.html#epp_variables

> Unlike Embedded Ruby (ERB) templates, EPP templates cannot directly access variables in the calling class without namespacing.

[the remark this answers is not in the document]
cocker-cc: As explained, I see this as a good Practice.  Do you want me to remove it anyways?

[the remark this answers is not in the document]
ekohl: I think it's redundant and makes thing needlessly verbose. As @alexjfisher said, I consider EPP safe enough (especially if you use type validations inside templates).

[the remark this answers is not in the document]
cocker-cc: `epp_` removed

manifests/database.pp
ekohl: I'm leaning to see when the Puppet core PR will be merged since there appears to be agreement on the general design already. Perhaps we can wait for a Puppet version with improved unwrap and set that as the minimum version? It'll keep the code cleaner.

cocker-cc: [First one](https://github.com/theforeman/puppet-foreman/pull/962) of the split-off and separated Pullrequests.

ekohl:
>  So I'd say that the first one is converting templates from ERB to EPP.

This part is now merged. Do you want to rebase this PR or open a new one?

cocker-cc:
I will rebase.

cocker-cc:
Rebasing done.

ekohl:
Looks like something went wrong in the rebase because I see some unrelated commits that are already in master. Also has more conflicts now.

cocker-cc:
> Looks like something went wrong in the rebase because I see some unrelated commits that are already in master. Also has more conflicts now.

I squashed and repaired my Branch.

lib/puppet/provider/foreman_config_entry/cli.rb
```diff
@@ -65,6 +65,8 @@ def value
 
   def value=(value)
     return if resource[:dry]
+
+    value = value.unwrap if value.respond_to?('unwrap')
```
ekohl, line 69: I always thought `respond_to` needed a symbol. However, the docs say:
> When the method name parameter is given as a string, the string is converted to a symbol.

That said, I'd still prefer to see a symbol to avoid the conversion:
```suggestion
    value = value.unwrap if value.respond_to?(:unwrap)
```

manifests/params.pp
```diff
@@ -37,7 +37,7 @@
   $db_username = 'foreman'
   # Generate and cache the password on the master once
   # In multi-puppetmaster setups, the user should specify their own
-  $db_password = extlib::cache_data('foreman_cache_data', 'db_password', extlib::random_password(32))
+  $db_password = Sensitive(extlib::cache_data('foreman_cache_data', 'db_password', extlib::random_password(32)))
```
ekohl, line 40: I am worried that our installer doesn't support Sensitive yet. That would be a blocker to merging this. We are entering our stablization period next week.

I feel like we can find a compromise if we don't default to using Sensitive for now but do merge the other changes to accept `Sensitive`. Do you agree or would you prefer to wait until after we've had a chance to check out installer?

templates/hammer_root.yml.epp
```diff
@@ -1,8 +1,8 @@
 <%- |
-  Optional[String] $username,
-  Optional[String] $password,
+  Optional[String]                             $username,
```
ekohl, line 2: I'd prefer not to align since it's a lot of work to maintain.
```suggestion
  Optional[String] $username,
```

manifests/database.pp
ekohl: > Not necessary anymore with [this Pullrequest](https://github.com/puppetlabs/puppet/pull/8651/files).

Just to close this discussion: this is available in Puppet 6.24.0 but I feel like requiring that version is a bit too early.  So let's stick with the long form for now.

templates/foreman_supervisory_authority.yaml.epp
ekohl: We merged that so you can remove this template again.

lib/puppet/provider/foreman_config_entry/cli.rb
cocker-cc: Changed.

templates/hammer_root.yml.epp
cocker-cc: Changed.

manifests/params.pp
cocker-cc: I removed the `Sensitive` here.

ekohl: To be clear, I was talking about all changes in params.pp, not just this single one.

templates/foreman_supervisory_authority.yaml.epp
cocker-cc: Done.

manifests/params.pp
cocker-cc: 👍  I removed all `Sensitive` from `params.pp` and left a `TODO`-Reminder.

manifests/cli.pp
```diff
@@ -93,7 +93,11 @@
         'foreman/hammer_root.yml.epp',
         {
           username => $username_real,
-          password => $password_real,
+          # epp() returns a Content of Datatype Sensitive, if at least one Parameter is of Datatype Sensitive.
+          # The "unwrap" therefore is not necessary, BUT: I do not know, how to make the Test pass in
+          # spec/classes/foreman_cli_spec.rb:110 otherwise.
```
ekohl, line 98: Perhaps https://github.com/theforeman/puppet-foreman/pull/976 would help with that.

manifests/params.pp
```diff
@@ -37,7 +37,11 @@
   $db_username = 'foreman'
   # Generate and cache the password on the master once
   # In multi-puppetmaster setups, the user should specify their own
+
+  # TODO: Use Sensitive, as soon, as the Installer supports it
+  #$db_password = Sensitive(extlib::cache_data('foreman_cache_data', 'db_password', extlib::random_password(32)))
```
ekohl, line 42: I'd prefer to avoid TODOs in the code for this and rather open an issue in the repo.

manifests/database/postgresql.pp
ekohl: The puppetlabs-postgresql PR was merged and released so this can be updated.

manifests/cli.pp
ekohl: Can this now be resolved?

ekohl (commented):
It took a while to circle back to this, but I still see this as a good enhancement and I'd like to see this finished.

manifests/database/postgresql.pp
cocker-cc: I removed the Workaround and amended the Commit.

manifests/params.pp
cocker-cc: I removed the TODOs and amended the Commit.

manifests/cli.pp
cocker-cc: I removed the Comment and the `unwrap`, but did not manage to establish a Rspec-Test.  
I amended the Commit.

ekohl (approved):
This looks good to me. Thanks for the big effort.

ekohl:
Sadly it looks like what we were afraid of is true: Kafo doesn't understand the Sensitive data type so it breaks: https://community.theforeman.org/t/foreman-installer-develop-package-release-1035-failed/26898

I'll see about a fix. For inspiration, do you know what's best practice with regards to Hiera?
I see https://puppet.com/docs/puppet/6/securing-sensitive-data.html#lookup-options but could that live in puppet-foreman?

cocker-cc:
> For inspiration, do you know what's best practice with regards to Hiera? I see https://puppet.com/docs/puppet/6/securing-sensitive-data.html#lookup-options but could that live in puppet-foreman?

Sure.  `lookup_options` is the Way to go.  
Hiera in general delivers a plain `String`, But with `lookup_options` Hiera is able to convert ithe `String` to `Sensitive[String]` before passing it over to Puppet.  
Where can I have a Look at the Hiera-File in Question?

ekohl:
Right now it's failing on parsing the data types for its own validation. We have a list of data types:
https://github.com/theforeman/kafo/blob/9ad08513b0b9ecbb31fea784bba2379ec74b5308/lib/kafo/data_type.rb#L114-L133
Perhaps right now we should implement it as a subclass of String. See:
* https://github.com/theforeman/kafo/tree/master/lib/kafo/data_types
* https://github.com/theforeman/kafo/blob/master/lib/kafo/data_types/string.rb

alexjfisher:
>Perhaps right now we should implement it as a subclass of String

Probably ok.  I've used `Sensitive[Hash]` as a data type before, but I think that's pretty rare and something you might never use in Foreman modules.

ekohl:
Technically I suppose we can implement it as a transparent container. I might have a try.

ekohl:
I took a stab at it: https://github.com/theforeman/kafo/pull/345. I haven't tested it end to end, but at least it recognizes the type. I also submitted https://github.com/theforeman/puppet-foreman/pull/1018 which I think should close the loop.

templates/hammer_root.yml.epp
```diff
@@ -1,8 +1,8 @@
...
   Optional[String] $username,
-  Optional[String] $password,
+  Optional[Variant[String, Sensitive[String]]] $password,
 | -%>
 :foreman:
   # Credentials. You'll be asked for the interactively if you leave them blank here
-  :username: '<%= $username %>'
-  :password: '<%= $password %>'
+  :username:<%= if $username { " '${username}'" } %>
+  :password:<%= if $password { " '${password}'" } %>
```
ekohl, line 8: Somehow this doesn't work for me. In #1018 I'm adding a test, but doesn't unwrap the string. I also checked if it perhaps was the interpolation, but that also doesn't appear to be it.

PR #962, opened 2021-06-24
Use EPP instead of ERB for some templates

cocker-cc:
As a Preparation for making this Module "Sensitive"-aware, we prefer to
use EPP instead of ERB in such Situations, where "Sensitive" could come
into Play, as EPP is able to handle Data of Type Sensitive natively.


templates/hammer_root.yml.epp
```diff
@@ -0,0 +1,8 @@
+<%- |
+  Optional[String] $username,
+  Optional[String] $password,
+| -%>
+:foreman:
+  # Credentials. You'll be asked for the interactively if you leave them blank here
+  :username: '<%= if $username { $username } else { 'undef'} %>'
+  :password: '<%= if $password { $password } else { 'undef'} %>'
```
ekohl, line 8: I'm not sure if `undef` is really intended. You're probably right that that was the previous behavior but it wasn't really intended. IIRC Puppet < 4 evaluated it to an empty string and we didn't have a test for it. I doubt hammer handles `undef`.

tl;dr: I think it should be this:

```suggestion
  :username: '<%=$username %>'
  :password: '<%= $password %>'
```

manifests/plugin/supervisory_authority.pp
```diff
@@ -42,6 +42,21 @@
...
+        log_level             => $log_level,
+        pool_size             => $pool_size,
+        api_buffer_size       => $api_buffer_size,
+        api_request_size      => $api_request_size,
+        api_request_time      => $api_request_time,
+        transaction_max_spans => $transaction_max_spans,
+        http_compression      => $http_compression,
+        metrics_interval      => $metrics_interval,
+      }
+    ),
```
ekohl, line 60: Untested, but I think it can actually not use any template and use this instead:
```suggestion
    config => foreman::to_symbolized_yaml(
      {
        server_url            => $server_url,
        secret_token          => $secret_token,
        service_name          => $service_name,
        log_level             => $log_level,
        pool_size             => $pool_size,
        api_buffer_size       => $api_buffer_size,
        api_request_size      => $api_request_size,
        api_request_time      => $api_request_time,
        transaction_max_spans => $transaction_max_spans,
        http_compression      => $http_compression,
        metrics_interval      => $metrics_interval,
      }
    ),
```

ekohl (commented):
Some inline suggestions for improvements. While this doesn't convert every file, I'm actually content with this as a start.

templates/hammer_root.yml.epp
cocker-cc: Suggestion implemented.

manifests/plugin/supervisory_authority.pp
cocker-cc: Maybe You can test it later, if the Function can handle `Sensitive[String] $secret_token`.

ekohl: You don't need to because you can unwrap when passing the hash.

cocker-cc: I thought, this was the whole Point of the Story: Not having to spread random `unwrap`s in the Code.  If the Function `foreman::to_symbolized_yaml` cannot handle `Sensitive`, then it has to be empowered to do so, and just like the Function `epp` it should return `Sensitive` then.

cocker-cc: Besides that, `foreman::to_symbolized_yaml` generates a symbolized(!) YAML like… 
```
---
:a: A
:b:
  x: X
  y: Y
:c:
- 1
- 2
- 3

```
…  (note the leading `:`)  I guess that's not, what the Configfile wants.

ekohl: Have a look at the template: it's exactly what it wants.

cocker-cc: Yes, You are right 👍   

[I upgraded Your Function](https://github.com/cocker-cc/puppet-foreman/compare/6c902a4f1f484137ebccafe4d3390881fd0b3d61..1fba024a8d226a0d7c77822c09bce363bc107f96) to be sensitive-aware.  
What do You think of a Pullrequest?

ekohl: For our use case right now I think it only complicates matters. I'd prefer to unwrap it in Puppet code.

ekohl:
I decided to push https://github.com/theforeman/puppet-foreman/pull/964 which is what I meant. I also left a comment where I think we should unwrap it.

manifests/plugin/supervisory_authority.pp
cocker-cc: If You want to, save my Code somewhere.  I will delete it, after this Story's End.

ekohl:
I've merged #964 so would you mind rebasing to resolve the conflict?

cocker-cc:
> I've merged #964 so would you mind rebasing to resolve the conflict?

Done.

ekohl (approved):
I'll let the tests complete, but otherwise :+1:

ekohl:
And those tests are green. Thanks!

PR #964, opened 2021-06-25
Use to_symbolized_yaml instead of a template for supervisory

ekohl:
This results in the same content on disk and removes the need for a template. This gives more guarantees that it's correct.


manifests/plugin/supervisory_authority.pp
```diff
@@ -41,7 +41,23 @@
   Boolean                      $http_compression      = false,
   String                       $metrics_interval      = '30s',
 ) {
+  $config = {
+    foreman_supervisory_authority => {
+      server_url            => $server_url,
+      secret_token          => $secret_token,
```
ekohl, line 47: Once we accept `Sensitive`, it should be unwrapped here.

ehelms (approved):
After fixing the test, can be merged

spec/classes/plugin/supervisory_authority_spec.rb
```diff
@@ -1,12 +1,13 @@
...
-    { 
+  let(:params) do
+    {
       'server_url'   => 'https://example.com',
       'secret_token' => 'secret_example',
       'service_name' => 'foreman prod',
     }
   end
   include_examples 'basic foreman plugin tests', 'supervisory_authority'
+  it { is_expected.to contain_foerman__plugin('supervisory_authority').with_config(%r{^---\n:foreman_supervisory_authority:\n  server_url: "https://example.com"\n}) }
```
ehelms, line 12:
```suggestion
  it { is_expected.to contain_foreman__plugin('supervisory_authority').with_config(%r{^---\n:foreman_supervisory_authority:\n  server_url: "https://example.com"\n}) }
```

ekohl: You know, I spotted this yesterday but didn't this morning ....

PR #965, opened 2021-06-30
Infrastructure roles v2

adamruzicka:


adamruzicka:
```
       Host 'centos8-64.example.com' exited with 6 running:
        puppet apply --verbose --detailed-exitcodes /tmp/apply_manifest.pp.rsvlRO
       Last 10 lines of output were:
       	Notice: /Stage[main]/Foreman::Service/Foreman::Dynflow::Pool[worker]/Foreman::Dynflow::Worker[worker-2]/File[/etc/foreman/dynflow/worker-2.yml]/ensure: defined content as '{sha256}4755afd02e227c21bd5b6bf2a98038ba42d6ae942c3ed7d21123a315cf5444d7'
       	Info: /Stage[main]/Foreman::Service/Foreman::Dynflow::Pool[worker]/Foreman::Dynflow::Worker[worker-2]/File[/etc/foreman/dynflow/worker-2.yml]: Scheduling refresh of Service[dynflow-sidekiq@worker-2]
       	Notice: /Stage[main]/Foreman::Service/Foreman::Dynflow::Pool[worker]/Foreman::Dynflow::Worker[worker-2]/Service[dynflow-sidekiq@worker-2]/ensure: ensure changed 'stopped' to 'running'
       	Info: /Stage[main]/Foreman::Service/Foreman::Dynflow::Pool[worker]/Foreman::Dynflow::Worker[worker-2]/Service[dynflow-sidekiq@worker-2]: Unscheduling refresh on Service[dynflow-sidekiq@worker-2]
       	Notice: /Stage[main]/Foreman::Service/Foreman::Dynflow::Pool[worker]/Foreman::Dynflow::Worker[worker]/Service[dynflow-sidekiq@worker]: Triggered 'refresh' from 1 event
       	Notice: /Stage[main]/Foreman::Service/Foreman::Dynflow::Pool[worker]/Foreman::Dynflow::Worker[worker]/File[/etc/foreman/dynflow/worker.yml]/ensure: removed
       	Info: Stage[main]: Unscheduling all events on Stage[main]
       	Error: Could not find a suitable provider for foreman_host
       	Error: Could not find a suitable provider for foreman_instance_host
```

Could anyone point me in the right direction?

ekohl (changes requested):
You need to include [foreman::providers](https://github.com/theforeman/puppet-foreman/blob/master/manifests/providers.pp) and ensure that they're installed before you call the new types. For the Smart Proxy it happens here:

https://github.com/theforeman/puppet-foreman_proxy/blob/2d47004f5dca8509bb30c90dc303893d27fdf0c3/manifests/install.pp#L25-L30

Note that this isn't really the best way  and we should really fix this with some autorequire in the type, but this has worked stable for a long time so we didn't bother too much with it.

You must also ensure that the service is up before using the types. For the Smart Proxy it happens here:

https://github.com/theforeman/puppet-foreman/blob/e16eaa3e25a77b6685d9143b400b6c6ff71bbc69/manifests/init.pp#L324

(again, having some autorequire in the type would be nicer)

manifests/init.pp
```diff
@@ -334,6 +337,8 @@
     fail("${facts['networking']['hostname']}: External authentication via Keycloak can only be enabled when Apache is used.")
   }
 
+  include foreman::register
```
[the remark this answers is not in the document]
adamruzicka: Would an additional parameter be enough for the clone usecase? So in the end we'd have one to control if it should manage the resource or not and the other would control in which state it needs to be?

lib/puppet/type/foreman_host.rb
```diff
@@ -48,4 +48,8 @@
 
     defaultto 500
   end
+
+  autorequire(:class) do
```
ekohl, line 52: Sadly, you can't autorequire classes. I'll see if I can patch up the existing types so you can copy that pattern.

manifests/init.pp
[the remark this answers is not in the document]
ekohl: You now added `$register_in_foreman` and that looks like a good solution to me. We may want to expand the parameter comment a bit, but let's first see if we can get the actual code to work :)

lib/puppet/type/foreman_host.rb
adamruzicka: That would be much appreciated

lib/puppet/type/foreman_instance_host.rb
```diff
@@ -0,0 +1,49 @@
+Puppet::Type.newtype(:foreman_instance_host) do
+  desc 'foreman_instance_host marks a host as foreman.'
```
ehelms, line 2: The language here gets tricky. Borrowing your language from other PRs, maybe something like:

```
foreman_instance_host marks a host as belonging to the set of hosts that make up the Foreman instance/application
```

manifests/register.pp
```diff
@@ -0,0 +1,26 @@
+# @summary Register the host as foreman
```
ehelms, line 1:
```suggestion
# @summary Register the host as Foreman
```

Nit across all of the files in here, Foreman is a proper noun so we should strive to treat it that way in our writing.

manifests/register.pp
```diff
@@ -0,0 +1,26 @@
...
+# @api private
+class foreman::register (
+  Stdlib::Fqdn $foreman_host_name = $facts['networking']['fqdn'],
+) {
+  if $foreman::register_in_foreman {
```
ehelms, line 6: Feels a little odd to have the `if` here as this basically makes the class pointless for someone that included it. I think it also makes the code slightly harder to read for me. When I compare it to:

```
if $foreman::register_in_foreman {
  include foreman::register
}
```

lib/puppet/provider/foreman_host/rest_v3.rb
```diff
@@ -0,0 +1,55 @@
...
+  end
+
+  def create
+    path = "api/v2/hosts/facts"
+    payload = {
+      :name => resource[:name],
+      :certname => resource[:name],
+      :facts => resource[:facts]
+    }
+    r = request(:post, path, {}, payload.to_json)
```
ehelms, line 15: Personal nit, could this at least be `req` or `request`. I find single letter names mental overhead when reading code.

adamruzicka:
The tests should now failing because they're running against foreman without the new required endpoints

lib/puppet/type/foreman_host.rb
ekohl: https://github.com/theforeman/puppet-foreman/pull/966

manifests/register.pp
```diff
@@ -0,0 +1,25 @@
...
+  }
+
+  foreman_instance_host { $foreman_host_name:
+    ensure          => present,
+    base_url        => $foreman::foreman_url,
+    consumer_key    => $foreman::oauth_consumer_key,
+    consumer_secret => $foreman::oauth_consumer_secret,
+    effective_user  => $foreman::oauth_effective_user,
+    ssl_ca          => $foreman::server_ssl_ca,
+    require         => Foreman_host[$foreman_host_name],
```
ekohl, line 23: In the type you can add:
```ruby
autorequire(:foreman_host) do
  [self[:name]]
end
````

Then you can remove this line

Comment (ekohl) [reply manifests/register.pp:23]: https://github.com/theforeman/puppet-foreman_proxy/pull/687 tells me that the smart proxy host can also be an autorequire.

File: lib/puppet/type/foreman_host.rb (hunk 2 above)
Comment (adamruzicka) [reply lib/puppet/type/foreman_host.rb:52]: Just to be sure I got this right before I start making changes. I should place a named anchor into `manifests/init.pp` and then autorequire it from here?

File: lib/puppet/type/foreman_smartproxy_host.rb [hunk 8]
```diff
@@ -0,0 +1,9 @@
+Puppet::Type.newtype(:foreman_smartproxy_host) do
```
ehelms, line 1: Needs `require_relative '../../puppet_x/foreman/common'`

ekohl:
> Maybe @ekohl knows why if both places are using `ensure_resource` it leads to a duplicate resource. At a minimum, we could declare one of them directly and the other use `ensure_resource`.

This can happen if the parameters that are passed in are different. They must match.

ehelms:
Current failures are due to the new Foreman code missing in the Debian repositories. This is from a failing nightly Debian pipeline.

manifests/register.pp
```diff
@@ -0,0 +1,24 @@
...
+class foreman::register (
+  Stdlib::Fqdn $foreman_host_name = $facts['networking']['fqdn'],
+) {
+  ensure_resource('foreman_host', $foreman_host_name, {
+      ensure          => present,
+      base_url        => $foreman::foreman_url,
+      consumer_key    => $foreman::oauth_consumer_key,
+      consumer_secret => $foreman::oauth_consumer_secret,
+      effective_user  => $foreman::oauth_effective_user,
+      ssl_ca          => $foreman::server_ssl_ca,
```
ehelms, line 12: Oh doh, I am an idiot. This CA file is going to be different from the context of Foreman and from the context of the smart-proxy. On Foreman, this path may be the Puppet CA or in Katello's case a path in `/etc/foreman` whereas from the smart-proxy perspective, the CA file used to communicate with Foreman is in `/etc/foreman-proxy`. So we cannot rely on `ensure_resource` giving us the same resource across both modules due to this parameter.

I think then we have two options off the top of my head:

 1) Add a parameter for the hostname and then make the titles unique e.g. `foreman-foreman.example.com` and `foreman-proxy-foreman.example.com`
 2) A class that is shared between puppet-foreman and puppet-foreman_proxy to ensure creation of the host object since puppet-foreman_proxy requires puppet-foreman

ekohl (commented):
Argh, I had a review pending that forgot to submit. The use of `$facts` in `foreman_host.rb` is invalid and resolves to `nil`.

It should also should be reflected in https://github.com/theforeman/puppet-foreman#support-policy that this now targets Foreman 3.1+ but you can set `$register` to `false` for compatibility. That does make this a backwards incompatible change, but so does dropping Ubuntu 18.04 so it's ok.

lib/puppet/provider/foreman_instance_host/rest_v3.rb
```diff
@@ -0,0 +1,30 @@
+Puppet::Type.type(:foreman_instance_host).provide(:rest_v3, :parent => Puppet::Type.type(:foreman_host).provider(:rest_v3)) do
+  confine :feature => [:json, :oauth]
+
+  def exists?
+    return false if host.nil?
+
+    host['foreman']
```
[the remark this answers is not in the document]
adamruzicka: It should always be a boolean

PR #966, opened 2021-07-02
Autorequire provider in smartproxy type

ekohl:
With this it's sufficient to include foreman::providers where prior to this the caller was responsible for ensuring the correct dependency.

### DIFF
--- a/spec/classes/foreman_providers_spec.rb
+++ b/spec/classes/foreman_providers_spec.rb
@@ -8,7 +8,8 @@ describe 'foreman::providers' do
 
       context 'with defaults' do
         it { should compile.with_all_deps }
-        it { should contain_package(oauth_os) }
+        it { should contain_package(oauth_os).that_comes_before('Anchor[foreman::providers::oauth]') }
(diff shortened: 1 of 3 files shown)


spec/classes/foreman_providers_spec.rb
```diff
@@ -8,7 +8,8 @@
 
       context 'with defaults' do
         it { should compile.with_all_deps }
-        it { should contain_package(oauth_os) }
+        it { should contain_package(oauth_os).that_comes_before('Anchor[foreman::providers::oauth]') }
+        it { should contain_anchor('foreman::providers::oauth]') }
```
ehelms, line 12:
```suggestion
        it { should contain_anchor('foreman::providers::oauth') }
```

ekohl: So I ran the tests and then forgot to rerun them after adding this for resource coverage ... I've made this exact mistake a few times :(

ehelms (approved):

PR #967, opened 2021-07-05
Fixes #32352 - use mod_auth_gssapi instead of mod_auth_kerb

evgeni:


manifests/config.pp
```diff
@@ -168,8 +168,14 @@
         ssl_content => template('foreman/lookup_identity.conf.erb'),
       }
 
-      foreman::config::apache::fragment { 'auth_kerb':
-        ssl_content => template('foreman/auth_kerb.conf.erb'),
+      if $facts['os']['family'] == 'RedHat' && versioncmp($facts['os']['release']['major'], '8') >= 0 {
```
evgeni, line 171: technically, we could also use the same on debian, as it has mod_gssapi too.

and actually, EL7 has it too, so probably let's switch to it completely?

templates/auth_gssapi.conf.erb
```diff
@@ -0,0 +1,14 @@
+
+<Location /users/extlogin>
+  SSLRequireSSL
+  AuthType GSSAPI
+  AuthName "GSSAPI Single Sign On Login"
+  GssapiCredStore keytab:<%= scope.lookupvar('::foreman::http_keytab') %>
+  GssapiSSLonly On
```
evgeni, line 7: this is not the default :roll_eyes: 

probably also want to set 
```
GssapiUseSessions …
GssapiAllowedMech krb5
```

manifests/config.pp
```diff
@@ -168,8 +168,14 @@
         ssl_content => template('foreman/lookup_identity.conf.erb'),
       }
 
-      foreman::config::apache::fragment { 'auth_kerb':
-        ssl_content => template('foreman/auth_kerb.conf.erb'),
+      if ($facts['os']['family'] == 'RedHat' and versioncmp($facts['os']['release']['major'], '8') >= 0) {
+        foreman::config::apache::fragment { 'auth_gssapi':
+          ssl_content => template('foreman/auth_gssapi.conf.erb'),
+        }
```
ekohl, line 174: Rather than including a fragment, should we use a directory entry? It was added in https://github.com/puppetlabs/puppetlabs-apache/commit/891fdaf24992e0e96b0ae14e2eca8f1991db373a which means we should also raise the minimum version to 5.7.0 in metadata.json, but that's ok IMHO.

templates/auth_gssapi.conf.erb
```diff
@@ -0,0 +1,14 @@
+
+<Location /users/extlogin>
+  SSLRequireSSL
+  AuthType GSSAPI
+  AuthName "GSSAPI Single Sign On Login"
+  GssapiCredStore keytab:<%= scope.lookupvar('::foreman::http_keytab') %>
```
ekohl, line 6: Since we're Puppet 4 only, `::` prefix is no longer needed since everything is by definition absolute
```suggestion
  GssapiCredStore keytab:<%= scope.lookupvar('foreman::http_keytab') %>
```

ekohl (commented):
I really should find time to set up some FreeIPA server and auth against it.

manifests/config.pp
evgeni: Maybe. Probably.

Right now it's just copy paste to see if it even works :D

manifests/config.pp
evgeni: I did that now, fuck that mod_auth_kerb

templates/auth_gssapi.conf.erb
```diff
@@ -1,15 +1,13 @@
 
 <Location /users/extlogin>
   SSLRequireSSL
-  AuthType Kerberos
-  AuthName "Kerberos Login"
-  KrbMethodNegotiate On
-  KrbMethodK5Passwd Off
-  KrbAuthRealms <%= @facts['foreman_ipa']['default_realm'] %>
```
evgeni, line 8: I couldn't find an equivalent in the `mod_auth_gssapi` config :woman_shrugging:

manifests/config.pp
evgeni: Right now, I'd prefer to skip this, as that would require more changes: when you specify `directories`, you also need to pass all other directories you want to configure, like the one for the docroot, as specifying it makes the defaults be dropped.

ekohl (changes requested):
Could you change this bit to no longer require `default_realm`?
https://github.com/theforeman/puppet-foreman/blob/88ee2c8f30266c3ef6cd8a8fb91bc17b7e3ccfea/manifests/config.pp#L129-L131

ekohl (approved):

templates/auth_gssapi.conf.erb
evgeni: After thinking, `GssapiUseSessions` is not really much of a concern, as we only hit kerberos when the user visits `/extlogin`, not on *every* page load.

evgeni:
Should we call that out in the release notes explicitly?

CC @tbrisker

tbrisker:
@evgeni makes sense. we should also probably update the docs for configuring it. currently it's at https://theforeman.org/manuals/nightly/index.html#5.7ExternalAuthentication, but this might be a good opportunity to actually drop at least part of that section and point it to https://docs.theforeman.org/nightly/Administering_Red_Hat_Satellite/index-foreman-el.html#sect-Administering-Configuring_External_Authentication-Using_Identity_Management with the new instructions.

evgeni:
Ack, will put on my todo

evgeni:
@tbrisker  https://github.com/theforeman/theforeman.org/pull/1857

PR #968, opened 2021-07-06
Fixes #32947 - Use Apache module variables

ekohl:
Apache is packaged in different locations between Red Hat and Debian. The user differs (apache vs www-data) and conf dir (/etc/httpd vs /etc/apache2). This changes the code to use variables already defined on the apache module to avoid duplicating this logic.

### DIFF
--- a/manifests/params.pp
+++ b/manifests/params.pp
@@ -165,7 +165,7 @@ class foreman::params {
   $initial_location = undef
 
   $ipa_authentication = false
-  $http_keytab = '/etc/httpd/conf/http.keytab'
+  $http_keytab = undef
   $pam_service = 'foreman'
   $ipa_manage_sssd = true
 
--- a/templates/auth_gssapi.conf.erb
+++ b/templates/auth_gssapi.conf.erb
@@ -3,7 +3,7 @@
   SSLRequireSSL
   AuthType GSSAPI
   AuthName "GSSAPI Single Sign On Login"
-  GssapiCredStore keytab:<%= scope.lookupvar('foreman::http_keytab') %>
+  GssapiCredStore keytab:<%= @http_keytab %>
   GssapiSSLonly On
   GssapiLocalName On
   # require valid-user
--- a/manifests/config.pp
+++ b/manifests/config.pp
@@ -153,14 +153,16 @@ class foreman::config {
         content => template('foreman/pam_service.erb'),
       }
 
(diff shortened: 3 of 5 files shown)


manifests/params.pp
```diff
@@ -165,7 +165,7 @@
   $initial_location = undef
 
   $ipa_authentication = false
-  $http_keytab = '/etc/httpd/conf/http.keytab'
+  $http_keytab = undef
```
evgeni, line 168: Any reason not to use apache_conf_dir here directly?

evgeni (approved):

manifests/params.pp
ekohl: In params we haven't included apache yet, nor should we.

templates/auth_gssapi.conf.erb
```diff
@@ -3,7 +3,7 @@
   SSLRequireSSL
   AuthType GSSAPI
   AuthName "GSSAPI Single Sign On Login"
-  GssapiCredStore keytab:<%= scope.lookupvar('foreman::http_keytab') %>
+  GssapiCredStore keytab:<%= @http_keytab %>
```
evgeni, line 6: I was about to say that this should have broken https://github.com/theforeman/puppet-foreman/blob/91850a2b4522406a28dbc98e2c3f215130d08dab/spec/classes/foreman_config_ipa_spec.rb#L53

But we don't test this on Debian at all…
https://github.com/theforeman/puppet-foreman/blob/91850a2b4522406a28dbc98e2c3f215130d08dab/spec/classes/foreman_config_ipa_spec.rb#L5

Mind adding tests? :)

evgeni: I took the liberty of pushing some simple tests now ;)

ekohl:
I can't approve my own PR, but :+1: and thanks for fixing the tests.

evgeni:
@tbrisker how noteworthy do you find that Debian can now use Kerberos auth too? :)

ekohl:
Perhaps we can capture it together in "Kerberos is now available on EL8 and Debian"

tbrisker:
Makes sense to add a headline feature regarding the switch to gssapi and its availability on all distros

tbrisker:
and now i see you already did that ^_^

evgeni:
Yupp. And now also explicitly added Debian/Ubuntu in https://github.com/theforeman/theforeman.org/pull/1858

PR #969, opened 2021-07-08
Handle duplicate file declaration for foreman::app_root

chr1s692:
Currently we're using your modul to install and configure Foreman. 
We want to install Foreman in its own filesystem, but the file ressource for $foreman::app_root makes it hard to implement this in puppet - it will always result in duplicate ressource.  Furthermore i'm not sure, if this ressource is necessary or placed in the right class (ordering, first install the package and then create the directory?!).

Testing if this directory is already defined gives other the chance to do some stuff, e.g. creating a filesystem before installing foreman :)

### DIFF
--- a/manifests/config.pp
+++ b/manifests/config.pp
@@ -49,9 +49,11 @@ class foreman::config {
     unit     => "${foreman::foreman_service}.service",
     content  => template('foreman/foreman.service-overrides.erb'),
   }
-
-  file { $foreman::app_root:
-    ensure  => directory,
+  
+  if ! defined(File[$foreman::app_root]) {
+    file { $foreman::app_root:
+      ensure  => directory,
+    }
   }
 
   if $foreman::db_root_cert {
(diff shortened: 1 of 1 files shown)


ekohl (approved):

manifests/config.pp
```diff
@@ -49,9 +49,11 @@
     unit     => "${foreman::foreman_service}.service",
     content  => template('foreman/foreman.service-overrides.erb'),
   }
-
-  file { $foreman::app_root:
-    ensure  => directory,
+
```
ekohl, line 52: Looks like there's trailing whitespace on this line which fails puppet-lint.

ekohl (changes requested):
>  Furthermore i'm not sure, if this ressource is necessary or placed in the right class (ordering, first install the package and then create the directory?!).

Normally the package install creates the directory. So the whole definition may not be needed. So I wonder why you want to manage this in a different way.

manifests/config.pp
chr1s692: oh, you're right. should not be corrected :)

ekohl (approved):
I'm still wondering why you do need to manage it, but I don't mind this.

PR #972, opened 2021-07-11
Let Function to_symbolized_yaml handle Datatype Sensitive

cocker-cc:


cocker-cc:
This is a simplified Version, of what I recently did [here](https://github.com/cocker-cc/puppet-foreman/compare/6c902a4f1f484137ebccafe4d3390881fd0b3d61..1fba024a8d226a0d7c77822c09bce363bc107f96).

ekohl (commented):
Overall I think this looks correct, but is it possible to write a test that it returns a sensitive?

cocker-cc:
> Overall I think this looks correct, but is it possible to write a test that it returns a sensitive?

I guess the Test is implicite with [this Line](https://github.com/theforeman/puppet-foreman/pull/972/files#diff-18f6de7afb3ce17a08c1c683a8262957c6c3532b9f7a685735994b2f071be9d2R12):  
```
expect(subject.execute({'a' => sensitive('b')}).unwrap).to eq("---\n:a: b\n")
```
 If the Returnvalue wasn't `Sensitive` the `unwrap` would fail.

ekohl:
For now. However, now that https://github.com/puppetlabs/puppet/pull/8651 is merged is that still true? My initial reading suggests it's still true, bu I must admit that I have some doubts about the implicit testing. Another way to ensure would be testing that executing with non-sensitive values doesn't respont to `unwrap`.

Another possible test is passing a hash with one sensitive value and one non-sensitive value. That should still return a sensitive.

ekohl (approved):

PR #974, opened 2021-07-21
Fixes #33089 - move (hammer_)plugin_prefix to globals

evgeni:


evgeni:
@ekohl I guess moving `version` to globals would be something you'd be in favor of? What else?

manifests/cli/params.pp
```diff
@@ -15,35 +15,36 @@
...
+        $_hammer_plugin_prefix = 'tfm-rubygem-hammer_cli_'
       }
     }
     'Debian': {
-      $hammer_plugin_prefix = 'ruby-hammer-cli-'
+      $_hammer_plugin_prefix = 'ruby-hammer-cli-'
     }
     'Linux': {
       case $facts['os']['name'] {
         'Amazon': {
```
ekohl, line 28: I just checked and it looks like Amazon Linux is still qualified in the Red Hat OS family. At some point we should clean this up.

evgeni: Not me, not today, not enough :boom: ;)

manifests/cli/plugin.pp
```diff
@@ -9,7 +9,7 @@
 # $version:: The package version to ensure
 #
 define foreman::cli::plugin (
-  String $package = "${foreman::cli::hammer_plugin_prefix}${title}",
+  String $package = "${foreman::cli::params::hammer_plugin_prefix}${title}",
```
ekohl, line 12: Technically this doesn't have to change due to inheritance.

manifests/params.pp
```diff
@@ -93,27 +93,27 @@
       # We use system packages except on EL7
       if versioncmp($facts['os']['release']['major'], '8') >= 0 {
         $passenger_ruby_package = undef
-        $plugin_prefix = 'rubygem-foreman_'
+        $_plugin_prefix = 'rubygem-foreman_'
```
ekohl, line 96: I think for both EL7 & EL8 we can nowadays use `foreman-plugin-` as a prefix. Puppet should support `Provides` and it would simplify matters. For example, the puppetdb plugin also follows the same convention (tasks appears to be `foreman-plugin-foreman-tasks`). Thoughts?

manifests/cli/plugin.pp
evgeni: Maybe. I was blindly copying from https://github.com/theforeman/puppet-foreman_proxy/commit/935d45ab53e59fa5424a60eb54b29ccf6b666313 ;)

manifests/params.pp
evgeni: Would surely make a great follow up PR? ;)

evgeni: But yes, puppet 5.5 already has the `virtual_packages` feature in `yum` and `dnf` providers, so it's probably safe to switch to that.

ekohl: An noting it down explicitly: we don't have a similar provides for Hammer plugins. We should, but that means we can't immediately depend on it.

manifests/cli/plugin.pp
evgeni: Honestly, if I'd read `foreman::cli::hammer_plugin_prefix` here, and wouldn't find it in `cli.pp`, I'd be confused at first, so I think explicitly using the one from `params` seems friendlier to the reader to me

ekohl: I've seen others confused by that as well so I guess it's ok.

ekohl (approved):
I'm ok with this. Let's move other variables in separate PRs.

@ehelms @wbclark please also have a look.

As for other variables:
* `version` & `plugin_version` - I can still see a use case so I'm not entirely sure.
* `user` & `group` - these are hardly ever changed so I'd be in favor of moving them
* `rails_env` & `app_root` same
* `vhost_priority` same (perhaps even move it to `apache.pp` and only allow changing it via Hiera)

manifests/params.pp
ehelms: If we'd like to do that in the future, can an issue be filed to track it?

evgeni: Yes, I was about to open follow up issues for the things we talked about.

evgeni: https://projects.theforeman.org/issues/33095 - Use "Provides" for hammer plugins too
https://projects.theforeman.org/issues/33096 - Use virtual package names for Foreman plugins
https://projects.theforeman.org/issues/33097 - Use virtual package names for Hammer plugins

ekohl: Thanks! Should we file issues for the Foreman Proxy equivalents too while we're at it?

evgeni: Yeah, sounds good. I just didn't think of it as it's not part of this PR

ekohl: Shall I create them or will you? Just so we don't end up with duplicates.

evgeni: You. I'm away from the pc

ekohl: https://projects.theforeman.org/issues/33099 / https://projects.theforeman.org/issues/33100 / https://projects.theforeman.org/issues/33101

evgeni:
Thanks!

I've also opened https://projects.theforeman.org/issues/33106 to track the extraction of the other variables

PR #976, opened 2021-07-23
Replace use of verify_exact_contents

ekohl:
This can be tested with native methods. The only difference is that now we need to test for comments and whitespace too, but this is not a bad thing.


ehelms (approved):
Looks like tests here exposed some package dependency updates that are needed

ekohl:
Yes, https://ci.theforeman.org/job/foreman-plugins-nightly-rpm-pipeline/ is waiting on a release of the webhooks plugin.

PR #979, opened 2021-08-03
Add hammer plugin for foreman_puppet

amirfefer:


ekohl (commented):
Code looks good but I expect the acceptance test to fail. The packages have not yet been published. Somehow they appear to stop after 3 seconds. I have a theory why they fail.

ekohl:
https://github.com/theforeman/puppet-foreman/pull/980 should fix that.

ekohl (approved):
We'll fix the Debian packages later.

PR #981, opened 2021-08-03
Add Ubuntu 20.04 support & drop Ubuntu 18.04

ekohl:
Foreman 2.5 introduced support for Ubuntu Focal. This should be reflected here.

### DIFF
--- a/metadata.json
+++ b/metadata.json
@@ -81,7 +81,7 @@
     {
       "operatingsystem": "Ubuntu",
       "operatingsystemrelease": [
-        "18.04"
+        "20.04"
       ]
     }
   ]


ehelms (approved):

ekohl:
I suspect this fails because somehow the Puppet Ruby is used.
```
2021-08-24T15:21:56.8404522Z ubuntu2004-64.example.com 15:21:56$ /bin/sh -c hammer\ --version
2021-08-24T15:21:56.9090123Z [00;00m[00;33m  /opt/puppetlabs/puppet/lib/ruby/2.5.0/rubygems/core_ext/kernel_require.rb:59:in `require': cannot load such file -- clamp (LoadError)
2021-08-24T15:21:56.9091547Z [00;00m[00;33m  	from /opt/puppetlabs/puppet/lib/ruby/2.5.0/rubygems/core_ext/kernel_require.rb:59:in `require'
2021-08-24T15:21:56.9092380Z   	from /usr/bin/hammer:4:in `<main>'
```

I wonder why this fails on Ubuntu 20.04 but not 18.04. Perhaps because 18.04 ships Ruby 2.5 and somehow manages to find the system libraries?

For now I'd say we shouldn't hold 18.2.0 on this, but at least investigate.

evgeni:
I wonder why that happens (and why we don't see the same in forklift).

ekohl:
Beaker does insert some extra things to `$PATH`. https://community.theforeman.org/t/foreman-installer-command-dosent-work/25006 looks verify similar and I wonder if there are incorrect shebangs.

ekohl:
You're right, but I wonder if it's a problem with the space between `#!` and `/usr/bin/env`. Perhaps that breaks some automated replacement script. It is weird that the shebang is correct on Bionic. Nothing changed in the packaging (diff only shows differences at the seconds level in changelog).

evgeni:
Maybe, yeah. Easy to test ;)

ekohl:
Let's see if https://github.com/theforeman/foreman-packaging/pull/7217 builds and solves it.

ekohl:
Looks like it does. I've opened https://github.com/theforeman/hammer-cli/pull/355.

ekohl:
I suspect that it's now failing because the Debian pipeline hasn't completed in a while and this module is no longer compatible with it (due to the registration bits).

evgeni:
nightly is finally green, kicked GHA

ekohl:
I'm not quite sure what causes this. The only thing that comes to mind is https://github.com/puppetlabs/puppetlabs-postgresql/commit/2f6ab27e32565c5b246a5fdc0578c1a5d230d440 which is unrelased for now. However, we should be using a git checkout. I'm going to try to pin it to an older version to see if that improves things.

ehelms:
This passes locally and is only failing inside GHA. Therefore we are deciding to merge this and accept the CI failures for investigation in the future.

PR #984, opened 2021-08-04
Fixes #33214: Set minimum Puma threads equal to maximum puma threads …

ehelms:
…by default

This has been shown to prevent memory leaks that can appear in Foreman
when running under Puma. This only changes the default to match and
users must still be aware of this potential when changing the tunings.

Some guidance from Satellite performance team -- https://github.com/RedHatSatellite/satellite-performance-tuning/blob/devel/docs/satellite-configuration-tuning.rst#puma-tuning

Some info from upstream Puma -- https://github.com/puma/puma/issues/2665

There is most likely one or more memory leaks within the application itself leading to this issue. This alleviates the issue for now by pre-allocating the needed memory and preventing spin-up and spin-down of threads that exacerbates the memory leak.

### DIFF
--- a/manifests/params.pp
+++ b/manifests/params.pp
@@ -63,7 +63,7 @@ class foreman::params inherits foreman::globals {
   $foreman_service = 'foreman'
   $foreman_service_ensure = 'running'
   $foreman_service_enable = true
-  $foreman_service_puma_threads_min = 0
+  $foreman_service_puma_threads_min = undef
   $foreman_service_puma_threads_max = 16
   $foreman_service_puma_workers = 2
 
--- a/templates/foreman.service-overrides.erb
+++ b/templates/foreman.service-overrides.erb
@@ -2,6 +2,6 @@
 User=<%= scope['foreman::user'] %>
 Environment=FOREMAN_ENV=<%= scope['foreman::rails_env'] %>
 Environment=FOREMAN_HOME=<%= scope['foreman::app_root'] %>
-Environment=FOREMAN_PUMA_THREADS_MIN=<%= scope['foreman::foreman_service_puma_threads_min'] %>
+Environment=FOREMAN_PUMA_THREADS_MIN=<%= scope['foreman::config::min_puma_threads'] %>
 Environment=FOREMAN_PUMA_THREADS_MAX=<%= scope['foreman::foreman_service_puma_threads_max'] %>
 Environment=FOREMAN_PUMA_WORKERS=<%= scope['foreman::foreman_service_puma_workers'] %>
--- a/manifests/config.pp
+++ b/manifests/config.pp
@@ -44,6 +44,7 @@ class foreman::config {
     content => template('foreman/database.yml.erb'),
   }
 
+  $min_puma_threads = pick($foreman::foreman_service_puma_threads_min, $foreman::foreman_service_puma_threads_max)
   systemd::dropin_file { 'foreman-service':
     filename => 'installer.conf',
     unit     => "${foreman::foreman_service}.service",
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -160,7 +160,9 @@
 #
 # $cors_domains::                 List of domains that show be allowed for Cross-Origin Resource Sharing
 #
-# $foreman_service_puma_threads_min::     Minimum number of threads for every Puma worker
+# $foreman_service_puma_threads_min::     Minimum number of threads for every Puma worker. If no value is specified, this defaults
+#                                         to setting min threads to maximum threads. Setting min threads equal to max threads has
+#                                         been shown to alleviate memory leaks and in some cases produce better performance.
 #
 # $foreman_service_puma_threads_max::     Maximum number of threads for every Puma worker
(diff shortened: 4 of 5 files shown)


ehelms:
The alternative would be to drop the minimum parameter and always force `min = max`, that requires a major version change. The upside would be that it prevents users from putting themselves accidentally into a situation of having a memory leak. This is a tough challenge as ideally we would just eliminate all the memory leaks and all users to tune things however they want. Finding the leaks is the hard part.

manifests/params.pp
```diff
@@ -63,7 +63,7 @@
   $foreman_service = 'foreman'
   $foreman_service_ensure = 'running'
   $foreman_service_enable = true
-  $foreman_service_puma_threads_min = 0
+  $foreman_service_puma_threads_min = 16
```
ekohl, line 66: How about we make it dynamic and set it to undef:
```suggestion
  $foreman_service_puma_threads_min = undef
```

Then use `$min = pick($foreman::foreman_service_puma_threads_min, $foreman::foreman_service_puma_threads_max)`. That way users don't need to specify it twice if they want them to match.

You probably also need to change the type to `Optional[Integer[0]]`.

Also an installer migration that clears the answer if they match will be nice for users.

ekohl (commented):
Overall :+1: for making them match, but I'd prefer to have it dynamic.

manifests/params.pp
ehelms: I like the dynamic aspect from a code perspective. I am wondering if this will confuse users of the installer who look at it and see that the minimum is set to nothing. Asking out loud if that will lead to support posts or users to think that the minimum is therefore 0 if blank.

ekohl: We can document this in the parameter help that the default (undef) means matching max threads.

ehelms: Ehh, I'll update the documentation of the parameter and let that suffice.

ekohl:
Do you recall if min threads set to 1 also works around the leak? Or does matching mean it doesn't rotate out old threads which means the leak doesn't show up?

ehelms:
> Do you recall if min threads set to 1 also works around the leak? Or does matching mean it doesn't rotate out old threads which means the leak doesn't show up?

I do not know if 1 also works around the leak. I can test this.

My understanding was, that by setting `min = max` and pre-allocating the memory for each thread, the leak does not occur as there is no spin-up and spin-down.

ehelms:
Another note, the Satellite performance team found generally better overall performance when `min = max`. Further, the Gitlab project sets their defaults to `min = max` as well for their helm chart (https://docs.gitlab.com/charts/charts/gitlab/webservice/index.html#installation-command-line-options -- search for threads).

ekohl:
Sounds like we should start with `min = max` then.

ehelms:
Failing tests are getting fixed with https://github.com/theforeman/foreman-packaging/pull/7001

wbclark (commented):
still requires the update to param documentation at manifests/init.pp#L163

it would be nice also to see an additional unit test for correct behavior of non-default params

templates/foreman.service-overrides.erb
```diff
@@ -2,6 +2,6 @@
 User=<%= scope['foreman::user'] %>
 Environment=FOREMAN_ENV=<%= scope['foreman::rails_env'] %>
 Environment=FOREMAN_HOME=<%= scope['foreman::app_root'] %>
-Environment=FOREMAN_PUMA_THREADS_MIN=<%= scope['foreman::foreman_service_puma_threads_min'] %>
+Environment=FOREMAN_PUMA_THREADS_MIN=<%= @min_puma_threads %>
```
wbclark, line 5: it feels a little funny to me that puppet's erb templating supports `@variable` in the local scope from where the template is called. is it more efficient in case the scope object is truly huge? consistently using the same syntax would be nicer to look at otherwise.

wbclark:
I read the previous note about a migration that would clear the value of the minimum if it matches the maximum. :+1: to that.

I think we should also in that migration clear the value of the minimum if it used the old default of 0. this would alleviate the memory leaks for the vast majority of users not using any non-default params, while still allowing users to set a non-default value of the minimum if they really desire it.

ehelms:
> I read the previous note about a migration that would clear the value of the minimum if it matches the maximum.  to that.
> 
> I think we should also in that migration clear the value of the minimum if it used the old default of 0. this would alleviate the memory leaks for the vast majority of users not using any non-default params, while still allowing users to set a non-default value of the minimum if they really desire it.

That's over here -- https://github.com/theforeman/foreman-installer/pull/713

wbclark (approved):
Overall this looks good to me. The new param documentation mentions the leaks and performance benefits of using min = max. It's possible there is some unknown future condition where separately configuring the min to be different from the max would still be desirable.

wbclark:
@ehelms I went with 'bug' for now. It's not so much adding a feature that we traditionally think of as an 'enhancement', and the intent behind it is very much 'fix a thing that wasn't working the way it was intended'

PR #986, opened 2021-08-17
Fixes #33277: Change Puma default workers to 1.5 * CPU, max threads to 5

ehelms:
The Puma documentation recommends for MRI based installations to
start with number of CPUs times 1.5 and defaults to 5 on Ruby MRI.
The previous maximum threads of 16 was based on an incorrect reading
of the Puma documentation and is only the default on non-MRI Ruby.

I am recommending we make these default changes (and consider resetting existing installations that match the old defaults by way of an installer migration) based upon some of the following info:

 * Puma recommendations for MRI: https://github.com/puma/puma/blob/master/docs/deployment.md#mri
 * Puma worker utilization: https://github.com/puma/puma/blob/master/docs/deployment.md#worker-utilization
 * Gitlab's thread guidance: https://docs.gitlab.com/ee/install/requirements.html#puma-threads

### DIFF
--- a/manifests/config.pp
+++ b/manifests/config.pp
@@ -44,6 +44,17 @@ class foreman::config {
     content => template('foreman/database.yml.erb'),
   }
 
+  # CPU based calculation is based on https://github.com/puma/puma/blob/master/docs/deployment.md#mri
+  # Memory based calculation is based on https://docs.gitlab.com/ee/install/requirements.html#puma-settings
+  $puma_workers = pick(
+    $foreman::foreman_service_puma_workers,
+    floor(
+      min(
+        $facts['processors']['count'] * 1.5,
+        ($facts['memory']['system']['total_bytes']/(1024 * 1024 * 1024)) - 1.5
+      )
+    )
+  )
   $min_puma_threads = pick($foreman::foreman_service_puma_threads_min, $foreman::foreman_service_puma_threads_max)
   systemd::dropin_file { 'foreman-service':
     filename => 'installer.conf',
--- a/manifests/params.pp
+++ b/manifests/params.pp
@@ -64,8 +64,8 @@ class foreman::params inherits foreman::globals {
   $foreman_service_ensure = 'running'
   $foreman_service_enable = true
   $foreman_service_puma_threads_min = undef
-  $foreman_service_puma_threads_max = 16
-  $foreman_service_puma_workers = 2
+  $foreman_service_puma_threads_max = 5
+  $foreman_service_puma_workers = undef
 
   # Define job processing service properties
   $dynflow_manage_services = true
--- a/spec/classes/foreman_spec.rb
+++ b/spec/classes/foreman_spec.rb
@@ -3,7 +3,7 @@ require 'spec_helper'
 describe 'foreman' do
   on_supported_os.each do |os, facts|
     context "on #{os}" do
-      let(:facts) { facts }
+      let(:facts) { override_facts(facts, processors: { count: 3 }, memory: { system: { total_bytes:  10_737_418_240}}) }
       let(:params) { {} }
       let(:apache_user) { facts[:osfamily] == 'Debian' ? 'www-data' : 'apache' }
 
@@ -100,9 +100,9 @@ describe 'foreman' do
             .with_content(%r{^User=foreman$})
             .with_content(%r{^Environment=FOREMAN_ENV=production$})
             .with_content(%r{^Environment=FOREMAN_HOME=/usr/share/foreman$})
-            .with_content(%r{^Environment=FOREMAN_PUMA_THREADS_MIN=16$})
-            .with_content(%r{^Environment=FOREMAN_PUMA_THREADS_MAX=16$})
-            .with_content(%r{^Environment=FOREMAN_PUMA_WORKERS=2$})
+            .with_content(%r{^Environment=FOREMAN_PUMA_THREADS_MIN=5$})
+            .with_content(%r{^Environment=FOREMAN_PUMA_THREADS_MAX=5$})
+            .with_content(%r{^Environment=FOREMAN_PUMA_WORKERS=4$})
         end
 
         it { should contain_apache__vhost('foreman').without_custom_fragment(/Alias/) }
--- a/templates/foreman.service-overrides.erb
+++ b/templates/foreman.service-overrides.erb
@@ -4,4 +4,4 @@ Environment=FOREMAN_ENV=<%= scope['foreman::rails_env'] %>
 Environment=FOREMAN_HOME=<%= scope['foreman::app_root'] %>
 Environment=FOREMAN_PUMA_THREADS_MIN=<%= scope['foreman::config::min_puma_threads'] %>
 Environment=FOREMAN_PUMA_THREADS_MAX=<%= scope['foreman::foreman_service_puma_threads_max'] %>
-Environment=FOREMAN_PUMA_WORKERS=<%= scope['foreman::foreman_service_puma_workers'] %>
+Environment=FOREMAN_PUMA_WORKERS=<%= scope['foreman::config::puma_workers'] %>
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -166,7 +166,9 @@
 #
 # $foreman_service_puma_threads_max::     Maximum number of threads for every Puma worker
 #
-# $foreman_service_puma_workers::         Number of workers for Puma
+# $foreman_service_puma_workers::         Number of workers for Puma.
+#                                         If not set, the value is dynamically calculated based on available number of
+#                                         CPUs and memory.
 #
 # $rails_cache_store::            Set rails cache store
 #
@@ -276,7 +278,7 @@ class foreman (
   Array[Stdlib::HTTPUrl] $cors_domains = $foreman::params::cors_domains,
   Optional[Integer[0]] $foreman_service_puma_threads_min = $foreman::params::foreman_service_puma_threads_min,
   Integer[0] $foreman_service_puma_threads_max = $foreman::params::foreman_service_puma_threads_max,
-  Integer[0] $foreman_service_puma_workers = $foreman::params::foreman_service_puma_workers,
+  Optional[Integer[0]] $foreman_service_puma_workers = $foreman::params::foreman_service_puma_workers,
   Hash[String, Any] $rails_cache_store = $foreman::params::rails_cache_store,
   Boolean $keycloak = $foreman::params::keycloak,
   String[1] $keycloak_app_name = $foreman::params::keycloak_app_name,


manifests/params.pp
```diff
@@ -64,8 +64,8 @@
   $foreman_service_ensure = 'running'
   $foreman_service_enable = true
   $foreman_service_puma_threads_min = undef
-  $foreman_service_puma_threads_max = 16
-  $foreman_service_puma_workers = 2
+  $foreman_service_puma_threads_max = 5
+  $foreman_service_puma_workers = $facts['processors']['count'] * 1.5
```
ekohl, line 68: Should we have some upper bound as well? In the past we've seen that too many workers could also easily consume way too much memory.

ehelms:
We could borrow from Gitlab and do the min of 1.5 * CPU and (memory - 1.5).

manifests/params.pp
wbclark:
```suggestion
  $foreman_service_puma_workers = ($facts['processors']['count'] * 3) / 2
```

I think this works better since `$facts['processors']['count']` isn't guaranteed to be an even number -- there are a few physical CPUs that have an odd number of cores, or a VM could be created which has an arbitrary number of CPU cores.

This way we guarantee the value of `$foreman_service_puma_workers` remains an integer.

ekohl: Good point, but I'd prefer to use [floor](https://puppet.com/docs/puppet/6/function.html#floor) then. `1.5` is easier to read than `* 3 / 2`. Especially since it depends on the language how integer division happens (for example, Python 2 and Python 3 actually have different behavior).
```suggestion
  $foreman_service_puma_workers = floor($facts['processors']['count'] * 1.5)
```

ekohl:
> We could borrow from Gitlab and do the min of 1.5 * CPU and (memory - 1.5).

What's memory here? Memory in GB?

ehelms:
> > We could borrow from Gitlab and do the min of 1.5 * CPU and (memory - 1.5).
> 
> What's memory here? Memory in GB?

Correct

ehelms:
I am working on a version with memory considered, while I fix up test this would look something like:

```
  $foreman_service_puma_workers = floor(min($facts['processors']['count'] * 1.5, (($facts['memory']['system']['total_bytes'] * 1024 * 1024 * 1024) - 1.5)))
```

ekohl:
Something else: we talked about making things more dynamic. If we do this in `params.pp` it'll be static for the installer (since the answer is saved). If we make the default undef and do the calculation right before we determine the real min threads value (was that `config.pp`?) it'll be dynamic. Maybe something we also want to consider in this PR.

ehelms:
> Something else: we talked about making things more dynamic. If we do this in `params.pp` it'll be static for the installer (since the answer is saved). If we make the default undef and do the calculation right before we determine the real min threads value (was that `config.pp`?) it'll be dynamic. Maybe something we also want to consider in this PR.

That does make this more robust to memory or CPU changes which is a nice feature. Users would get an automatically scaling number of workers.  The only downside is users understanding that in the installer when looking to the installers answer file or the help text for knowing what the currently configured value is. They have to know to go look at the systemd file or use systemctl to print out that info. While not a blocker, it just adds knowledge complexity. 

So two workflows: 

 1) We store it in answers and give guidance to users: If you adjust CPU or memory, unset this value and re-run the installer
 2) We don't store it, and give guidance to users: If you adjust CPU or memory, re-run the installer

ekohl:
That's why I said "consider". I'm fine with leaving that out for now. However, we should try to find an answer. Perhaps do a poll in the community?

spec/classes/foreman_spec.rb
```diff
@@ -6,6 +6,7 @@
       let(:facts) { facts }
       let(:params) { {} }
       let(:apache_user) { facts[:osfamily] == 'Debian' ? 'www-data' : 'apache' }
+      let(:puma_workers) { [facts[:processors]['count'] * 1.5, ((facts[:memory]['system']['total_bytes'] * 1024 * 1024 * 1024) - 1.5)].min.floor }
```
ekohl, line 9: How about we override the processors and memory facts instead? I think if you modify line 6 to be something like this it should work (but I may have missed some braces somewhere):
```suggestion
      let(:facts) { override_facts(facts, processors: { count: 3 }, memory: { system: { total_bytes: SOMETHING }}) }
```

spec/classes/foreman_spec.rb
```diff
@@ -3,7 +3,7 @@
 describe 'foreman' do
   on_supported_os.each do |os, facts|
     context "on #{os}" do
-      let(:facts) { facts }
+      let(:facts) { override_facts(facts, processors: { count: 3 }, memory: { system: { total_bytes:  10737418240}}) }
```
ekohl, line 6: In Ruby it's valid to write:
```suggestion
      let(:facts) { override_facts(facts, processors: { count: 3 }, memory: { system: { total_bytes: 10_737_418_240}}) }
```

I think that makes it easier to spot it's 10 GB.

ehelms:
Thinking more, if we did set this optional it would allow things like
installer tuning profiled to set this value via hiera given where those
tunings exist in the heirarchy.

ehelms:
This now moves Puma workers to be optional, and if not defined then it uses the prior mentioned calculation to determine what to set this as. This will allow more dynamic setting of this value based on system resources and allow tuning profiles to set these values.

manifests/config.pp
```diff
@@ -45,6 +45,7 @@
   }
 
   $min_puma_threads = pick($foreman::foreman_service_puma_threads_min, $foreman::foreman_service_puma_threads_max)
+  $puma_workers = pick($foreman::foreman_service_puma_workers, floor(min($facts['processors']['count'] * 1.5, (($facts['memory']['system']['total_bytes']/(1024 * 1024 * 1024)) - 1.5))))
```
wbclark, line 48:
```suggestion
  # CPU based calculation is based on https://github.com/puma/puma/blob/master/docs/deployment.md#mri
  $default_puma_workers_cpu_limit = floor($facts['processors']['count'] * 1.5)
  # Memory based calculation is based on https://docs.gitlab.com/ee/install/requirements.html#puma-settings
  $default_puma_workers_memory_limit = floor($facts['memory']['system']['total_bytes']/(1024 * 1024 * 1024) - 1.5)
  # The default is the min of these two limits. It is not saved as a parameter, to allow it to be more dynamic in case
  # of hardware resource changes
  $default_puma_workers = min($default_puma_workers_cpu_limit, $default_puma_workers_memory_limit)
  # It can still be overridden if the user desires to set it to a fixed value
  $puma_workers = pick($foreman::foreman_service_puma_workers, $default_puma_workers)
```

Trying to break this up a bit for readability, and add comments explaining why it is structured this way

wbclark:
```suggestion
  # CPU based calculation is based on https://github.com/puma/puma/blob/master/docs/deployment.md#mri
  # Memory based calculation is based on https://docs.gitlab.com/ee/install/requirements.html#puma-settings
  # The default is the min of these two limits. It is not saved as a parameter, to allow it to be more dynamic in case
  # of hardware resource changes
  $default_puma_workers_cpu_limit = floor($facts['processors']['count'] * 1.5)
  $default_puma_workers_memory_limit = floor($facts['memory']['system']['total_bytes']/(1024 * 1024 * 1024) - 1.5)
  $default_puma_workers = min($default_puma_workers_cpu_limit, $default_puma_workers_memory_limit)

  # The default can still be overridden if the user desires to set it to a fixed value
  $puma_workers = pick($foreman::foreman_service_puma_workers, $default_puma_workers)
```

ekohl: I'm not sure you need additional variables. Alternative:
```suggestion
  $puma_workers = pick(
    $foreman::foreman_service_puma_workers,
    floor(min(
      $facts['processors']['count'] * 1.5,
      ($facts['memory']['system']['total_bytes']/(1024 * 1024 * 1024)) - 1.5
     ))
   )
```

ehelms: I did a hybrid of these suggestions, adding some inline documentation, updating the param documentation and updating the style of the code.

ehelms:
Installer migration -- https://github.com/theforeman/foreman-installer/pull/714

wbclark (approved):
At this point it looks great to me

wbclark:
We discussed this some more and one idea that came up was making `foreman_service_puma_threads_max` also Optional and using a similar pattern where the default is determined dynamically.

This allows users to start with a baseline that is based on the recommendations from Puma documentation. Then as we gain knowledge about what works best for Katello specifically, those values can be defined in the tuning profiles and will override the module defaults here unless the user has already supplied their own values.

The idea is that the tuning profiles can also start with initial values that are based on the same calculations used in the module defaults, and can be updated in the future as our knowledge improves. An installer migration would reset manually values, with a release note explaining the change.

Users of the individual puppet modules would still have any of their settings persisted unless they clear them to opt-in to the new defaults defined here.

Users of foreman-installer would have their settings reset at first, but they could be manually set again. If they are not manually set again, foreman-installer users would get settings from their applied tuning profile. As the tuning profiles get updated based on new data, they would automatically receive the performance improvement benefits unless they have opted out by supplying their own values.

ekohl:
I feel like that would be premature optimization now. At this point we're not really certain what the best practice will be. If we migrate all our settings from one style to the other and then find out it doesn't work well it's going to be painful.

Having a static value does mean you can use --(full-)help to see the value which can make it easier to admins to know what's going to happen on their system.

Currently we have a technical limitation where we can't modify answers with our tuning, but perhaps there are ways around that. We'll also need to think about how we want to really implement auto tuning. There are several ways I can think of.

So for now I'd like to table that a bit, especially since I'm really not sure if we'll end up tuning threads dynamically.

templates/foreman.service-overrides.erb
```diff
@@ -4,4 +4,4 @@ Environment=FOREMAN_ENV=<%= scope['foreman::rails_env'] %>
 Environment=FOREMAN_HOME=<%= scope['foreman::app_root'] %>
 Environment=FOREMAN_PUMA_THREADS_MIN=<%= scope['foreman::config::min_puma_threads'] %>
 Environment=FOREMAN_PUMA_THREADS_MAX=<%= scope['foreman::foreman_service_puma_threads_max'] %>
-Environment=FOREMAN_PUMA_WORKERS=<%= scope['foreman::foreman_service_puma_workers'] %>
+Environment=FOREMAN_PUMA_WORKERS=<%= scope['foreman::config::puma_workers'] %>
```
ekohl, line 7: This is generated from the `foreman::config` scope so I think this is equal:
```suggestion
Environment=FOREMAN_PUMA_WORKERS=<%= @puma_workers %>
```

The same is true for `min_puma_threads`.

ekohl (approved):
I think I'm happy with this. Small suggestion in the template, but both work.

PR #987, opened 2021-08-19
Construct foreman_url from servername and drop parameter

ehelms:
Drops the foreman_url parameter and constructs it from the servername
given to Apache. The foreman_url and servername should match
and this gives one less parameter that needs to be configured to get
a correct installation.

@ekohl I was not sure if this is what you mean tor if there is some nuance of these parameters that I am missing where you would want them to be different. I also realize this is a "breaking" change but if accurate, feels cleaner.

### DIFF
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -48,8 +48,6 @@
 #
 # === Advanced parameters:
 #
-# $foreman_url::                  URL on which foreman is going to run
-#
 # $unattended::                   Should Foreman manage host provisioning as well
 #
 # $unattended_url::               URL hosts will retrieve templates from during build (normally http as many installers don't support https)
@@ -203,7 +201,6 @@
 # $keycloak_realm::               The realm as passed to keycloak-httpd-client-install
 #
 class foreman (
-  Stdlib::HTTPUrl $foreman_url = $foreman::params::foreman_url,
   Boolean $unattended = true,
   Optional[Stdlib::HTTPUrl] $unattended_url = undef,
   Boolean $apache = true,
@@ -302,6 +299,8 @@ class foreman (
   } else {
     $db_sslmode_real = $db_sslmode
   }
+
+  $foreman_url = "https://${servername}"
 
   include foreman::install
   include foreman::config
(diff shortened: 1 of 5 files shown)


manifests/init.pp
```diff
@@ -296,6 +293,8 @@
     timeout => 0,
   }
 
+  $foreman_url = "https://${foreman::servername}"
```
ekohl, line 296: This can be simplified to using a local variable.
```suggestion
  $foreman_url = "https://${servername}"
```

ekohl:
@ehelms any plans to revisit this?

ehelms:
I do. And I am wondering how to integrate this into a related request. If we look at https://bugzilla.redhat.com/show_bug.cgi?id=2127515 it denotes multiple URLs that the user wishes could be driven off a single installer input to update:

--foreman-foreman-url
--foreman-unattended-url
--foreman-proxy-foreman-base-url
--foreman-proxy-puppet-url
--foreman-proxy-template-url
--puppet-server-foreman-url

ekohl:
> --foreman-unattended-url

I started some investigations into if we can actually move to a world where the whole unattended URL is no longer needed. It's going to be a long one because you want kickstart to get the template over HTTPS (which it can do, even today) while establishing a trusted chain (which we haven't solved today).

The main benefit is that we can get rid of serving HTTP on port 80 using Puma altogether and move everything to HTTPS on port 443. Apache can keep a redirect to HTTPS for users.

Given it's going to be long term, we should probably simplify it in the installer in the short term.

> --foreman-proxy-foreman-base-url
> --foreman-proxy-puppet-url
> --foreman-proxy-template-url

These all live in the puppet-foreman_proxy module, so should be considered out of scope here.

> --puppet-server-foreman-url

This lives in puppet-puppet, so also out of scope for this PR.

I've been thinking about how to link them up on the same system. You can use Hiera aliases, but for technical reasons that now interferes with how Kafo treats the answer file.

What you can do in those other modules is default to `undef` and then later in the code use `lookup()` explicitly, but that requires some thought.

PR #988, opened 2021-08-23
Fixes #33320 - Refer to FQDN instead of "Foreman server" in SmartProx…

wbclark:
…y registration error messages

### DIFF
--- a/lib/puppet/provider/foreman_resource/rest_v3.rb
+++ b/lib/puppet/provider/foreman_resource/rest_v3.rb
@@ -99,14 +99,16 @@ Puppet::Type.type(:foreman_resource).provide(:rest_v3) do
   end
 
   def error_message(response)
+    fqdn = URI::parse(resource[:base_url]).host
+
     explanations = {
-      '400' => 'Something is wrong with the data sent to Foreman server',
-      '401' => 'Often this is caused by invalid Oauth credentials',
-      '404' => 'The requested resource was not found',
-      '500' => 'Check /var/log/foreman/production.log on Foreman server for detailed information',
-      '502' => 'The webserver received an invalid response from the backend service. Was Foreman unable to handle the request?',
-      '503' => 'The webserver was unable to reach the backend service. Is foreman.service running?',
-      '504' => 'The webserver timed out waiting for a response from the backend service. Is Foreman under unusually heavy load?'
+      '400' => "Something is wrong with the data sent to Foreman at #{fqdn}",
+      '401' => "Often this is caused by invalid Oauth credentials sent to Foreman at #{fqdn}",
+      '404' => "The requested resource was not found in Foreman at #{fqdn}",
+      '500' => "Check /var/log/foreman/production.log on #{fqdn} for detailed information",
+      '502' => "The webserver received an invalid response from the backend service. Was Foreman at #{fqdn} unable to handle the request?",
+      '503' => "The webserver was unable to reach the backend service. Is Foreman running at #{fqdn}?",
+      '504' => "The webserver timed out waiting for a response from the backend service. Is Foreman at #{fqdn} under unusually heavy load?"
     }
 
     if (explanation = explanations[response.code.to_str])
--- a/spec/unit/foreman_resource_rest_v3_spec.rb
+++ b/spec/unit/foreman_resource_rest_v3_spec.rb
@@ -141,6 +141,10 @@ describe provider_class do
   end
 
   describe '#error_message(response)' do
+    before { expect(resource).to receive(:[]).with(:base_url).and_return(base_url) }
+
+    let(:base_url) { 'https://foreman.example.com' }
+
     it 'returns array of errors from JSON' do
       expect(provider.error_message(double(:body => '{"error":{"full_messages":["error1","error2"]}}', :code => 'dummycode'))).to eq('error1 error2')
     end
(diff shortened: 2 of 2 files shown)


lib/puppet/provider/foreman_resource/rest_v3.rb
```diff
@@ -55,6 +55,11 @@ def generate_token
     OAuth::AccessToken.new(oauth_consumer)
   end
 
+  def fqdn
+    # remove protocol and path
+    resource[:base_url].gsub(%r{\Ahttps?:\/\/}, '').gsub(%r{\/.*\z}, '')
```
evgeni, line 60: I like the overall change, the only question I'd have is whether we want `gsub` instead of `URI::parse(base_url).host`?
([It's an regex underneath `URI::parse` too](https://github.com/ruby/ruby/blob/master/lib/uri/rfc3986_parser.rb), so there shouldn't be *too* much performance difference)

evgeni: But given you call that multiple times anyways, maybe caching the result would be good either way.

wbclark: Much nicer. Thanks! :)

lib/puppet/provider/foreman_resource/rest_v3.rb
```diff
@@ -55,6 +55,10 @@ def generate_token
     OAuth::AccessToken.new(oauth_consumer)
   end
 
+  def fqdn
+    URI::parse(resource[:base_url]).host
```
evgeni, line 59:
```suggestion
    @fqdn ||= URI::parse(resource[:base_url]).host
```

So you end up calling `URI::parse` only once

evgeni: (one day I will decide whether I like Ruby's "magic returns" or not…)

wbclark: Done.

lib/puppet/provider/foreman_resource/rest_v3.rb
```diff
@@ -100,13 +104,13 @@ def success?(response)
...
-      '400' => 'Something is wrong with the data sent to Foreman server',
+      '400' => "Something is wrong with the data sent to #{fqdn}",
       '401' => 'Often this is caused by invalid Oauth credentials',
       '404' => 'The requested resource was not found',
-      '500' => 'Check /var/log/foreman/production.log on Foreman server for detailed information',
-      '502' => 'The webserver received an invalid response from the backend service. Was Foreman unable to handle the request?',
-      '503' => 'The webserver was unable to reach the backend service. Is foreman.service running?',
-      '504' => 'The webserver timed out waiting for a response from the backend service. Is Foreman under unusually heavy load?'
+      '500' => "Check /var/log/foreman/production.log on #{fqdn} for detailed information",
+      '502' => "The webserver received an invalid response from the backend service. Was #{fqdn} unable to handle the request?",
```
ekohl, line 111: I think in this case we refer to the service Foreman, not the server. I'm not sure this statement still makes sense.

lib/puppet/provider/foreman_resource/rest_v3.rb
```diff
@@ -55,6 +55,10 @@ def generate_token
     OAuth::AccessToken.new(oauth_consumer)
   end
 
+  def fqdn
+    @fqdn ||= URI::parse(resource[:base_url]).host
+  end
```
ekohl, line 60: You could simplify it and leave this method out. Then in `error_message` you can locally define `fqdn = URI::parse(resource[:base_url]).host`.

ekohl (commented):
Overall :+1: for the concept.

lib/puppet/provider/foreman_resource/rest_v3.rb
evgeni: True!

lib/puppet/provider/foreman_resource/rest_v3.rb
wbclark: I will change this to `"The webserver received an invalid response from the backend service. Was foreman.service at #{fqdn} unable to handle the request?"`

lib/puppet/provider/foreman_resource/rest_v3.rb
wbclark: Cool, this way we call `URI::parse` only once and we avoid the extra method. I'll make the change.

lib/puppet/provider/foreman_resource/rest_v3.rb
```diff
@@ -99,14 +99,16 @@ def success?(response)
...
-      '400' => 'Something is wrong with the data sent to Foreman server',
+      '400' => "Something is wrong with the data sent to #{fqdn}",
       '401' => 'Often this is caused by invalid Oauth credentials',
       '404' => 'The requested resource was not found',
-      '500' => 'Check /var/log/foreman/production.log on Foreman server for detailed information',
-      '502' => 'The webserver received an invalid response from the backend service. Was Foreman unable to handle the request?',
-      '503' => 'The webserver was unable to reach the backend service. Is foreman.service running?',
-      '504' => 'The webserver timed out waiting for a response from the backend service. Is Foreman under unusually heavy load?'
+      '500' => "Check /var/log/foreman/production.log on #{fqdn} for detailed information",
+      '502' => "The webserver received an invalid response from the backend service. Was foreman.service at #{fqdn} unable to handle the request?",
```
ekohl, line 109: I still think `Foreman` is more correct. Technically Foreman is delivered by running `foreman.socket` and `foreman.service`. Either could have a problem.

It's also clearly documented that Foreman is a service. https://access.redhat.com/documentation/en-us/red_hat_satellite/6.9/html/planning_for_red_hat_satellite/chap-red_hat_satellite-architecture_guide-introduction_to_red_hat_satellite#sect-Red_Hat_Satellite-Architecture_Guide-Red_Hat_Satellite_6_System_Components states:

>Foreman
>    Foreman is an open source application used for provisioning and life cycle management of physical and virtual systems. Foreman automatically configures these systems using various methods, including kickstart and Puppet modules. Foreman also provides historical data for reporting, auditing, and troubleshooting.

ehelms: I would tend to agree. I think the trouble only comes when we combine the term to be 'Foreman server'. If we stick to either Foreman, Foreman service or Foreman application then it works for both cases as it identifier the actual service and not the higher level concept of the entire box that 'Foreman server' or 'Satellite server' that can lead to branding debates.

wbclark:
Fixed unit tests.

There are still some failures for acceptance tests on Debian and Ubuntu, but to me they look unrelated to this change.

lib/puppet/provider/foreman_resource/rest_v3.rb
wbclark: I went with the language `Foreman at #{fqdn}` in my latest update. Please let me know if there's any more feedback on the wording... thanks!

lib/puppet/provider/foreman_resource/rest_v3.rb
```diff
@@ -99,14 +99,16 @@ def success?(response)
   end
 
   def error_message(response)
+    fqdn = URI::parse(resource[:base_url]).host
+
     explanations = {
-      '400' => 'Something is wrong with the data sent to Foreman server',
+      '400' => "Something is wrong with the data sent to #{fqdn}",
```
ehelms, line 105: Should this then also be `sent to Foreman at #{fqdn}` ?

wbclark: I think that would be both more specific and more consistent.

I will make the change

wbclark: I'll use this consistently throughout, i.e. also changing to `"Often this is caused by invalid Oauth credentials sent to Foreman at #{fqdn}"` and `"The requested resource was not found in Foreman at #{fqdn}"` for the explanations for 401 and 404 codes

lib/puppet/provider/foreman_resource/rest_v3.rb
```diff
@@ -99,14 +99,16 @@ def success?(response)
...
-      '401' => 'Often this is caused by invalid Oauth credentials',
-      '404' => 'The requested resource was not found',
-      '500' => 'Check /var/log/foreman/production.log on Foreman server for detailed information',
-      '502' => 'The webserver received an invalid response from the backend service. Was Foreman unable to handle the request?',
-      '503' => 'The webserver was unable to reach the backend service. Is foreman.service running?',
-      '504' => 'The webserver timed out waiting for a response from the backend service. Is Foreman under unusually heavy load?'
+      '400' => "Something is wrong with the data sent to Foreman at #{fqdn}",
+      '401' => "Often this is caused by invalid Oauth credentials sent to Foreman at #{fqdn}",
+      '404' => "The requested resource was not found in Foreman at #{fqdn}",
+      '500' => "Check /var/log/foreman/production.log on #{fqdn} for detailed information",
```
wbclark, line 108:
```suggestion
      '500' => "Check Foreman logs at /var/log/foreman/production.log on #{fqdn} for detailed information",
```

Considering making this change as well

ekohl: I would keep it short. IMHO the path tells you enough what's there.

ehelms (approved):

lib/puppet/provider/foreman_resource/rest_v3.rb
```diff
@@ -99,14 +99,16 @@ def success?(response)
...
     explanations = {
-      '400' => 'Something is wrong with the data sent to Foreman server',
-      '401' => 'Often this is caused by invalid Oauth credentials',
-      '404' => 'The requested resource was not found',
-      '500' => 'Check /var/log/foreman/production.log on Foreman server for detailed information',
-      '502' => 'The webserver received an invalid response from the backend service. Was Foreman unable to handle the request?',
-      '503' => 'The webserver was unable to reach the backend service. Is foreman.service running?',
-      '504' => 'The webserver timed out waiting for a response from the backend service. Is Foreman under unusually heavy load?'
+      '400' => "Something is wrong with the data sent to Foreman at #{fqdn}",
+      '401' => "Often this is caused by invalid Oauth credentials sent to Foreman at #{fqdn}",
```
ekohl, line 106: I'm not sure this change really adds anything. At the CLI we have oauth parameters. Those are wrong and the user needs to correct those. Does this comment help them find that out?

ekohl (approved):
I am a bit worried about how long these lines are getting. In Foreman we have unique error codes. I'm starting to get the feeling this may get useful here too.

lib/puppet/provider/foreman_resource/rest_v3.rb
wbclark: The change does clarify that the issue is the Oauth credentials that are being _sent_. So it could dispel the notion that the credentials are wrong somewhere on the Foreman side.

The intent is that the user will compare 1. what does Foreman expect to receive? vs. 2. What am I passing to the CLI?

lib/puppet/provider/foreman_resource/rest_v3.rb
wbclark: OK, I basically agree. While this would be more consistent with the overall pattern, it doesn't really inform the user of anything new and there doesn't justify the added length.

PR #995, opened 2021-09-17
Fixes #33511 - configure redis before dynflow workers

evgeni:
sometimes dynflow workers are started before redis, resulting in ugly
errors in the logs until redis is available

### DIFF
--- a/manifests/config.pp
+++ b/manifests/config.pp
@@ -12,6 +12,7 @@ class foreman::config {
     } else {
       include redis
       $dynflow_redis_url = "redis://localhost:${redis::port}/6"
+      Class['redis'] -> Service <| tag == 'foreman::dynflow::worker' |>
     }
 
     file { '/etc/foreman/dynflow':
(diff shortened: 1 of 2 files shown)


ekohl (commented):
I think this makes sense. Would be nice to have a test. For example here:
https://github.com/theforeman/puppet-foreman/blob/d5c3c9f03e2e9fa4952bfb70f27eacf0af23b796/spec/classes/foreman_spec.rb#L152-L153
I think adding `.that_requires('Service[redis]')` would work.

evgeni:
>  I think adding `.that_requires('Service[redis]')` would work.

Narrator: it did not.

evgeni:
Should it be `Class[redis]`, as we never touch `Service[redis]` in the order assignment?

ekohl:
The service name is dynamic so on EL7 it'll be different. Depending on the Class is good enough then.

ekohl (approved):

PR #996, opened 2021-09-21
Apply version restrictions to all packages

nbarrientos:
As there are packages installed by the module that escape from the version pin that can be set via `foreman::version`, the following situation might occur:

(Assuming `foreman::version` set to `2.4.0-1` and that `2.4.1-1` is the latest version available in the configured repositories)

* Puppet run `n`
  * Puppet installs `foreman-postgresql-2.4.0-1`.
  * Puppet installs `foreman-service` latest (due to `ensure => installed`)
    * DNF installs `foreman-service-2.4.1-1`
    * This triggers the update of `foreman` to `foreman-2.4.1-1` to match `foreman-service`'s requirements.
    * This triggers the update of `foreman-postgresql` to `2.4.1-1` to match `foreman-postgresql`'s dependency on `foreman`.
  * The Puppet run ends leaving the system in an undesired state.

* Puppet run `n+1`:
   * Puppet downgrades `foreman-postgresql` to `2.4.0-1`.
     * This triggers the downgrade of several packages, namely `foreman`, `foreman-service`, `foreman-dynflow-sideqik`, etc. This downgrade operation might or might not succeed :)

By ensuring that all packages are installed with the same version restrictions since the beginning, the unnecessary downgrade and the temporary bad system state explained above can be avoided.

(Internal note: applied locally as `027197f54e704fcab9cc5187193fe2e0009a7510`).

### DIFF
--- a/manifests/install.pp
+++ b/manifests/install.pp
@@ -19,12 +19,12 @@ class foreman::install {
   }
 
   package { 'foreman-service':
-    ensure => installed,
+    ensure => $foreman::version,
   }
 
   if $foreman::dynflow_manage_services {
     package { 'foreman-dynflow-sidekiq':
-      ensure => installed,
+      ensure => $foreman::version,
     }
   }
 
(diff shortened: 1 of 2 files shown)


manifests/install.pp
```diff
@@ -19,36 +19,36 @@
...
-      ensure => installed,
+      ensure => $foreman::version,
     }
   }
 
   if $foreman::ipa_authentication and $foreman::ipa_manage_sssd {
     package { 'sssd-dbus':
-      ensure => installed,
+      ensure => $foreman::version,
     }
```
ekohl, line 34: This package is not a subpackage of `foreman` so it has its own versioning.

ekohl (changes requested):
Makes sense, just one small inline comment.

ekohl (approved):
These are all subpackages generated from foreman.spec and the Debian equivalent so :+1: when the tests are green.

manifests/install.pp
nbarrientos: Thanks, Ewoud. Indeed, this is a silly search&replace mistake. It's fixed now :)

PR #1002, opened 2021-11-03
Add basic `foreman_hostgroup` type

alexjfisher:
Limitations:
* Very few properties supported.

Features:
* Supports creating and deleting hostgroups with or without parents
* Autorequires parent hostgroups
* Supports hostgroup (or parents) names with slashes in them [1]
* Supports organizations/locations
* Updates to properties are combined in `flush` method.
 
TODO:

- [x] Update Readme
- [x] Add some level of unit tests
- [x] Add an acceptance test
- [x] Refactor type parameters shared with other types

[1]
The following works...
```puppet
foreman_hostgroup { 'weird 1':
  name        => 'weir/d',
  description => 'A hostgroup with a slash in',
}

foreman_hostgroup { 'weird2':
  parent_hostgroup      => 'weir/d',
  parent_hostgroup_name => 'weir/d',
  description           => 'A child of a weird hostgroup',
}
```

### DIFF
--- a/lib/puppet/provider/foreman_hostgroup/rest_v3.rb
+++ b/lib/puppet/provider/foreman_hostgroup/rest_v3.rb
@@ -1,0 +1,182 @@
+# frozen_string_literal: true
+
+Puppet::Type.type(:foreman_hostgroup).provide(
+  :rest_v3,
+  parent: Puppet::Type.type(:foreman_resource).provider(:rest_v3)
+) do
+  confine feature: %i[json oauth]
+
+  def initialize(value = {})
+    super
+    @property_flush = {}
+  end
+
+  def exists?
+    !id.nil?
+  end
+
+  def create
+    Puppet.debug("Creating Foreman Hostgroup #{resource[:name]} with parent #{resource[:parent_hostgroup]}")
+
+    if resource[:parent_hostgroup] && parent_hostgroup_id.nil?
+      raise Puppet::Error,
+            "Parent hostgroup #{resource[:parent_hostgroup]} for #{resource[:name]} not found"
+    end
+
+    organization_ids = resource[:organizations]&.map { |org| organization_id(org) }
+    location_ids = resource[:locations]&.map { |loc| location_id(loc) }
+
+    post_data = {
+      hostgroup: {
+        name: resource[:name],
+        parent_id: parent_hostgroup_id,
+        description: resource[:description],
+        organization_ids: organization_ids,
+        location_ids: location_ids
+      }
+    }.to_json
+    path = 'api/v2/hostgroups'
+    r = request(:post, path, {}, post_data)
+
+    return if success?(r)
+
+    raise Puppet::Error, "Error making POST request to Foreman at #{request_uri(path)}: #{error_message(r)}"
+  end
+
+  def destroy
+    Puppet.debug("Destroying Foreman Hostgroup #{resource[:name]} with parent #{resource[:parent_hostgroup]}")
+    path = "api/v2/hostgroups/#{id}"
+    r = request(:delete, path)
+
+    unless success?(r)
+      error_string = "Error making DELETE request to Foreman at #{request_uri(path)}: #{error_message(r)}"
+      raise Puppet::Error, error_string
+    end
+
+    @hostgroup = nil
+  end
+
+  def flush
+    return if @property_flush.empty?
+
+    Puppet.debug "Calling API to update properties for #{resource[:name]}"
+
+    path = "api/v2/hostgroups/#{id}"
+    r = request(:put, path, {}, { hostgroup: @property_flush }.to_json)
+
+    return if success?(r)
+
+    raise Puppet::Error, "Error making PUT request to Foreman at #{request_uri(path)}: #{error_message(r)}"
+  end
+
+  # Property getters
+  def description
+    hostgroup ? hostgroup['description'] : nil
+  end
+
+  def organizations
+    hostgroup ? hostgroup['organizations'].map { |org| org['title'] } : nil
+  end
+
+  def locations
+    hostgroup ? hostgroup['locations'].map { |org| org['title'] } : nil
+  end
+
+  # Property setters
+  # If one of more properties is being modified then group all of these updates in @property_flush so that we can update them in a single API call in `flush()`
+  def description=(value)
+    @property_flush[:description] = value
+  end
+
+  def organizations=(value)
+    @property_flush[:organization_ids] = value.map { |org| organization_id(org) }
+  end
+
+  def locations=(value)
+    @property_flush[:location_ids] = value.map { |loc| location_id(loc) }
+  end
+
+  private
+
+  def hostgroup
+    @hostgroup ||= begin
+      path = 'api/v2/hostgroups'
+      search_name = resource[:name]
+      search_title = if resource[:parent_hostgroup]
+                       "#{resource[:parent_hostgroup]}/#{resource[:name]}"
+                     else
+                       search_name
+                     end
+      Puppet.debug("Searching for hostgroup with name #{search_name} and title #{search_title}")
+      r = request(:get, path, search: %(title="#{search_title}" and name="#{search_name}"))
+
+      raise Puppet::Error, "Error making GET request to Foreman at #{request_uri(path)}: #{error_message(r)}" unless success?(r)
+
+      results = JSON.parse(r.body)['results']
+      unless results.empty?
+        raise Puppet::Error, "Too many hostgroups found when looking for hostgroup with name #{search_name} and title #{search_title}" if results.size > 1
+
+        get_hostgroup_by_id(results[0]['id'])
+      end
+    end
+  end
+
+  def get_hostgroup_by_id(id)
+    path = "api/v2/hostgroups/#{id}"
+    r = request(:get, path)
+
+    raise Puppet::Error, "Error making GET request to Foreman at #{request_uri(path)}: #{error_message(r)}" unless success?(r)
+
+    JSON.parse(r.body)
+  end
+
+  def id
+    hostgroup ? hostgroup['id'] : nil
+  end
+
+  def parent_hostgroup
+    return nil unless resource[:parent_hostgroup]
+
+    @parent_hostgroup ||= begin
+      path = 'api/v2/hostgroups'
+      search_title = resource[:parent_hostgroup]
+      search_name = resource[:parent_hostgroup_name] || search_title.split('/').last
+      Puppet.debug("Searching for parent hostgroup with name #{search_name} and title #{search_title}")
+      r = request(:get, path, search: %(title="#{search_title}" and name="#{search_name}"))
+
+      raise Puppet::Error, "Error making GET request to Foreman at #{request_uri(path)}: #{error_message(r)}" unless success?(r)
+
+      results = JSON.parse(r.body)['results']
+      raise Puppet::Error, "Parent hostgroup #{resource[:parent_hostgroup]} for #{resource[:name]} not found" if results.empty?
+      raise Puppet::Error, "Too many hostgroups found when looking for parent hostgroup with name #{search_name} and title #{search_title}" if results.size > 1
+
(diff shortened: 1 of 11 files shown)


lib/puppet/type/foreman_hostgroup.rb
```diff
@@ -0,0 +1,99 @@
...
+      begin
+        value = Integer(value)
+      rescue ArgumentError
+        raise ArgumentError, 'The timeout must be a number.', $ERROR_INFO.backtrace
+      end
+      [value, 0].max
+    end
+
+    defaultto 500
+  end
```
ekohl, line 72: Can these be reused from https://github.com/theforeman/puppet-foreman/blob/master/lib/puppet_x/foreman/common.rb?

lib/puppet/provider/foreman_hostgroup/rest_v3.rb
```diff
@@ -0,0 +1,185 @@
...
+                       search_name
+                     end
+      Puppet.debug("Searching for hostgroup with name #{search_name} and title #{search_title}")
+      r = request(:get, path, search: %(title="#{search_title}" and name="#{search_name}"))
+
+      raise Puppet::Error, "Error making GET request to Foreman at #{request_uri(path)}: #{error_message(r)}" unless success?(r)
+
+      results = JSON.parse(r.body)['results']
+      unless results.empty?
+        raise Puppet::Error, "Too many hostgroups found when looking for hostgroup with name #{search_name} and title #{search_title}" if JSON.parse(r.body)['results'].size > 1
```
ekohl, line 30:
```suggestion
        raise Puppet::Error, "Too many hostgroups found when looking for hostgroup with name #{search_name} and title #{search_title}" if results.size > 1
```

lib/puppet_x/foreman/common.rb
```diff
@@ -49,7 +43,17 @@ module Common
         end
 
         autorequire(:anchor) do
-          ['foreman::service']
+          ['foreman::service','foreman::providers::oauth']
```
alexjfisher, line 46: @ekohl Not sure about this.  The recent types had the autorequire on the `foreman::service` anchor resource, but `foreman_smartproxy` was autorequiring `foreman::providers::oauth`.  They both exist.  Not sure which is correct or if including both is bad/good.

lib/puppet/type/foreman_hostgroup.rb
alexjfisher: Yes.  I've had a go at that.  let's see if the tests still pass...

lib/puppet/provider/foreman_hostgroup/rest_v3.rb
alexjfisher: Fixed.  Good catch.

lib/puppet_x/foreman/common.rb
```diff
@@ -1,16 +1,10 @@
...
-
-        newparam(:name, :namevar => true) do
-          desc 'The name of the resource.'
-        end
 
-        newparam(:hostname) do
-          desc 'The name of the host.'
-        end
+      # Parameters common to several types that use the rest_v3 api provider
+      FOREMAN_COMMON_PARAMS = Proc.new do
```
ekohl, line 6: Perhaps this could be named `REST_API_COMMON_PARAMS`?

lib/puppet/type/foreman_hostgroup.rb
```diff
@@ -0,0 +1,59 @@
...
+require_relative '../../puppet_x/foreman/common'
+Puppet::Type.newtype(:foreman_hostgroup) do
+  desc 'foreman_hostgroup manages hostgroups in foreman.'
+
+  instance_eval(&PuppetX::Foreman::Common::FOREMAN_COMMON_PARAMS)
+
+  def self.title_patterns
+    [
+      [
+        %r{^(.*)/(.*)$},
```
ekohl, line 12: Should this also use `.+` to disallow empty strings? Same below for the name.

spec/acceptance/foreman_hostgroups_spec.rb
```diff
@@ -0,0 +1,66 @@
...
+  end
+
+  it_behaves_like 'the foreman application'
+  it_behaves_like 'hammer'
+
+  describe command('hammer hostgroup list') do
+    its(:stdout) { is_expected.to include('example_hostgroup/child') }
+  end
+
+  describe 'removing a hostgroup' do
```
ekohl, line 40: Is the order defined in this way? I'm never sure with rspec.

lib/puppet/provider/foreman_hostgroup/rest_v3.rb
```diff
@@ -0,0 +1,185 @@
+# frozen_string_literal: true
+
+Puppet::Type.type(:foreman_hostgroup).provide(
+  :rest_v3,
+  parent: Puppet::Type.type(:foreman_resource).provider(:rest_v3)
+) do
+  confine feature: %i[json oauth]
+
+  def initialize(value = {})
+    super(value)
```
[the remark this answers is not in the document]
alexjfisher: Fixed

lib/puppet/provider/foreman_hostgroup/rest_v3.rb
```diff
@@ -0,0 +1,185 @@
...
+
+  def parent_hostgroup
+    return nil unless resource[:parent_hostgroup]
+
+    @parent_hostgroup ||= begin
+      path = 'api/v2/hostgroups'
+      search_title = resource[:parent_hostgroup]
+      search_name = if resource[:parent_hostgroup_name]
+                      resource[:parent_hostgroup_name]
+                    elsif %r{^(?<_parent>.*)/(?<name>.*)$} =~ resource[:parent_hostgroup]
```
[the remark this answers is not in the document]
alexjfisher: Changed to `.+`

lib/puppet/provider/foreman_hostgroup/rest_v3.rb
```diff
@@ -0,0 +1,185 @@
...
+                      name
```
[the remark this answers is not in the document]
alexjfisher: From https://medium.com/rubycademy/3-ways-to-use-regexp-capture-groups-with-back-references-in-ruby-b4969cc9b3ec
> The named capture feature, when it’s used with the =~ operator, has a powerful mechanism that automatically assigns the result of a matching group into a local variable using the name assigned to the group into the regular expression

lib/puppet/provider/foreman_hostgroup/rest_v3.rb
```diff
@@ -0,0 +1,185 @@
...
+  def title_to_id(type, title)
+    path = "api/v2/#{type}"
+    r = request(:get, path, search: %(title="#{title}"))
+    unless success?(r)
+      raise Puppet::Error,
+            "Error making GET request to Foreman at #{request_uri(path)}: #{error_message(r)}"
+    end
+    raise Puppet::Error, "#{title} not found in #{type}" unless JSON.parse(r.body)['results'].size == 1
+
+    JSON.parse(r.body)['results'][0]['id']
```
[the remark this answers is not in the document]
alexjfisher: Done.

lib/puppet_x/foreman/common.rb
alexjfisher: Done.

lib/puppet/type/foreman_hostgroup.rb
alexjfisher: Fixed.

spec/acceptance/foreman_hostgroups_spec.rb
alexjfisher: I think so.  Test runs look correct.

lib/puppet/provider/foreman_hostgroup/rest_v3.rb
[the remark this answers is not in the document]
ekohl: TIL. Perhaps a comment can help there. It may not be obvious to everyone. Another less magical way may be:
```ruby
elsif resource[:parent_hostgroup].include?('/')
  resource[:parent_hostgroup].split('/').last
```

Actually, that would also work if there is no `/` and since `search_title` is `resource[:parent_hostgroup]` it could be simplified to just
```ruby
else
  search_title.split('/').last
end
```

Which can then mean you can reduce it to:
```ruby
search_name = resource[:parent_hostgroup_name] || search_title.split('/').last
```

lib/puppet/provider/foreman_hostgroup/rest_v3.rb
```diff
@@ -0,0 +1,186 @@
...
+    r = request(:get, path, search: %(title="#{title}"))
+    unless success?(r)
+      raise Puppet::Error,
+            "Error making GET request to Foreman at #{request_uri(path)}: #{error_message(r)}"
+    end
+    results = JSON.parse(r.body)['results']
+    raise Puppet::Error, "#{title} not found in #{type}" unless results.size == 1
+
+    results[0]['id']
+  end
```
ekohl, line 104: Not that important, but I wonder if these could be in some `private` part to indicate what's public API and what are helpers.

spec/acceptance/foreman_hostgroups_spec.rb
ekohl: I've used https://relishapp.com/rspec/rspec-core/docs/configuration/overriding-global-ordering a few times in the past. May not be a bad thing to be explicit here.

lib/puppet_x/foreman/common.rb
ekohl: I think both is fine.

lib/puppet/provider/foreman_hostgroup/rest_v3.rb
```diff
@@ -0,0 +1,182 @@
...
+  def create
+    Puppet.debug("Creating Foreman Hostgroup #{resource[:name]} with parent #{resource[:parent_hostgroup]}")
+
+    if resource[:parent_hostgroup] && parent_hostgroup_id.nil?
+      raise Puppet::Error,
+            "Parent hostgroup #{resource[:parent_hostgroup]} for #{resource[:name]} not found"
+    end
+
+    organization_ids = (resource[:organizations]&.map { |org| organization_id(org) })
+    location_ids = (resource[:locations]&.map { |loc| location_id(loc) })
```
ekohl, line 27: Technically the braces aren't needed
```suggestion
    organization_ids = resource[:organizations]&.map { |org| organization_id(org) }
    location_ids = resource[:locations]&.map { |loc| location_id(loc) }
```

lib/puppet/provider/foreman_hostgroup/rest_v3.rb
```diff
@@ -0,0 +1,182 @@
...
+    end
+  end
+
+  def get_hostgroup_by_id(id)
+    path = "api/v2/hostgroups/#{id}"
+    r = request(:get, path)
+
+    return JSON.parse(r.body) if success?(r)
+
+    raise Puppet::Error, "Error making GET request to Foreman at #{request_uri(path)}: #{error_message(r)}"
```
ekohl, line 130: I'm generally not a fan of this pattern. It'll work, but if you at some point need to extend it you'll need to rewrite it. So I prefer to first raise and then "return".
```suggestion
    unless success?(r)
      raise Puppet::Error, "Error making GET request to Foreman at #{request_uri(path)}: #{error_message(r)}"
    end

    JSON.parse(r.body)
```

It's a bit longer, but my reasoning is also that methods should always reach the end unless something abnormally happens.

ekohl (commented):
This looks good to me. Just some very minor things inline. Feel free to squash it and move it out of draft.

lib/puppet/provider/foreman_hostgroup/rest_v3.rb
alexjfisher: Brackets removed

alexjfisher:
@ekohl Updated for your last 2 comments and all squashed.  Should be ready to merge?

ekohl (approved):
Thanks!

PR #1007, opened 2021-11-19
add outofsync_interval parameter support

fe80:
Hello,

Just a little patch for support the outofsync_interval param. Tested and validate with forman 3.0.1

### DIFF
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -174,6 +174,8 @@
 #
 # $register_in_foreman::          Register host in Foreman
 #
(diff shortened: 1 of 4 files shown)


ekohl:
I'm a bit hesitant to add this. We may end up with a lot of parameters. Have you considered using `foreman_config_entry` or creating your own `concat::fragment`? For the latter we could also add a defined type to provide a proper API: I've opened https://github.com/theforeman/puppet-foreman/pull/1016.

fe80:
Yes I agree,

I've an other params to add...

PR #1010, opened 2021-12-16
Fixes #34161 - Run apipie:cache:index after db:migrate

ekohl:
With Puppet you can write:

```puppet
Class['A'] ~> Class['B'] ~> Class['C']
```

This implies `Class['A'] ~> Class['C']` and rspec-puppet actually will tell you that chaining is there, but it doesn't actually happen. This was reported in https://github.com/rodjek/rspec-puppet/pull/821.

In this particular case it means that if DB seeding doesn't happen then the apipie caches indexes aren't refreshed. By chaining it to db:migrate there's a much bigger chance it actually happens.

Normally apipie:cache:index needs to run after a package is updated so ideally this would actually be done in packaging, but this is the workaround we've been using for a long time.

### DIFF
--- a/manifests/database.pp
+++ b/manifests/database.pp
@@ -27,6 +27,7 @@ class foreman::database(
     foreman::rake { 'db:migrate':
       timeout => $timeout,
       unless  => '/usr/sbin/foreman-rake db:abort_if_pending_migrations',
+      notify  => Foreman::Rake['apipie:cache:index', 'apipie_dsl:cache'],
     }
     ~> foreman_config_entry { 'db_pending_seed':
       value => false,
@@ -34,7 +35,7 @@ class foreman::database(
     }
     ~> foreman::rake { 'db:seed':
       environment => delete_undef_values($seed_env),
+      notify      => Foreman::Rake['apipie:cache:index', 'apipie_dsl:cache'],
     }
-    ~> Foreman::Rake['apipie:cache:index', 'apipie_dsl:cache']
   }
 }


ekohl:
@evgeni the theory why this should be the fix is in the description but I haven't tested this.

evgeni:
This matches my memory.

Do you want me to run this past my reproducer, or should we make the nightly pipe be that test?

evgeni:
Hmmm… I've applied the patch in my environment, and while I now *do* see an explicit relationship between `db:seed` and `apipie:cache:index`, it's not being executed:

```
[DEBUG ] [configure] /Stage[main]/Foreman::Database/Foreman::Rake[db:migrate]/notify: notify to Foreman::Rake[apipie:cache:index]
[DEBUG ] [configure] /Stage[main]/Foreman/Foreman::Rake[apipie:cache:index]/Exec[foreman-rake-apipie:cache:index]: Starting to evaluate the resource (1305 of 1908)
[DEBUG ] [configure] /Stage[main]/Foreman/Foreman::Rake[apipie:cache:index]/Exec[foreman-rake-apipie:cache:index]: '/usr/sbin/foreman-rake apipie:cache:index' won't be executed because of failed check 'refreshonly'
[DEBUG ] [configure] /Stage[main]/Foreman/Foreman::Rake[apipie:cache:index]/Exec[foreman-rake-apipie:cache:index]: Evaluated in 0.00 seconds
```

evgeni (approved):

PR #1011, opened 2021-12-17
Fixes #34089 - Add trusted proxies setting

sbernhard:


sbernhard:
I don't think the failed tests are because of this PR. Am I right?

manifests/init.pp
```diff
@@ -160,6 +160,8 @@
 #
 # $cors_domains::                 List of domains that show be allowed for Cross-Origin Resource Sharing
 #
+# $trusted_proxies::              List of trusted proxies (at least every smart-proxy)
```
ehelms, line 163: Could this be named to include that they are http(s) proxies? I worry about confusion with the ole smart-proxy which is what I thought this was aimed at at first.

sbernhard: It fixes: https://community.theforeman.org/t/using-cloud-init-templates-with-foreman-proxy-foreman-proxy-content/20269/20

Should I change the description or the key?

ekohl: I would say the description because `$trusted_proxies` matches the setting in the settings file.

templates/settings.yaml.erb
```diff
@@ -42,8 +42,16 @@
 <% scope.lookupvar('foreman::cors_domains').each do |domain| -%>
  - '<%= domain %>'
 <% end -%>
+<% end -%>
```
ekohl, line 46: Please keep the old style line endings and also apply it to `trusted_proxies`. This helps avoid empty lines. with your change here you end up with:
```yaml
:logging:
  # ...


 # Individual logging types can be toggled on/off here
:loggers:
```

The previous code left out that empty line in the resulting file. That's why the empty line was within the `unless` block.

ekohl (changes requested):

templates/settings.yaml.erb
sbernhard: hopefully closed.

manifests/init.pp
ekohl: I still consider this unresolved. This is ambiguous because there's multiple proxies. This should explicitly say that these are reverse HTTP proxies.

The Smart Proxy *can* act as a reverse HTTP proxy, but only if you use some modules. Templates is one example here (and probably the most important).

In the Katello case, there is also a reverse HTTP proxy for RHSM and there it may also be important for correctly logging the remote IP. In that case Apache acts as a reverse proxy.

sbernhard: Ok. If you have something in mind which I should write just let me know. Any suggestions is welcome 😁

sbernhard: Changed the description again.

manifests/init.pp
```diff
@@ -160,6 +160,8 @@
 #
 # $cors_domains::                 List of domains that show be allowed for Cross-Origin Resource Sharing
 #
+# $trusted_proxies::              List of trusted HTTP(s) reverse proxies, which are allowed to access.
```
ehelms, line 163: To access... what? I suppose the implication is the Foreman puma application? But I would opt to tweak the wording to make this clearer.

sbernhard: which are allowed to access the API?

manifests/init.pp
```diff
@@ -160,6 +160,9 @@
 #
 # $cors_domains::                 List of domains that show be allowed for Cross-Origin Resource Sharing
 #
+# $trusted_proxies::              List of trusted IPs / networks. Default: localhost and all private IP address ranges.
```
ekohl, line 163: This is incorrect because Foreman overrides this to only localhost on v4 and v6:
https://github.com/theforeman/foreman/blob/a6471d0ac59ce83f6fa765da149e367aa6d459ec/config/application.rb#L200-L208

Also note that the way it works is that you must list **all** trusted IPs. That also means localhost if you override it.

ekohl (changes requested):

ekohl (approved):
I'm ok with the phrasing now. @ehelms?

ehelms (approved):

evgeni:
I think this [broke our builds](http://koji.katello.org/kojifiles/work/tasks/3417/473417/build.log), because [`Stdlib::IP::Address` is defined as](https://github.com/puppetlabs/puppetlabs-stdlib/blob/v7.1.0/types/ip/address.pp):
```ruby
  type Stdlib::IP::Address = Variant[
    Stdlib::IP::Address::V4,
    Stdlib::IP::Address::V6,
  ]
```

And Kafo can't parse multi-line strings properly. The resulting error message is:
```
D, [2022-01-04T03:15:37.585013 #6846] DEBUG -- : Using Puppet module parser KafoParsers::PuppetStringsModuleParser
F, [2022-01-04T03:15:38.845230 #6846] FATAL -- : Unable to continue because of: data type not recognized Variant[
rake aborted!
```

ekohl:
I forgot to link back https://community.theforeman.org/t/foreman-installer-develop-package-release-1015-failed/26627/3 and https://github.com/theforeman/puppet-foreman/pull/1013 but here it is.

PR #1016, opened 2022-01-10
Introduce foreman::settings_fragment

ekohl:
This provides a stable API to provide fragments to settings.yaml. This means we don't need to create parameters for every single setting.

### DIFF
--- a/spec/defines/foreman_settings_fragment_spec.rb
+++ b/spec/defines/foreman_settings_fragment_spec.rb
@@ -1,0 +1,23 @@
+require 'spec_helper'
+
+describe 'foreman::settings_fragment' do
+  let(:title) { 'myfragment' }
+
+  context 'with string' do
+    let(:params) do
+      {
+        content: 'mycontent',
+        order: '42',
+      }
+    end
+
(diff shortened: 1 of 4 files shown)


fe80:
It's possible to create a create_resource for manage that with hiera ? And maybe a validation with a custom type (or this is too much)?

I can look that in different pull request

ekohl:
I'd prefer that in a separate PR. I can't find the bug right now, but note we have some issues in our installer parsing types that span multiple lines. That is likely to show up there.

spec/defines/foreman_settings_fragment_spec.rb
```diff
@@ -0,0 +1,21 @@
...
+  let(:title) { 'myfragment' }
+
+  context 'with string' do
+    let(:params) do
+      {
+        content: 'mycontent',
+        order: '42',
+      }
+    end
+
```
bastelfreak, line 13: should we add a:
```
it { is_expected.to compile.with_all_deps }
```
?

evgeni (approved):

PR #1018, opened 2022-01-20
Pass potentially Sensitive params as Sensitive

ekohl:
In 699f944c93b8cf3718b16e6b9de7143c1e55bbe8 the parameters started to accept Sensitive but it didn't default to Sensitive. They also weren't converting data coming from Hiera. This adds a data-in-modules setup and sets lookup_options for those. This means Kafo (which heavily relies on Hiera) will pass sensitive values.

Currently a draft so I can verify it end to end with the installer.


ekohl:
Oh, this is an interesting one:
```
2022-01-20 12:16:54 [ERROR ] [configure] Evaluation Error: Error while evaluating a Method call, Class[Foreman::Cli]: parameter 'password' expects a value of type Undef, String, or Sensitive[String], got Sensitive[Undef] (file: /usr/share/gems/gems/kafo-6.4.0/modules/kafo_configure/manifests/init.pp, line: 16, column: 39) on node centos8.wisse.example.com
```

Because it converts it to `Sensitive` unconditionally, it doesn't work.

ekohl:
After changing the data type this doesn't appear to work?
```
2022-01-20 12:24:09 [ERROR ] [configure] Evaluation Error: Error while evaluating a Function Call, lambda: parameter 'password' expects a value of type Undef, String, or Sensitive[String], got Sensitive[Undef] (file: /usr/share/foreman-installer/modules/foreman/manifests/cli.pp, line: 92, column: 18) on node centos8.wisse.example.com
```

ekohl:
@cocker-cc this isn't pretty, but I couldn't get it to work otherwise. Please have a look.

cocker-cc:
Hiera itself will 

- either find the Password in YAML – then it will convert it to `Sensitive[String]` according to `lookup_options`
- or it will not find the Password, then the Lookup will produce an Error

Hiera will **not** return `undef` (and neither will return `Sensitive[undef]` as a Consequence).

This means IMHO: If the EPP receives `Sensitive[undef]`, then the Cast to `Sensitive` has to happen in Puppetcode **after** the Hiera-Lookup.

ekohl:
That does not reflect my findings in our installer. That is because our installer copies all answers and persists them. This results in `Sensitive[Undef]`. Hence why I chose these data types.

I also couldn't get https://puppet.com/docs/puppet/6/securing-sensitive-data.html#epp-templates to work. Using `scope` in epp resulted in:
```
Evaluation Error: A substring operation does not accept a String as a character index. Expected an Integer (file: /home/ekohl/dev/puppet-foreman/spec/fixtures/modules/foreman/templates/hammer_root.yml.epp
```

cocker-cc:
My Thoughts on this – without having had a deep Insight into the Code:

[Here](https://github.com/theforeman/puppet-foreman/commit/120efb3f3e80dfbcc6cfdc5f4e1af60e0bf59617) as an Example-Commit I see:
```
Variant[Optional[String], Sensitive[Optional[String]]] $password = $foreman::cli::params::password,
```
`Sensitive[Optional[Any]]` does not make Sense, because noone wants to accept `Sensitive[undef]` to pass in. (`Optional` allows `undef`)
Instead it should read like
```
Optional[Variant[String, Sensitive[String]]] $password
```

Furthermore:
```
$db_password = Sensitive(extlib::cache_data('foreman_cache_data', 'db_password', extlib::random_password(32)))
```
Eventually `extlib::cache_data()` is returning `undef`?  In this Case you would get `Sensitive(undef)` – exactly, what your Error-Message complains about.  
So instead it should read something like
```
$_cached_pw = extlib::cache_data('foreman_cache_data', 'db_password', extlib::random_password(32))
$db_password = $_cached_pw ? {
  undef   => undef,
  default => Sensitive($_cached_pw),
}
```

The other Variables are accordingly.

ekohl:
For now I want to revert it: https://github.com/theforeman/puppet-foreman/pull/1019. This week we're planning to do module releases for Foreman 3.2 and it needs to be stable. Reverting it now is the quickest path. It can be added in the release candidate phase. Once it's reverted, I'll update this PR to make a complete implementation.

jhoblitt:
What about a function that converts `Sensitive[undef]` to simply `undef`. It would be ugly to have to use such a function pretty much everywhere but it might work.

PR #1022, opened 2022-01-26
Fixes #34317 - Use the correct certificate to register

ekohl:
The server SSL CA is used to authenticate client certificates. The chain is actually the one that has signed the public key of the certificate.


evgeni (approved):

evgeni:
this fixes the issue in my reproduced

ekohl:
It looks closed, but that's because I accidentally pushed the rebased commit first to master and then to this branch. It's really merged.

PR #1023, opened 2022-01-26
Explicitly enable the foreman dnf module on Foreman 3.2+

evgeni:


manifests/repo.pp
```diff
@@ -41,6 +41,14 @@
         provider    => 'dnfmodule',
       }
     }
+
+    if $facts['os']['release']['major'] == '8' and ($repo == 'nightly' or versioncmp($repo, '3.2') >= 0) {
+      package { 'foreman':
+        ensure      => 'latest',
```
evgeni, line 47: maybe "latest" wasn't that a good name after all? :D

manifests/repo.pp
```diff
@@ -41,6 +41,14 @@
         provider    => 'dnfmodule',
       }
     }
+
+    if $facts['os']['release']['major'] == '8' and ($repo == 'nightly' or versioncmp($repo, '3.2') >= 0) {
```
ekohl, line 45: With this, is it still needed to enable Ruby? Or should this be first and then the `>= 2.5` as an `elsif`.

evgeni: yea, the >=2.5 can be an eslif

manifests/repo.pp
```diff
@@ -34,7 +34,13 @@
       Foreman::Repos['foreman'] -> Package['centos-release-scl-rh']
     }
 
-    if $facts['os']['release']['major'] == '8' and ($repo == 'nightly' or versioncmp($repo, '2.5') >= 0) {
+    if $facts['os']['release']['major'] == '8' and ($repo == 'nightly' or versioncmp($repo, '3.2') >= 0) {
+      package { 'foreman:latest':
```
evgeni, line 38: yes, this essentially breaks the fix from #1015, but I need to see if it works before we rename the stream on `yum.tfm`

evgeni:
this should not be merged before we didn't get a decision what we rename `latest` to in https://github.com/theforeman/foreman-packaging/pull/7522

manifests/repo.pp
```diff
@@ -34,7 +34,14 @@
       Foreman::Repos['foreman'] -> Package['centos-release-scl-rh']
     }
 
-    if $facts['os']['release']['major'] == '8' and ($repo == 'nightly' or versioncmp($repo, '2.5') >= 0) {
+    if $facts['os']['release']['major'] == '8' and ($repo == 'nightly' or versioncmp($repo, '3.2') >= 0) {
+      package { 'foreman':
```
evgeni, line 38: funny how we nowhere have a `package { 'foreman' }` in other places of the code…

ekohl: That also surprised me.

evgeni:
of course, this now also fails to install hammer_cli_katello, because the katello module ain't enabled.

ekohl (approved):
I think we'll also need a similar change in [katello::repo](https://github.com/theforeman/puppet-katello/blob/master/manifests/repo.pp).

PR #1025, opened 2022-01-27
Reflect Foreman 3.2+ support for Debian 11

ekohl:


evgeni (approved):

evgeni:
Why is EL8 failing?

ekohl:
> Why is EL8 failing?

It looks like `host_inventory['fqdn']` is `nil`. I wonder if that's because I just released the switch from CentOS Linux 8 to CentOS Stream 8 and something in the tooling can no longer fetch the FQDN. I'll have a look.

ekohl:
If I read this right, it should run `hostname -f`:
https://github.com/mizzy/specinfra/blob/5eb0940fa837445aa51160884c5951ca594206fa/lib/specinfra/command/linux/base/inventory.rb#L19-L21
Perhaps that isn't properly working in the Stream 8 container?

ekohl:
I'll fix that in a separate PR.

ekohl:
> I'll fix that in a separate PR.

https://github.com/voxpupuli/beaker-hostgenerator/pull/242 should be that PR. looks like it doesn't install `hostname` in the minimal container and specinfra relies on it.

PR #1028, opened 2022-02-10
Refs #34394 - trigger dynflow restart when DB restarts

evgeni:
When PostgreSQL is restarted (or worse, unavailable for a while),
Dynflow doesn't always cope well with this fact and sometimes ends in a
bad state, not able to recover itself.

This is clearly not fixing the underlying Dynflow issue, but at least
lets restart Dynflow in the cases we *know* the DB dropped connections.


manifests/dynflow/worker.pp
```diff
@@ -51,6 +51,8 @@
       enable    => true,
       subscribe => Class['foreman::database'],
     }
+
+    Anchor <| title == 'postgresql::server::service::end' |> ~> Service[$service]
```
ekohl, line 55: The chaining certainly makes sense but I'm not sure this would really notify since I don't see a `Service[postgresqld] ~> Anchor[postgresql::server::service::end]`. Have you verified this?

And spec testing here is unreliable (we've seen that in the past in https://github.com/rodjek/rspec-puppet/pull/821), but you could try to see what this test shows. If it fails, you know for certain it isn't there ;)
```ruby
it { is_expected.to contain_service('postgresqld').that_notifies('Service[dynflow-sidekiq@SOMETHING]') }
```

ekohl: Something something theory and practice ...

What may happen is that something sends a notification to `Class[Postgresql::Serverver::Service]` so all resources within the class are refreshed. I wonder what happens if you run:
```
foreman-installer
systemctl stop postgresql
foreman-installer
```

Does it then also restart Dynflow?

evgeni: Yes, that validates with both rspec and the real world, nice.

ekohl (approved):
I'm ok with not introducing the strict chaining and just sticking with this.

markuslanz77:
@evgeni  @ekohl   
Unfortunately this fix is not applicable for installations that don't use the local PostgreSQL instance. 
see: https://projects.theforeman.org/issues/34485

am i missing something here?

adamruzicka:
Not really, https://projects.theforeman.org/issues/34485 should be the thing that covers it. Let's continue in there

PR #1032, opened 2022-03-09
Switch to gha-puppet for CI

ekohl:


ekohl:
PostgreSQL not starting on Ubuntu 20.04 is something we've seen for a while. I can't reproduce it in a regular VM so it is likely something that has to do with the container. While we should fix it, I think it shouldn't block this PR. It does mean we can't enable branch protection just yet (which means no auto-merge), but it's not something we had before this either.

ehelms (approved):

PR #1037, opened 2022-03-17
Fixes #32323 - ensure systemd override loads for SCL PostgreSQL

evgeni:


manifests/database/postgresql.pp
```diff
@@ -24,4 +24,11 @@
     encoding => 'utf8',
     locale   => 'en_US.utf8',
   }
+
+  if $facts['os']['family'] == 'RedHat' and $facts['os']['release']['major'] == '7' {
+    file { '/etc/systemd/system/rh-postgresql12-postgresql.service.d':
```
evgeni, line 29: To take effect, this needs to trigger a `daemon-reload` once, but not sure how to do this from here?

ehelms: Notify the service?

ekohl:
My biggest concern with this is that it doesn't really map to the correct service. For example, we also have puppet-pulpcore and puppet-candlepin which pull in PostgreSQL. That means we'll keep the warning on content proxies which only run puppet-pulpcore.

I wonder if the real fix isn't to modify this:
https://github.com/theforeman/foreman-installer/blob/ae449b051d6e3648bb6dcb34e12f2bd7bc7bf6f8/config/foreman.hiera/family/RedHat-7.yaml#L8
To use rh-postgresql12-postgresql (which is the real name). Then we probably still need to clean the old override somewhere, but I believe it should work. We should verify it still works with upgrades etc.

ekohl:
I opened https://github.com/theforeman/foreman-installer/pull/756 with what I mean, but I haven't verified it yet.

evgeni:
Ewouds solution ended up being more correct!

PR #1041, opened 2022-04-20
Pre-install foreman-selinux in acceptance environment

wbclark:
Also do not purge foreman-selinux between tests. This is a workaround
for the Puppet issue [PUP-10548].


spec/setup_acceptance_node.pp
```diff
@@ -3,11 +3,18 @@
...
 # Needed for idempotency when SELinux is enabled
-if $foreman::repo::configure_scl_repo {
-  package { 'rh-redis5-redis':
-    ensure  => installed,
+if $facts['os']['selinux']['enabled'] {
+  package { 'foreman-selinux':
+    ensure  => latest,
     require => Class['foreman::repo'],
   }
+
```
ehelms, line 11:
```suggestion
}
```

spec/setup_acceptance_node.pp
```diff
@@ -3,11 +3,18 @@
...
+  if $foreman::repo::configure_scl_repo {
```
ehelms, line 12:
```suggestion
if $foreman::repo::configure_scl_repo {
```

spec/setup_acceptance_node.pp
```diff
@@ -3,11 +3,18 @@
...
+    package { 'rh-redis5-redis':
+      ensure  => installed,
+      require => Class['foreman::repo'],
+    }
+  }
```
ehelms, line 17: This whole block is indented too far I think. And you'll have a trailing `}` on line 18.

wbclark: It was my intention to have the `rh-redis5-redis` package inside `if $facts['os']['selinux']['enabled'] { ... }` because of the existing comment on L#5: `# Needed for idempotency when SELinux is enabled` (which now conveniently applies to both package resources)

IMO it was an oversight that this wasn't already conditioned on selinux being enabled.

I thought about making the change in a separate commit but it seemed too minor to warrant it. That would have made this easier to review but probably the git log clutter is not worth it when this only affects tests.

What do you think?

ehelms: That works. Can you rebase on latest master branch to ensure tests pass?

wbclark:
Hey @ehelms , I triggered the tests to run again. The failures on Ubuntu look non-related:

```
  Warning: PostgresqlValidator.attempt_connection: Sleeping for 1 seconds
  Warning: PostgresqlValidator.attempt_connection: Sleeping for 1 seconds
  Warning: PostgresqlValidator.attempt_connection: Sleeping for 1 seconds
  Error: Unable to connect to PostgreSQL server! (:5432)
  Error: /Stage[main]/Postgresql::Server::Service/Postgresql_conn_validator[validate_service_is_running]/ensure: change from 'absent' to 'present' failed: Unable to connect to PostgreSQL server! (:5432)
  ```

**Comment by evgeni** [comment]
Yeah, ubuntu 20.04 tests are busted.

File: spec/support/acceptance/purge.rb [hunk 4]
```diff
@@ -1,7 +1,7 @@
 def purge_foreman
   case fact('osfamily')
   when 'RedHat'
-    on default, 'yum -y remove foreman* tfm-*'
```
ekohl, line 4: Isn't there a way to exclude a package in yum/dnf remove?

ekohl (changes requested):
Needs a rebase now.

PR #1043, opened 2022-04-22
Add $selinux_ignore_defaults parameter with default value true

wbclark:
The idea is that `foreman::selinux_ignore_defaults: true` can be used for initial install so that the system will determine security context for files at the time they are created, and `foreman::selinux_ignore_defaults: false` can be used on subsequent runs to remediate drift from the already installed policy.

This allows foreman-installer to eliminate hooks/pre/32-install_selinux_packages.rb.

The one case this doesn't adequately cover would be the case where `foreman-selinux` package is updated by this module with changes to the policy. This would be solved for foreman_maintain users by having foreman_maintain run the installer with `foreman::selinux_ignore_defaults: false` since packages would always be updated prior to the installer running.

### DIFF
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -190,6 +190,8 @@
 #
 # $dynflow_redis_url::            If set, the redis server is not managed and we use the defined url to connect
 #
+# $selinux_ignore_defaults::      Do not lookup default security context for file resources in catalogue compilation and attempt to manage them; instead defer context lookups to the system itself when the files are actually created. Useful during initial installs, because Puppet can install packages which modify the security policy after the context lookups were performed, which breaks idempotence. This can be disabled after the initial install, to allow Puppet to remedy drift in security context.
+#
 # === Keycloak parameters:
 #
 # $keycloak::                     Enable Keycloak support. Note this is limited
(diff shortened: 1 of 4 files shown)


manifests/init.pp
```diff
@@ -313,6 +316,12 @@
   include foreman::database
   contain foreman::service
 
+  if $facts['os']['selinux']['enabled'] {
+    File {
+      selinux_ignore_defaults => $foreman::selinux_ignore_defaults,
+    }
+  }
```
ekohl, line 323: I'm against this. It's way too broad and magic. Pretty sure this affects the whole catalog, not just the files created by this module.

Instead, can we set the property everywhere? It's more verbose, but also easier to follow.

spec/setup_acceptance_node.pp
```diff
@@ -2,14 +2,6 @@
   repo => 'nightly',
 }
 
-# Needed for idempotency when SELinux is enabled
-if $foreman::repo::configure_scl_repo {
-  package { 'rh-redis5-redis':
-    ensure  => installed,
-    require => Class['foreman::repo'],
-  }
-}
```
ekohl, line 11: We've dropped EL7 so this should no longer be needed.

ekohl (commented):
I wonder about this approach. On the one hand, it does allow an initial install to pass. On the other, it can't rectify labels if they are incorrect.

A further enhancement to this could be to introduce a custom fact that checks if `foreman-selinux` is installed. Something like `foreman_selinux_installed`. Then you use `selinux_ignore_defaults => !facts['foreman_selinux_installed']`. In the initial run it'll avoid setting labels while in the future it rectifies contexts on managed files.

Thoughts on this?

PR #1046, opened 2022-04-29
Fixes #34602 - restart services after plugin installation

evgeni:


manifests/init.pp
```diff
@@ -343,4 +343,10 @@
 
   contain 'foreman::settings' # lint:ignore:relative_classname_inclusion (PUP-1597)
   Class['foreman::database'] -> Class['foreman::settings']
+
+  file { '/usr/share/foreman/tmp/updated_plugins':
```
evgeni, line 347: for Debian, this file is created via https://github.com/theforeman/foreman-packaging/pull/7853
for RPM, this needs implementatin

manifests/init.pp
```diff
@@ -343,4 +343,10 @@
 
   contain 'foreman::settings' # lint:ignore:relative_classname_inclusion (PUP-1597)
   Class['foreman::database'] -> Class['foreman::settings']
+
+  file { '/usr/share/foreman/tmp/restart_required_changed_plugins':
+    ensure  => absent,
+    notify  => Class['foreman::service'],
```
jhoblitt, line 349: I recommend adding rspec coverage for this.  Notifications seem to frequently become accidentally broken during a refactor.

evgeni: You're right!

Updated.

ehelms (approved):

PR #1049, opened 2022-05-05
Fixes #34824 - properly restart foreman when puma config changed

evgeni:
we need to restart foreman.service *before* a (possible) restart of
foreman.socket, as the later *also* does restart foreman.service which
leads to foreman.socket being *started* instead of *restarted*

    /Service[foreman.socket]: Starting to evaluate the resource (2265 of 2522)
    Executing: '/bin/systemctl is-active -- foreman.socket'
    Executing: '/bin/systemctl restart -- foreman.socket'
    /Service[foreman]: Starting to evaluate the resource (2266 of 2522)
    Executing: '/bin/systemctl is-active -- foreman'
    Executing: '/bin/systemctl show --property=NeedDaemonReload -- foreman'
    Executing: '/bin/systemctl daemon-reload'
    Executing: '/bin/systemctl unmask -- foreman'
    Executing: '/bin/systemctl start -- foreman'
    Executing: '/bin/systemctl is-enabled -- foreman'
    /Stage[main]/Foreman::Service/Service[foreman]/ensure: ensure changed 'stopped' to 'running'

But in this case the now running foreman.service didn't see the changes
to the service file that daemon-reload would have loaded.

(cherry picked from commit 6b7b05e0dbfdc26623422a9b38979524ac9d29d9)

### DIFF
--- a/manifests/repos/apt.pp
+++ b/manifests/repos/apt.pp
@@ -2,7 +2,7 @@
 # @api private
 define foreman::repos::apt (
   Variant[Enum['nightly'], Pattern['^\d+\.\d+$']] $repo,
-  String $key = 'AE0AF310E2EA96B6B6F4BD726F8600B9563278F6',
+  String $key = '5B7C3E5A735BCB4D615829DC0BDDA991FD7AAC8A',
   Stdlib::HTTPUrl $key_location = 'https://deb.theforeman.org/foreman.asc',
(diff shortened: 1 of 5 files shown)


evgeni:
this is a pick for 3.1/3.2, which I don't want to push to the new 20.x, but still fix that issue.

ehelms (approved):

PR #1053, opened 2022-05-19
derive base_url from foreman-proxy/settings.yml by default

jhoblitt:
Setting the base_url param on all foreman_* resource declarations isn't
very DRY (resource defaults are another option).  Deriving this value
from /etc/foreman-proxy/settings.yml is consistent with looking up
tls/oauth data in /etc/foreman/settings.yaml.


jhoblitt:
rebased on current master

ehelms (approved):

jhoblitt:
rebased on current master.

jhoblitt:
rebased on current master.

PR #1060, opened 2022-06-03
Add foreman plugin for netbox

dgoetz:
Depends on https://github.com/theforeman/foreman-packaging/pull/8021


dgoetz:
Rebased and I think Ubuntu 20.04 is simply not working and not related to the code in the PR.

ekohl (approved):
Correct - I'm not sure why Ubuntu 20.04 fails in containers in CI while it works on real VMs.

PR #1067, opened 2022-06-27
Release 20.3.0

upadhyeammit:


upadhyeammit:
As per the discussion with @ekohl we dont require this unless we need these plugins in Foreman 3.3 . @dgoetz do we need these in 3.3 or the changes are targeted for 3.4?

dgoetz:
Support for all these plugins was more for completeness, so no need to go in a specific version. As you have seen there was also a series of pull requests to the installer to have this fully functional, so if it is more easy I am happy with 3.4, if this does not cause any pain feel free to add it to 3.3.

PR #1069, opened 2022-07-04
Set postgresql service status command

ekohl:
This is to test out if https://github.com/puppetlabs/puppetlabs-postgresql/pull/1349 could fix our broken Ubuntu 20.04 tests.


ekohl:
Doesn't appear to help. Not sure if I'm setting it wrong, but closing this PR now.

PR #1073, opened 2022-08-09
Include authn_core when using keycloak

ekohl:
When disabling built-in apache modules in puppetlabs-apache, this module isn't present but the generated file uses AuthType

A draft to see if this is the only module needed or if we need more.


manifests/config/apache.pp
```diff
@@ -225,6 +225,7 @@
     include apache::mod::auth_gssapi
   } elsif $keycloak {
     include apache::mod::auth_openidc
+    include apache::mod::authn_core
```
evgeni, line 228: We also need it in the GSSAPI branch, I guess?

https://github.com/theforeman/puppet-foreman/blob/97f1517a86d04b2f5571f6803dc317639ae0d7fc/templates/auth_gssapi.conf.erb#L4

ekohl: Yes. Ideally we would be passing it in the `$directories` option. In puppetlabs-apache 8.0.0 (just released) is more autodetection (https://github.com/puppetlabs/puppetlabs-apache/commit/040eab99941978665c4ebe37e0932470d8627918) though this particular case isn't covered.

ekohl: > We also need it in the GSSAPI branch, I guess?

Turns out, no because auth_gssapi includes authn_core here:
https://github.com/puppetlabs/puppetlabs-apache/blob/92865b3d81dfd0895c60668b6f5db3e915a4e030/manifests/mod/auth_gssapi.pp#L8

And Apache starts up with it. I'm trying to expand this to have some acceptance tests.

evgeni: Should oidc then also have such a dependency inside puppet-apache?

ekohl:
This will depends on https://github.com/puppetlabs/puppetlabs-apache/pull/2283

manifests/config/apache.pp
ekohl: Not sure. Maybe?

evgeni: My reasoning is, one can't select an authentication Backend without AuthType directive, so every backend should just pull that in?

adamruzicka (approved):
fwiw, adding `authn_core` seems to be enough to get it working

ekohl:
Except on EL8 where you need to enable the DNF module.

ekohl:
This is properly solved in puppetlabs-apache. I'll track this in another way.

PR #1074, opened 2022-08-10
Fixes #35356 - Don't proxy /server-status

ekohl:
This is the default path that Apache uses for the server-status. Right now it's not used by Foreman. If mod_status is present, this allows monitoring of Apache.

### DIFF
--- a/manifests/config/apache.pp
+++ b/manifests/config/apache.pp
@@ -96,7 +96,7 @@ class foreman::config::apache (
   Pattern['^(https?|unix)://'] $proxy_backend = 'unix:///run/foreman.sock',
   Boolean $proxy_add_headers = true,
   Hash $proxy_params = { 'retry' => '0' },
-  Array[String] $proxy_no_proxy_uris = ['/pulp', '/pulp2', '/streamer', '/pub', '/icons'],
+  Array[String] $proxy_no_proxy_uris = ['/pulp', '/pulp2', '/streamer', '/pub', '/icons', '/server-status'],
   Boolean $ssl = false,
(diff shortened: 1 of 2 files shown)


manifests/config/apache.pp
```diff
@@ -96,7 +96,7 @@
   Pattern['^(https?|unix)://'] $proxy_backend = 'unix:///run/foreman.sock',
   Boolean $proxy_add_headers = true,
   Hash $proxy_params = { 'retry' => '0' },
-  Array[String] $proxy_no_proxy_uris = ['/pulp', '/pulp2', '/streamer', '/pub', '/icons'],
+  Array[String] $proxy_no_proxy_uris = ['/pulp', '/pulp2', '/streamer', '/pub', '/icons', '/server-status'],
```
ekohl, line 99: Are `/pulp2` and `/streamer` still things we need to proxy?

ehelms: Streaming is handled directly through the standard API in Pulp 3 now. For `/pulp2` I think this was used during transition and I cannot find it in use anymore.

evgeni: Let's get that in as-is, and drop pulp/stream stuff in a different PR?

evgeni (approved):

manifests/config/apache.pp
ekohl: Sounds good. We'll have a major version bump coming up anyway.

evgeni: Exactly my thoughts, yeah!

PR #1076, opened 2022-08-17
Add hammer plugin for ssh

dgoetz:


ekohl (changes requested):
Could you add tests?
https://github.com/theforeman/puppet-foreman/blob/02bdaf23367991d80f7763366072d990b5f7525f/spec/classes/cli_plugins_spec.rb#L5
https://github.com/theforeman/puppet-foreman/blob/02bdaf23367991d80f7763366072d990b5f7525f/spec/acceptance/foreman_cli_plugins_spec.rb#L17-L34

dgoetz:
Added the tests and they are still running but first already finished and its looking good.

ekohl (approved):

PR #1077, opened 2022-08-18
install oauth gem on puppet agent

jhoblitt:
puppet6 aio bundles the oauth gem but it is not bundled with puppet7
aio.  Most of the native types in this module require this gem to be
present.

### DIFF
--- a/manifests/install.pp
+++ b/manifests/install.pp
@@ -47,4 +47,9 @@ class foreman::install {
   if $foreman::register_in_foreman {
     contain foreman::providers
   }
+
+  package { 'oauth':
+    ensure   => installed,
+    provider => 'puppet_gem',
+  }
 }
--- a/spec/classes/foreman_install_spec.rb
+++ b/spec/classes/foreman_install_spec.rb
@@ -6,6 +6,8 @@ describe 'foreman' do
       let(:facts) { facts }
       let(:params) { {} }
 
+      it { is_expected.to contain_package('oauth').with_provider('puppet_gem') }
+
       describe 'with version' do
         let(:params) { super().merge(version: 'latest') }
         it { should_not contain_foreman__repos('foreman') }


ekohl (changes requested):
I think this is invalid. We have [foreman::providers](https://github.com/theforeman/puppet-foreman/blob/master/manifests/providers.pp) already to take care of this. That installs `puppet-agent-oauth` which installs the gem in the puppetserver gem environment. However, when you upgrade from Puppetserver 6 to 7 you need to reinstall it. Our installer has [a hook](https://github.com/theforeman/foreman-installer/blob/develop/hooks/pre/31-puppet_agent_oauth.rb) to do this automatically.

jhoblitt:
@ekohl I was ignorant to the existence of `foreman::providers`.  It appears that it is installing the `puppet-agent-oauth` package on my EL7 systems. That package contains:

```bash
[jhoblitt@foreman ~]$ repoquery --list puppet-agent-oauth-0.5.5-1.el7.noarch
/usr/share/foreman-installer/gems/oauth-0.5.5.gem
```

Which doesn't solve the issue my problem as that gem does not end up being installed as part of the puppet agents env. I am not using the types in this package via the installer but as part of regular puppet agent runs.

ekohl:
That's handled via `%post`:
https://github.com/theforeman/foreman-packaging/blob/b38523b6dd1edb81d90e21875275b4ce97f29cff/packages/foreman/puppet-agent-oauth/puppet-agent-oauth.spec#L32

It really is a hack, but native packaging was hard and it works OK-ish. Except when puppet-agent changes its Ruby version, like with Puppet's major versions.

jhoblitt:
What about installing `/usr/share/foreman-installer/gems/oauth-0.5.5.gem` with the `puppet_gem` provider?

ekohl:
> What about installing `/usr/share/foreman-installer/gems/oauth-0.5.5.gem` with the `puppet_gem` provider?

While I like the idea, we bumped it to 0.5.10 in current versions. I think it should provide a symlink `oauth-latest.gem` pointing to the correct version. We could do this in packaging and only then start relying on it. That would also allow us to get rid of the hook to reinstall.

jhoblitt:
I migrated my production sites from 6 -> 7 a month ago, so this PR is no longer relevant for me.  Should we leave it open for someone else to potentially pick up or close it?

ekohl:
This may become relevant again when Puppet 8 comes out so I'd leave it open for now.

PR #1078, opened 2022-08-22
Fixes #33956 - serve static assets directly via Apache

evgeni:


evgeni:
Depends on https://github.com/theforeman/foreman-selinux/pull/145

ekohl (commented):
I recall back then @ehelms said Katello's content wasn't properly served. Is that no longer an issue?

evgeni:
> I recall back then @ehelms said Katello's content wasn't properly served. Is that no longer an issue?

And @ehelms said to me @ekohl had reported issues with Katello… ;-)

In my testing, I could open Katello pages just fine and interact with even more complicated UI/JS elements without any issue.

ekohl:
Perhaps it was the apidoc but that's now served dynamically anyway.

Which also reminds me of a semi-related matter: do we need to purge the statically generated apidoc files from systems?

evgeni:
Apidoc doesn't serve anything in the `/(webpack|assets)/` space, so I'd doubt it?

But yes, it might be a good idea to clean stuff up.

ekohl (approved):
Merging this would break compatibility for anyone without the foreman-selinux change while we normally aim for compatibility with older versions (https://github.com/theforeman/puppet-foreman#support-policy). Does it deserve a note?

I am thinking about compatibility now (which I should perhaps have done before the foreman-selinux merge). Can Puma still read the static content files with the new type or is that denied? If it's still allowed to, then it'll be safe to cherry pick the foreman-selinux change into a stable release. If not, it really needs to go hand in hand.

evgeni:
> I am thinking about compatibility now (which I should perhaps have done before the foreman-selinux merge). Can Puma still read the static content files with the new type or is that denied? If it's still allowed to, then it'll be safe to cherry pick the foreman-selinux change into a stable release. If not, it really needs to go hand in hand.

While unexpected, puma *can* read the files (or gets them from somewhere else, idk) and works just fine.

evgeni:
Nice, https://katello.example.com/lifecycle_environments is broken!

evgeni:
> Nice, https://katello.example.com/lifecycle_environments is broken!

https://github.com/theforeman/foreman-packaging/pull/8377 should fix it!

evgeni:
@ekohl added a few lines to `README.md`, please tell me what you think :)

README.md
```diff
@@ -77,6 +77,11 @@ previous stable release.
 This module targets Foreman 3.1+.
 The module can not be used to manage Foreman installations on EL7.
 
+This module configures Apache to serve static assets from
+`/var/lib/foreman/public` directly, which requires an appropriate
+SELinux policy. Such a policy was introduced in [`foreman-selinux`
+version 3.5](https://projects.theforeman.org/issues/35402).
```
ekohl, line 83: Additionally, plugins need some work. To use the old style, set `foreman::config::apache::proxy_no_proxy_uris` to not include `/assets`.

Or something along those lines.

evgeni:
Plugins are rebuilt, https://github.com/theforeman/foreman-selinux/pull/147 was merged (and should show up in nightly tomorrow), so I think this is good to be merged now.

ekohl (approved):
Should we label this as backwards-incompatible?

evgeni:
> Should we label this as backwards-incompatible?

image

I did!

ekohl:
We may want to do a release before. For example, https://github.com/theforeman/puppet-foreman/pull/1074 could go into 3.4 (and would create a merge conflict here).

evgeni:
> We may want to do a release before. For example, #1074 could go into 3.4 (and would create a merge conflict here).

I've acked that one, would you prep a release then?

evgeni:
(rebased)

PR #1079, opened 2022-08-26
Release 21.1.0

ekohl:


evgeni (approved):

ekohl:
Shall we just release this now with failing tests? It's only because of the reports plugin and this way it's compatible for an older release if needed.

evgeni:
> Shall we just release this now with failing tests? It's only because of the reports plugin and this way it's compatible for an older release if needed.

I'm voting release as is, yes!

PR #1080, opened 2022-08-26
Drop /pulp2 and /streamer from no_proxy_uris

evgeni:
Includes #1078, because I am lazy and don't want to rebase again.


evgeni:
Should this be labeled `backwards-incompatible`? I think it technically is?

ekohl:
> Should this be labeled `backwards-incompatible`? I think it technically is?

Yes. We're doing a major release anyway so it's just a clearer signal to users reading the changelog.

> because I am lazy and don't want to rebase again.

Looks like GH doesn't recognize the commit is redundant. Needs a (trivial) rebase now :(

ekohl (approved):

evgeni:
Rebased!

Yes, GH doesn't recognize such trivial rebases, but better than getting a vim window again ;)

PR #1083, opened 2022-09-05
Refs #35473 - Configure Apache for API extlogin

ofedoren:
Required by https://github.com/theforeman/foreman/pull/9406

### DIFF
--- a/templates/auth_gssapi.conf.erb
+++ b/templates/auth_gssapi.conf.erb
@@ -1,5 +1,5 @@
 
-<Location /users/extlogin>
+<LocationMatch ^(/api(/v2)?)?/users/extlogin/?$>
   SSLRequireSSL
   AuthType GSSAPI
   AuthName "GSSAPI Single Sign On Login"
@@ -11,5 +11,4 @@
   ErrorDocument 401 '<html><meta http-equiv="refresh" content="0; URL=/users/login"><body>Kerberos authentication did not pass.</body></html>'
   # The following is needed as a workaround for https://bugzilla.redhat.com/show_bug.cgi?id=1020087
   ErrorDocument 500 '<html><meta http-equiv="refresh" content="0; URL=/users/login"><body>Kerberos authentication did not pass.</body></html>'
-</Location>
-
+</LocationMatch>
--- a/templates/lookup_identity.conf.erb
+++ b/templates/lookup_identity.conf.erb
@@ -1,5 +1,5 @@
 
-<LocationMatch ^/users/(ext)?login/?$>
+<LocationMatch ^(/api(/v2)?)?/users/(ext)?login/?$>
   LookupUserAttr email REMOTE_USER_EMAIL
   LookupUserAttr firstname REMOTE_USER_FIRSTNAME
   LookupUserAttr lastname REMOTE_USER_LASTNAME
@@ -13,4 +13,3 @@
   RequestHeader set REMOTE_USER_LASTNAME %{REMOTE_USER_LASTNAME}e
   RequestHeader set REMOTE_USER_GROUPS %{REMOTE_USER_GROUPS}e
 </LocationMatch>
-
--- a/templates/settings-external-auth.yaml.erb
+++ b/templates/settings-external-auth.yaml.erb
@@ -1,3 +1,3 @@
 :authorize_login_delegation: true
+:authorize_login_delegation_api: true
 :authorize_login_delegation_auth_source_user_autocreate: External
-


templates/auth_gssapi.conf.erb
```diff
@@ -13,3 +13,16 @@
...
+  AuthType GSSAPI
+  AuthName "GSSAPI Single Sign On Login"
+  GssapiCredStore keytab:<%= @http_keytab %>
+  GssapiSSLonly On
+  GssapiLocalName On
+  # require valid-user
+  require pam-account <%= scope.lookupvar('foreman::pam_service') %>
+  ErrorDocument 401 '<html><meta http-equiv="refresh" content="0; URL=/users/login"><body>Kerberos authentication did not pass.</body></html>'
+  # The following is needed as a workaround for https://bugzilla.redhat.com/show_bug.cgi?id=1020087
+  ErrorDocument 500 '<html><meta http-equiv="refresh" content="0; URL=/users/login"><body>Kerberos authentication did not pass.</body></html>'
```
adamruzicka, line 27: Since this is an API endpoint, maybe not returning html on errrors would be a good idea?

adamruzicka (commented):
One nitpick inline, otherwise seems to work well.

templates/auth_gssapi.conf.erb
```diff
@@ -13,3 +13,16 @@
...
+  SSLRequireSSL
+  AuthType GSSAPI
+  AuthName "GSSAPI Single Sign On Login"
+  GssapiCredStore keytab:<%= @http_keytab %>
+  GssapiSSLonly On
+  GssapiLocalName On
+  # require valid-user
+  require pam-account <%= scope.lookupvar('foreman::pam_service') %>
+  ErrorDocument 401 '<html><meta http-equiv="refresh" content="0; URL=/users/login"><body>Kerberos authentication did not pass.</body></html>'
+  # The following is needed as a workaround for https://bugzilla.redhat.com/show_bug.cgi?id=1020087
```
ekohl, line 26: This was reported against RHEL 6, which used Apache 2.2. Today we use Apache 2.4. Is that bug still really a problem? It was also reported against mod_auth_kerb while today we use mod_gssapi. Perhaps there are better ways now.

I realize this is the same as the old extlogin, but given you have a setup where you can test it's a good chance.

templates/auth_gssapi.conf.erb
```diff
@@ -13,3 +13,16 @@
   ErrorDocument 500 '<html><meta http-equiv="refresh" content="0; URL=/users/login"><body>Kerberos authentication did not pass.</body></html>'
 </Location>
 
+<Location /api/users/extlogin>
```
ekohl, line 16: To minimize duplication you can also use `LocationMatch`:
```
<LocationMatch "^(/api)?/users/extlogin">
```

templates/lookup_identity.conf.erb
```diff
@@ -14,3 +14,17 @@
   RequestHeader set REMOTE_USER_GROUPS %{REMOTE_USER_GROUPS}e
 </LocationMatch>
 
+<Location /api/users/extlogin>
```
ekohl, line 17: Perhaps this should be folded into the existing `LocationMatch`?

templates/auth_gssapi.conf.erb
ofedoren: I thought so as well, but after changes per Ewoud suggestions it's not obvious to me how to distinguish between UI vs API...

templates/auth_gssapi.conf.erb
ofedoren: Surely I'm missing here something, since I've never touched Apache before :/ But why do we even support `500` here? Does it come from one of the modules just because of some bug?..

ofedoren:
Thanks, @adamruzicka and @ekohl, updated.

templates/lookup_identity.conf.erb
```diff
@@ -1,5 +1,5 @@
 
-<LocationMatch ^/users/(ext)?login/?$>
+<LocationMatch ^(/api)?/users/(ext)?login/?$>
```
adamruzicka, line 2: This would also match `/api/users/login` which doesn't exist in foreman. Is that an issue?

templates/auth_gssapi.conf.erb
```diff
@@ -1,5 +1,5 @@
 
-<Location /users/extlogin>
+<LocationMatch ^(/api)?/users/extlogin/?$>
```
adamruzicka, line 2: Should it also match `/api/v2/users/extlogin`?

adamruzicka:
Is it just me or do the tests seem to be stuck?

Anyway, I'd say this is good. @ekohl anything else to do here?

templates/lookup_identity.conf.erb
evgeni, line 2: Should this be
```suggestion
<LocationMatch ^(/api(/v2)?)?/users/(ext)?login/?$>
```
To match the same routes as `auth_gssapi.conf.erb` above?

evgeni:
Kicked CI and added a tiny comment, otherwise LGTM!

ofedoren:
Thanks, @evgeni, updated.

evgeni:
@ekohl any comments from your side?

ekohl (approved):
No more comments, I think they're all addressed.

PR #1089, opened 2022-10-18
Convert per_page in foreman::foreman to string

ekohl:
CGI.escape can only deal with strings and the function accepts per_page as an integer.


evgeni (approved):
IGNORE ALL FAILURES!

PR #1091, opened 2022-10-25
Refs #35641 - Add option to configure FOREMAN_PUMA_WORKER_TIMEOUT

wbclark:
This specifies the time before a worker process that has not checked in will be considered lost and automatically restarted. This can be useful to tune if worker timeouts are observed that are not actually due to a hanging process; it is NOT a request timeout. When undefined, it falls back to the default value of 60 seconds.

Requires https://github.com/theforeman/foreman/pull/9485 and co-authored with @jpasqualetto

### DIFF
--- a/templates/foreman.service-overrides.erb
+++ b/templates/foreman.service-overrides.erb
@@ -5,3 +5,8 @@ Environment=FOREMAN_HOME=<%= scope['foreman::app_root'] %>
 Environment=FOREMAN_PUMA_THREADS_MIN=<%= scope['foreman::config::min_puma_threads'] %>
 Environment=FOREMAN_PUMA_THREADS_MAX=<%= scope['foreman::foreman_service_puma_threads_max'] %>
 Environment=FOREMAN_PUMA_WORKERS=<%= scope['foreman::config::puma_workers'] %>
+<% unless scope['foreman::foreman_service_puma_worker_timeout'].nil? -%>
+# Puma worker_timeout, which specifies the time for each worker to check in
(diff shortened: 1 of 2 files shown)


templates/foreman.service-overrides.erb
```diff
@@ -5,3 +5,11 @@ Environment=FOREMAN_HOME=<%= scope['foreman::app_root'] %>
 Environment=FOREMAN_PUMA_THREADS_MIN=<%= scope['foreman::config::min_puma_threads'] %>
 Environment=FOREMAN_PUMA_THREADS_MAX=<%= scope['foreman::foreman_service_puma_threads_max'] %>
 Environment=FOREMAN_PUMA_WORKERS=<%= scope['foreman::config::puma_workers'] %>
+# Puma worker_timeout, which specifies the time for each worker to check in
+# before it is assumed hanging and then restarted. This is not a request timeout.
+# If not specified here, the default value of 60 seconds is used.
```
ekohl, line 10: I don't like hardcoding defaults all over the place. Especially in comments. If they change, they're out of sync and misleading. Since it's normal that when it's not specified, the default is used so I'd leave this line out.
```suggestion
```

wbclark: I updated to remove all hardcoded values, and moved all comments into the conditional so they will only be visible in the template when the user has configured a value here.

wbclark:
Closing this as it was agreed not to expose this configuration in the installer, so that we don't set the expectation that this value should be altered under any normal circumstances.

PR #1100, opened 2022-12-07
Fixes #35832 - set default_domain_suffix in sssd.conf

snagoor:
Setting a default_domain_suffix for IDM based authentication would reduce an effort to type in the @domain while authenticating.

https://docs.theforeman.org/nightly/Administering_Project/index-foreman-el.html#Using_FreeIPA_admin

### DIFF
--- a/manifests/config.pp
+++ b/manifests/config.pp
@@ -165,6 +165,12 @@ class foreman::config {
           enable  => true,
           require => Package['sssd-dbus'],
         }
+
+        $ipa_config = '/etc/ipa/default.conf'
+        if find_file($ipa_config) {
+          $realm_entry = file($ipa_config).match('^realm\s*=\s*(.*)$')
+          $default_realm = downcase($realm_entry[1])
+        }
       }
 
       file { "/etc/pam.d/${foreman::pam_service}":
@@ -206,7 +212,7 @@ class foreman::config {
 
       if $foreman::ipa_manage_sssd {
         $sssd = pick(fact('foreman_sssd'), {})
-        $sssd_default_domain_suffix = $facts['networking']['domain']
+        $sssd_default_domain_suffix = $default_realm
         $sssd_services = join(unique(pick($sssd['services'], []) + ['ifp']), ', ')
         $sssd_ldap_user_extra_attrs = join(unique(pick($sssd['ldap_user_extra_attrs'], []) + ['email:mail', 'lastname:sn', 'firstname:givenname']), ', ')
         $sssd_allowed_uids = join(unique(pick($sssd['allowed_uids'], []) + [$apache::user, 'root']), ', ')
--- a/manifests/config.pp
+++ b/manifests/config.pp
@@ -206,6 +206,7 @@ class foreman::config {
 
       if $foreman::ipa_manage_sssd {
         $sssd = pick(fact('foreman_sssd'), {})
+        $sssd_default_domain_suffix = $facts['networking']['domain']
         $sssd_services = join(unique(pick($sssd['services'], []) + ['ifp']), ', ')
         $sssd_ldap_user_extra_attrs = join(unique(pick($sssd['ldap_user_extra_attrs'], []) + ['email:mail', 'lastname:sn', 'firstname:givenname']), ', ')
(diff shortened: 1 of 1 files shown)


manifests/config.pp
```diff
@@ -206,6 +206,7 @@
 
       if $foreman::ipa_manage_sssd {
         $sssd = pick(fact('foreman_sssd'), {})
+        $sssd_default_domain_suffix = join(unique(pick($sssd['default_domain_suffix'], []) + [$::domain]), ', ')
```
ekohl, line 209: The use of `$::X` is deprecated. Instead, we now use `$facts['domain']`.
```suggestion
        $sssd_default_domain_suffix = join(unique(pick($sssd['default_domain_suffix'], []) + [$facts['domain']]), ', ')
```

This implies `default_domain_suffix` is an array. Is that correct?

Additionally, the fact is not retrieved from sssd (see https://github.com/theforeman/puppet-foreman/blob/master/lib/facter/sssd.rb) so `$sssd['default_domain_suffix']` is always empty.

That means right now it's effectively:
```suggestion
        $sssd_default_domain_suffix = $facts['domain']
```

ekohl (changes requested):
Why is it needed to set this value if it matches the domain? Are there cases where it may differ from the actual domain and this actually breaks things? In other words: should it actually be a parameter?

manifests/config.pp
snagoor: And sincere apologies for the delay in replying.

ekohl: > And sincere apologies for the delay in replying.

Not at all - I actually had PTO on Thursday and Friday so to me there was no delay.

snagoor:
Validations seems to have failed as below
~~~
manifests/config.pp:209:legacy_facts:WARNING:legacy fact
Error: Process completed with exit code 1.
~~~

this `$facts['domain']` seems to be deprecated and I am not sure but reading the docs https://puppet.com/docs/puppet/6/core_facts.html#domain it seems the new format https://puppet.com/docs/puppet/6/core_facts.html#networking to be `$facts['networking']['domain']` . Please correct me if I am wrong.

manifests/config.pp
```diff
@@ -206,6 +206,7 @@
 
       if $foreman::ipa_manage_sssd {
         $sssd = pick(fact('foreman_sssd'), {})
+        $sssd_default_domain_suffix = $facts['networking']['domain']
```
ekohl, line 209: My biggest concern here is where the domain does not match the realm. For example, you may have `foreman.lab.example.com` joined to the `example.com` realm. `$facts['networking']['domain']` would then be `lab.example.com`, but is that really correct? I'm not too familiar with this, so feel free to correct me if I'm wrong.

ekohl (commented):
Sorry, this fell off my radar.

manifests/config.pp
snagoor: @ekohl, I think the default realm configured can be identified from the `/etc/ipa/default.conf` file. This file exists in both the authentication methods i.e. integration with IPA and AD (specifically we ask users to configure this file in direct AD auth using gssproxy https://docs.theforeman.org/3.0/Administering_Red_Hat_Satellite/index-foreman-el.html#sect-Administering-Configuring_External_Authentication-Using_Active_Directory).

Another way is to rely on end users to provide this value as a installer parameter in order to configure it.

ekohl: > Another way is to rely on end users to provide this value as a installer parameter in order to configure it.

Right now I'm leaning in that direction.

manifests/config.pp
```diff
@@ -165,6 +165,12 @@
           enable  => true,
           require => Package['sssd-dbus'],
         }
+
+        $ipa_config = '/etc/ipa/default.conf'
+        if find_file($ipa_config) {
+          $realm_entry = file($ipa_config).match('^realm\s*=\s*(.*)$')
```
ekohl, line 171: I don't think this does what you think it does. AFAIK `file()` runs on the Puppetserver so in a networked environment this will copy the realm from the Puppetserver to the clients. That's why we use facts for all the other entries

adamruzicka:
@snagoor gentle ping

evgeni:
CLosing as https://github.com/theforeman/puppet-foreman/pull/1160 was merged

PR #1101, opened 2022-12-07
Refs #33956 - Ensure mod_expires is loaded

ekohl:
In case only minimal Apache modules are loaded then mod_expires may not be present. The result is that assets are sent without headers that allow browsers to cache it.


evgeni:
```
      # Puppet::Resource::Catalog::DuplicateResourceError:
      #   Duplicate declaration: Apache::Mod[expires] is already declared at (file: /home/runner/work/puppet-foreman/puppet-foreman/spec/fixtures/modules/apache/manifests/default_mods.pp, line: 73); cannot redeclare (file: /home/runner/work/puppet-foreman/puppet-foreman/spec/fixtures/modules/apache/manifests/mod/expires.pp, line: 23)
```

This somehow implies our tests don't run with `apache::default_mods: false`?

ekohl:
I just noticed this yesterday on my own server, but I'll do some more digging next week.

And it does indicate an incompatibility, which is actually good to know about: we have to choose one or the other. Now I'm reminded about https://github.com/puppetlabs/puppetlabs-apache/pull/2288, which I started but didn't have a discussion.

ekohl:
No code change, but I opened https://projects.theforeman.org/issues/35870 so we don't lose track of this.

evgeni:
FWIW, the regression is there also before #33956

The original ["disable default modules"](https://projects.theforeman.org/issues/34590) change triggered that and I see the headers missing on a 3.3 installation that does not yet have the "assets via apache" part.

ekohl:
I ended up working around `apache::mod::expires` including additional config so it can be cherry picked safely.

evgeni (approved):

PR #1105, opened 2023-01-30
External auth tweaks

adamruzicka:
@ofedoren's changes add support username+password based external auth for api logins to match what webui can do

My changes build on top of that and detach api external auth from webui external auth


templates/auth_gssapi.conf.erb
```diff
@@ -1,11 +1,19 @@
...
-  AuthName "GSSAPI Single Sign On Login"
-  GssapiCredStore keytab:<%= @http_keytab %>
-  GssapiSSLonly On
-  GssapiLocalName <%= @gssapi_local_name %>
+  <If "%{HTTP:Authorization} =~ /^Basic/">
+    AuthType Basic
+    AuthName "PAM Authentication"
+    AuthBasicProvider PAM
+    AuthPAMService foreman
+  </If>
```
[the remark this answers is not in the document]
adamruzicka: Ah, I see, done

manifests/config.pp
```diff
@@ -228,6 +229,25 @@
         content => template('foreman/settings-external-auth.yaml.erb'),
         order   => '02',
       }
+
+      file { "/etc/httpd/conf.modules.d/auth_basic.load":
```
ekohl, line 233: I'd really like to avoid this. I created [apache::mod::auth_basic](https://github.com/puppetlabs/puppetlabs-apache/blob/main/manifests/mod/auth_basic.pp) so can we use that instead? So `include apache::mod__auth_basic`.

ekohl (changes requested):

manifests/config.pp
ofedoren: Yeah, I guess that's the right approach, I just did know how to write it in the Puppet way :/

templates/auth_gssapi_api.conf.erb
```diff
@@ -0,0 +1,21 @@
+
```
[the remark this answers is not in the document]
adamruzicka: | ipa_auth | ipa_auth_api | contents |
| ---- | ----- | ----- |
| false | * | do not deploy the file |
| true | false | set up basic auth |
| true | true | set up gssapi |

So like this? Or should basic external auth also get its own flag/value?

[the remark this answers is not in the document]
ofedoren: Yeah, something like this table. Actually, as for me, the table illustrates ideally my idea.

templates/external_auth_api.conf.erb
```diff
@@ -0,0 +1,20 @@
...
+  AuthType GSSAPI
+  AuthName "GSSAPI Single Sign On Login"
+  GssapiCredStore keytab:<%= @http_keytab %>
+  GssapiSSLonly On
+  GssapiLocalName <%= @gssapi_local_name %>
+  <% else %>
+  AuthType Basic
+  AuthName "PAM Authentication"
+  AuthBasicProvider PAM
+  AuthPAMService foreman
```
ofedoren, line 14: Instead of hardcoding, I guess this should be 
```suggestion
  AuthPAMService <%= scope.lookupvar('foreman::pam_service') %>
```

templates/external_auth_api.conf.erb
```diff
@@ -0,0 +1,20 @@
+
+<LocationMatch ^/api(/v2)?/users/extlogin/?$>
+  SSLRequireSSL
+  <% if scope.lookupvar('$foreman::ipa_authentication_api') %>
+  AuthType GSSAPI
+  AuthName "GSSAPI Single Sign On Login"
+  GssapiCredStore keytab:<%= @http_keytab %>
+  GssapiSSLonly On
+  GssapiLocalName <%= @gssapi_local_name %>
+  <% else %>
```
ofedoren, line 10: Ruby's `if` statements would enable GSSAPI **or** Basic auth **only**, which is not quite right. The previous Apache's module `if` statement must remain since it allow us to use either negotiation or credentials for the auth depending on the request headers.

ofedoren:
I've opened a PR https://github.com/theforeman/puppet-foreman/pull/1108 that should finish work on this. Sorry for kind of spam, but due to permissions I think this is a better approach :/

adamruzicka:
Closing in favor of #1108

PR #1106, opened 2023-02-01
Fixes #36028 - ensure compressed assets are returned if available

evgeni:
Fixes: 6a02be6e8faecb4210f16a8a5aea37397bfe3f74

### DIFF
--- a/files/_assets.conf.erb
+++ b/files/_assets.conf.erb
@@ -1,30 +1,19 @@
-# Return compressed assets if they are precompiled
-<IfModule mod_rewrite.c>
-  RewriteEngine on
-  # Make sure the browser supports gzip encoding and file with .gz added
-  # does exist on disc before we rewrite with the extension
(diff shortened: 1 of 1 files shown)


files/_assets.conf.erb
```diff
@@ -38,4 +27,15 @@
     ExpiresActive On
     ExpiresDefault "access plus 1 year"
   </IfModule>
+
+  # Return compressed assets if they are precompiled
+  <IfModule mod_rewrite.c>
```
ekohl, line 32: I think the `IfModule` part mostly stems from when we used to support dropping in some file, but not really managing Apache. Any thoughts on dropping it and adding the rewrite rules unconditionally?

ekohl (approved):

files/_assets.conf.erb
evgeni: We use it unconditionally in other places already, so ack.

PR #1108, opened 2023-02-01
Add basic external auth for API

ofedoren:
This PR is a replacement since we decided with @adamruzicka that I'll take over https://github.com/theforeman/puppet-foreman/pull/1105.


adamruzicka:
Cross-posting from the older PR https://github.com/theforeman/puppet-foreman/pull/1105#discussion_r1090709897

This seems to cause some tests to fail. @ekohl could you please give us a hand here?

manifests/config.pp
```diff
@@ -228,6 +228,19 @@
         content => template('foreman/settings-external-auth.yaml.erb'),
         order   => '02',
       }
+
+      include apache::mod__auth_basic
```
ekohl, line 232:
```suggestion
      include apache::mod::auth_basic
```

templates/external_auth_api.conf.erb
```diff
@@ -0,0 +1,28 @@
+
+<LocationMatch ^/api(/v2)?/users/extlogin/?$>
+  SSLRequireSSL
+  <% if scope.lookupvar('$foreman::ipa_authentication_api') %>
```
ekohl, line 4:
```suggestion
  <% if scope.lookupvar('foreman::ipa_authentication_api') %>
```

templates/external_auth_api.conf.erb
```diff
@@ -0,0 +1,28 @@
+
```
ekohl, line 1: I'll note that we'd like to move to [EPP](https://www.puppet.com/docs/puppet/7/lang_template_epp.html) instead of ERB. I won't block on it, but I like that you can pass an explicit context. From the example:
```puppet
epp('example/example.epp', { 'logfile' => "/var/log/ntp.log" })
```

Then you can simply use `$logfile` in the template, rather than using `scope.lookupvar()`. In my experience it gives more reliability that templates actually render correctly, especially if you add in data types at the top.

ofedoren: I'd leave the current approach. I guess transition to EPP will be in batch anyway for all current files at some point...

adamruzicka (approved):
I tested this together with hammer and apipie-binding changes and everything seems to be working well.

templates/external_auth_api.conf.erb
```diff
@@ -0,0 +1,28 @@
...
+<LocationMatch ^/api(/v2)?/users/extlogin/?$>
+  SSLRequireSSL
+  <% if scope.lookupvar('foreman::ipa_authentication_api') %>
+  <If "%{HTTP:Authorization} =~ /^Basic/">
+    AuthType Basic
+    AuthName "PAM Authentication"
+    AuthBasicProvider PAM
```
ekohl, line 8: I think this implies the pam module should be included. Note that https://httpd.apache.org/docs/2.4/mod/mod_auth_basic.html#authbasicprovider doesn't list it as a built in provider, but https://www.adelton.com/apache/mod_authnz_pam/ does document what you're doing here.

There is `apache::mod::authnz_pam` which will ensure the correct package is installed and the module is loaded.

manifests/init.pp
```diff
@@ -46,6 +46,8 @@
 #
 # $ipa_authentication::           Enable configuration for external authentication via IPA
 #
+# $ipa_authentication_api::       Enable configuration for external authentication via IPA for API
```
ekohl, line 49: I'm trying to imagine the workflows. When would you set `ipa_authentication` but set this to `false`? Wouldn't it more common that you keep the two in sync?

Or in other words: do we need this additional parameter?

ekohl (commented):
I'd prefer to have some acceptance test to see if Apache even starts up with this. You can take `spec/acceptance/foreman_basic_spec.rb` and use some manifest like
```puppet
class { 'foreman':
  ipa_authentication => true,
  ipa_authentication_api => true,
}
```

Though perhaps there are more parameters that need to be set and we can't really test this well.

manifests/init.pp
ofedoren: Maybe it's worth to rename the option to something more explicit, since this option is for enabling GSSAPI auth for API. Basic external auth will be enabled by default if `ipa_authentication` is `true`.

ofedoren:
@ekohl, I've updated old tests and added a new one, but I'm not sure if it even tests something :/ Couldn't figure out how to write the acceptance test right.

spec/acceptance/foreman_basic_spec.rb
```diff
@@ -16,4 +16,24 @@
   describe package('foreman-telemetry') do
     it { is_expected.not_to be_installed }
   end
+
+  context 'GSSAPI auth enabled' do
+    let(apache_service) { %w[debian ubuntu].include?(os[:family]) ? 'apache2' : 'httpd' }
```
ekohl:
```suggestion
    let(:apache_service) { %w[debian ubuntu].include?(os[:family]) ? 'apache2' : 'httpd' }
```

spec/acceptance/foreman_basic_spec.rb
```diff
@@ -16,4 +16,24 @@
   describe package('foreman-telemetry') do
     it { is_expected.not_to be_installed }
   end
+
+  context 'GSSAPI auth enabled' do
+    let(:apache_service) { %w[debian ubuntu].include?(os[:family]) ? 'apache2' : 'httpd' }
```
ekohl, line 21: Technically it needs to be a symbol:
```suggestion
    let(:apache_service) { %w[debian ubuntu].include?(os[:family]) ? 'apache2' : 'httpd' }
```

But this is part of the [shared example](https://relishapp.com/rspec/rspec-core/docs/example-groups/shared-examples), defined in https://github.com/theforeman/puppet-foreman/blob/master/spec/support/acceptance/examples.rb. So it'd be sufficient to just use the `it_behaves_like` part.

ofedoren: Sorry, a typo I've noticed by failed tests as well :/

I've removed duplicate tests for service, thanks for pointing out :)

manifests/config.pp
```diff
@@ -228,6 +228,19 @@
         content => template('foreman/settings-external-auth.yaml.erb'),
         order   => '02',
       }
+
+      include apache::mod::auth_basic
```
ekohl, line 232: Doesn't this belong inside the `if` statement, given it's needed for the apache fragment?

spec/acceptance/foreman_basic_spec.rb
```diff
@@ -16,4 +16,17 @@
   describe package('foreman-telemetry') do
     it { is_expected.not_to be_installed }
   end
+
+  context 'GSSAPI auth enabled' do
+    it_behaves_like 'the foreman application' do
```
ekohl, line 21: I don't think you can use it this way since there's no application of it in the shared resource. So what I meant was:
```ruby
it_behaves_like 'an idempotent resource' do
  let(:manifest) do
    # ...
  end
end

it_behaves_like 'the foreman application'
```

ekohl (commented):
Two small comments, otherwise I think this should be good.

manifests/config.pp
ofedoren: I've removed that condition since the whole block is already under the same `if` statement.

spec/acceptance/foreman_basic_spec.rb
ofedoren: Thanks a lot for navigating here!

ekohl:
The problem you're running into is that it tries to enroll the system using `kinit`, but that's not found. Even if it was, it's not going to find anything since there is no Kerberos server.

You can trick it into skipping this by creating the file before the run.

I think this should work:
```ruby
before { on default, 'touch /path/to/http-keytab-file' }
```

That path is by default the HTTP conf dir + `/http.keytab`, but you can pass in `http_keytab` as an option which I'd recommend. That makes it more reliable.

ofedoren:
@ekohl, could you please help me once more? I've added your suggestions, but the tests still fail. Logs are not quite useful since I can't see what's the actual problem :/

ekohl:
You should be able to test it locally. I use (based on https://github.com/voxpupuli/voxpupuli-acceptance/#running-tests)
```
BEAKER_HYPERVISOR=vagrant_libvirt BEAKER_destroy=no bundle exec rspec spec/acceptance/foreman_basic_spec.rb -e GSSAPI
```

Then I use `cd .vagrant/beaker_vagrant_files/beaker_centos8* && vagrant ssh`. That allows me to inspect the logs.

Then when I run it again I add `BEAKER_provision=no` to reuse the same machine and only apply the tests, allowing me to iterate.

You should know that we have some idempotency issues in our CI with SELinux, which you will see in a VM that you don't see on Docker (which we use here in GitHub Actions).

ekohl:
Another useful thing: `BEAKER_PUPPET_DEBUG=yes` can be passed to get Puppet to run in debug mode. That tells me:

```
  Debug: Augeas[sssd-ifp-extra-attributes](provider=augeas): Opening augeas with root /, lens path , flags 32
  Debug: Augeas[sssd-ifp-extra-attributes](provider=augeas): Augeas version 1.12.0 is installed
  Debug: Augeas[sssd-ifp-extra-attributes](provider=augeas): Will attempt to save and only run if files changed
  Debug: Augeas[sssd-ifp-extra-attributes](provider=augeas): sending command 'set' with params ["/files/etc/sssd/sssd.conf/target[.=~regexp('domain/.*')]/ldap_user_extra_attrs", "email:mail, lastname:sn, firstname:givenname"]
  Debug: Augeas[sssd-ifp-extra-attributes](provider=augeas): sending command 'set' with params ["/files/etc/sssd/sssd.conf/target[.='sssd']/services", "ifp"]
  Debug: Augeas[sssd-ifp-extra-attributes](provider=augeas): sending command 'set' with params ["/files/etc/sssd/sssd.conf/target[.='ifp']", "ifp"]
  Debug: Augeas[sssd-ifp-extra-attributes](provider=augeas): sending command 'set' with params ["/files/etc/sssd/sssd.conf/target[.='ifp']/allowed_uids", "apache, root"]
  Debug: Augeas[sssd-ifp-extra-attributes](provider=augeas): sending command 'set' with params ["/files/etc/sssd/sssd.conf/target[.='ifp']/user_attributes", "+email, +firstname, +lastname"]
  Debug: Augeas[sssd-ifp-extra-attributes](provider=augeas): Put failed on one or more files, output from /augeas//error:
  Debug: Augeas[sssd-ifp-extra-attributes](provider=augeas): /augeas/files/etc/sssd/sssd.conf/error = put_failed
  Debug: Augeas[sssd-ifp-extra-attributes](provider=augeas): /augeas/files/etc/sssd/sssd.conf/error/path = /files/etc/sssd/sssd.conf/
  Debug: Augeas[sssd-ifp-extra-attributes](provider=augeas): /augeas/files/etc/sssd/sssd.conf/error/lens = /opt/puppetlabs/puppet/share/augeas/lenses/dist/sssd.aug:33.13-.53:
  Debug: Augeas[sssd-ifp-extra-attributes](provider=augeas): /augeas/files/etc/sssd/sssd.conf/error/message = Failed to match tree under /
  
       { "target" }
       { "target" }
       { "target" = "ifp" }
  
    with pattern
     (    { /#comment/ = /[^\t\n\r ].*[^\t\n\r ]|[^\t\n\r ]/ }
        | { })*
      { /target/ = /[^]\n\r]+/ }*
  Debug: Augeas[sssd-ifp-extra-attributes](provider=augeas): Closed the augeas connection
```

I recall that last time I played around with this I needed to do some special work on sssd when the file didn't exist at all. Normally that's ok because the domain enrollment takes care of it. Since we're stubbing it, it doesn't work. I recall having played around with making it work but can't find that anymore.

For now the shortest workaround is to add `ip_manage_sssd => false`. So I end up with the following manifest:
```puppet
class { 'foreman':
  ipa_authentication     => true,
  ipa_authentication_api => true,
  # Stub out ipa enrollment
  http_keytab            => '/etc/httpd/conf.keytab',
  ipa_manage_sssd        => false,
}
```

Then it at least applies, but idempotency is an issue. And the expectation in the shared example about login, which is now extlogin.

ekohl:
So the idempotency is the SELinux types, but that shouldn't happen in our GHA CI. Then it remains:
```
       expected: "https://centos8-64.example.com/users/login"
            got: "https://centos8-64.example.com/users/extlogin"
```

Which is this part:
https://github.com/theforeman/puppet-foreman/blob/f32405cfadabea8a3dc5bf81243543c055f257f0/spec/support/acceptance/examples.rb#L26-L29

There are 2 options:
* Parametrize the current shared example (https://relishapp.com/rspec/rspec-core/docs/example-groups/shared-examples shows how)
* Copy the tests

@ofedoren does this help?

ofedoren:
@ekohl, I guess so, yes. Although, I wanted to run the tests locally and I'm going nuts, since I don't quite follow what's going on there. It always fails with: Error: Evaluation Error: Error while evaluating a Resource Statement, Could not find declared class foreman::repo (file: /tmp/apply_manifest_175236864.pp.kOeUC5, line: 1, column: 1) on node centos8-64.example.com.

AFAIU puppet needs foreman modules, but the command is puppet apply --debug --detailed-exitcodes /tmp/apply_manifest_175236864.pp.kOeUC5. That temp file is created on local machine and then `scp`ed into VM. But I don't see how puppet command would use foreman's modules. They don't seem to be installed in VM...

ekohl:
We rely on fixtures (https://github.com/voxpupuli/voxpupuli-acceptance/#fixtures) and `rake spec_prep` will download all the fixtures. Normally if you run `rake beaker` it's a dependency, but if you run `rspec` directly that's not happening. There's also `spec_clean` to wipe the fixtures, so `spec_clean spec_prep` is what I sometimes use to update them.

ofedoren:
Thanks a ton, @ekohl, you're awesome. Now the tests seem to pass.

Actually, after you significant assistance here, I think I'm going to add those "TL;DR / Quick start" commands to readme/contribution for future reference or for new people in the next PR.

ekohl:
Do note we modulesync it, so https://github.com/theforeman/foreman-installer-modulesync/blob/master/moduleroot/CONTRIBUTING.md.erb is the correct place to submit the PR.

It may also be good to have it somewhere shared, rather than copied to every module. Not asking you to do so, but wondering what your thoughts are. My own opinions may be less useful since I'm looking for someone new. Not someone who has built up a routine over the past decade.

spec/support/acceptance/examples.rb
```diff
@@ -1,4 +1,4 @@
-shared_examples 'the foreman application' do
+shared_examples 'the foreman application' do |params = { login_url: '/users/login' }|
```
ekohl, line 1: I wonder about this. What if a we add more parameters, but a user doesn't specify a login_url. I don't think they'll get merged.

Instead I'd suggest to use `params = {}` and then `params.get(:login_url, '/users/login')`
 in the actual test. I think it's cleaner than storing a hash somewhere and merging them. Discoverability may be an issue, but I think it's not really an issue right now. We're not making a full API here.

Also a slight naming nit: I'd name this a path, not a URL. Perhaps even `login_url_path`. And if we're fully making it self describing I'd add `expected` to make it `expected_login_url_path`.

ekohl (approved):
Thanks! Do you and @adamruzicka mind if we squash these commits into a single commit?

adamruzicka:
I'm fine with having them squashed

PR #1109, opened 2023-02-02
Fixes #36037 - Manage Redis service for Redis cache

ekohl:
This automatically manages Redis when the cache store type is set to Redis, but no URLs have been provided. It's assumed that when a URL is provided that the service is managed in another way.

This ignores the edge case where the user wants to run it on localhost, but with a different database than 0.

Currently a draft since it doesn't have tests and I haven't verified it yet. The end goal of this is that we can set up Redis caching by default. File based caching is just not the right fit for large deployments, even if it's stored in tmpfs.

### DIFF
--- a/manifests/config.pp
+++ b/manifests/config.pp
@@ -20,6 +20,17 @@ class foreman::config {
     }
   }
 
+  if $foreman::rails_cache_store['type'] == 'redis' {
+    if $foreman::rails_cache_store['urls'] {
+      $redis_cache_urls = prefix($foreman::rails_cache_store['urls'], 'redis://')
+    } else {
+      include redis
+      $redis_cache_urls = ["redis://localhost:${redis::port}/0"]
+    }
+  } else {
+    $redis_cache_urls =  undef
+  }
+
   # Used in the settings template
   $websockets_ssl_cert = pick($foreman::websockets_ssl_cert, $foreman::server_ssl_cert)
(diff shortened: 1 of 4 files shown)


ekohl:
In https://bugzilla.redhat.com/show_bug.cgi?id=2165092#c4 it was confirmed this works.

evgeni:
Won't this clash with the redis service for pulpcore (in the setups with pulp)?

ekohl:
There we use DB 8 (https://github.com/theforeman/puppet-pulpcore/blob/ef0762aacb44d455fd1e2f06725f92c1ed63b896/manifests/init.pp#L221) so using DB 0 shouldn't clash. Dynflow uses DB 6 (by default):
https://github.com/theforeman/puppet-foreman/blob/f68bb4a183f3c8907147d8d465d74b6ca37a9930/manifests/config.pp#L14

evgeni:
Okay, this just sounds like magic numbers to me, but I guess that's OK for now? :)

ekohl:
They certainly are magic. Perhaps it's something we should document, but not sure how/where.

evgeni (approved):
I guess tests would be cool? ;-)

ekohl:
Now you made me file an issue to monitor the cache. But here are at least some basic tests.

ekohl:
@evgeni enough tests now?

evgeni (approved):
There can never be enough tests ;)

ekohl:
Patches welcome? ;)

PR #1110, opened 2023-02-09
Bump puppetlabs/apache to < 10.0.0

gcoxmoz:
https://github.com/puppetlabs/puppetlabs-apache/commit/6130c0e9ddb8bc0d50d69d49b545f34c2762430a rev'ed that module to 9.0.0, making metadata.json here be behind the times.

I don't know if there's knock-on effects / things you need to do before landing this, but, wanted to at least note the disconnect.

### DIFF
--- a/metadata.json
+++ b/metadata.json
@@ -18,7 +18,7 @@
     },
     {
       "name": "puppetlabs/apache",
-      "version_requirement": ">= 8.0.0 < 9.0.0"
+      "version_requirement": ">= 8.0.0 < 10.0.0"
     },
     {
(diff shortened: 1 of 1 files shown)


ekohl:
We're in the process of branching Foreman 3.6. After I've released RC1 I'll do an org wide check for compatibility.

ekohl (approved):
Note to self: https://github.com/theforeman/foreman-installer/blob/3757e220fe190aaf8e7c1d9c6bbf8f884ad8669c/Puppetfile#L4 has to be modified as well as some other modules.

PR #1111, opened 2023-02-13
Fixes #36090 - Support REX cockpit removal

ekohl:
This doesn't deal with the database, but cockpit is a special plugin in that it's a subpackage of the regular Remote Execution plugin.

By default it uses `undef` so you can still set the version globally as well.

While this plugin doesn't use a config file via `foreman::plugin`, I looked at the source and decided to fix that too.

### DIFF
--- a/manifests/plugin/remote_execution/cockpit.pp
+++ b/manifests/plugin/remote_execution/cockpit.pp
@@ -7,10 +7,17 @@
 # $origins::             Specify additional Cockpit Origins to configure cockpit.conf.
 #                        The $foreman_url is included by default.
 #
+# === Advanced parameters:
+#
+# $ensure::              Specify the package state, or absent to remove it
+#
 class foreman::plugin::remote_execution::cockpit (
+  Optional[String[1]] $ensure = undef,
   Array[Stdlib::HTTPUrl] $origins = [],
 ) {
-  require foreman::plugin::remote_execution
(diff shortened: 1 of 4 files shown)


manifests/plugin/remote_execution/cockpit.pp
```diff
@@ -25,41 +32,56 @@
...
-  foreman::plugin { 'remote_execution-cockpit': }
-  -> service { 'foreman-cockpit':
-    ensure => running,
-    enable => true,
+  foreman::plugin { 'remote_execution-cockpit':
+    version => $ensure,
   }
 
+  if $ensure != 'absent' {
+    service { 'foreman-cockpit':
```
ehelms, line 40: Why doesn't this leave the service running?

ekohl: I didn't verify it, but I'm counting on packaging to drop it. I think this should cover it:
https://github.com/theforeman/foreman-packaging/blob/6656cf0adbbd5097ed16fa039301bcb56090a42f/packages/plugins/rubygem-foreman_remote_execution/rubygem-foreman_remote_execution.spec#L117-L118

adamruzicka:
What's the status here?

ekohl:
Still waiting for a review.

ehelms (approved):

PR #1112, opened 2023-02-21
Release 22.2.0

ehelms:


CHANGELOG.md
```diff
@@ -1,5 +1,15 @@
 # Changelog
 
+## [22.2.0](https://github.com/theforeman/puppet-foreman/tree/22.2.0) (2023-02-21)
+
+[Full Changelog](https://github.com/theforeman/puppet-foreman/compare/22.1.2...22.2.0)
+
+**Implemented enhancements:**
+
+- Fixes [\#36037](https://projects.theforeman.org/issues/36037) - Manage Redis service for Redis cache [\#1109](https://github.com/theforeman/puppet-foreman/pull/1109) ([ekohl](https://github.com/ekohl))
```
ekohl, line 9: This does change the behavior since it now uses a different port by default. I'm not too worried about it, but if we do cherry pick it to Foreman 3.5.2 it may be good to mention that in the release notes.

ekohl (approved):

CHANGELOG.md
ehelms: My goal was to include this release in 3.5.2, so let's do the release note.

PR #1115, opened 2023-04-13
Refs #36319 - Add fog_proxmox plugin support

maximiliankolb:
In foreman-documentation, we're in the process of adding documentation on how to use Proxmox as compute resource provider in Foreman. See https://github.com/theforeman/foreman-documentation/pull/2142#issuecomment-1505340750

### DIFF
--- a/manifests/plugin/proxmox.pp
+++ b/manifests/plugin/proxmox.pp
@@ -1,0 +1,5 @@
+# = Foreman Proxmox plugin
+class foreman::plugin::proxmox {
+  foreman::plugin { 'fog_proxmox':
+  }
+}
--- a/spec/classes/plugin/proxmox_spec.rb
+++ b/spec/classes/plugin/proxmox_spec.rb
@@ -1,0 +1,5 @@
+require 'spec_helper'
+
+describe 'foreman::plugin::proxmox' do
+  include_examples 'basic foreman plugin tests', 'fog_proxmox'
+end


manifests/plugin/proxmox.pp
```diff
@@ -0,0 +1,5 @@
+# = Foreman Proxmox plugin
+class foreman::plugin::proxmox {
+  foreman::plugin { 'proxmox':
```
ekohl, line 3: I think the actual package is called `fog_proxmox` so it would probably fail to really install. Right now we don't have a test with actual package (RPM/deb) installations, so it's hard to verify with automated tests now.

ekohl (commented):
Technically this commit doesn't add it to the installer, but it does open the path. So I'd make this commit `Add fog_proxmox plugin support`.

You could also use a Redmine issue (https://projects.theforeman.org/projects/puppet-foreman/issues/new) and then make this one `Refs #NNN - Add fog_proxmox plugin support` and the one in foreman-installer `Fixes #NNN - ...`. That makes it show up in the changelogs and easier if we decide to cherry pick it to a stable release.

maximiliankolb:
Thanks @ekohl 

I have opened an issue and referenced it: https://projects.theforeman.org/issues/36319

spec/classes/plugin/proxmox_spec.rb
```diff
@@ -0,0 +1,5 @@
+require 'spec_helper'
+
+describe 'foreman::plugin::proxmox' do
+  include_examples 'basic foreman plugin tests', 'proxmox'
```
ekohl, line 4:
```suggestion
  include_examples 'basic foreman plugin tests', 'fog_proxmox'
```

ekohl (changes requested):

spec/classes/plugin/proxmox_spec.rb
maximiliankolb: Applied!

maximiliankolb: Applied!

ekohl:
Puppet 6 failures are unrelated.

ekohl (approved):

PR #1121, opened 2023-06-02
allow puppet/redis 9.x

evgeni:
the breaking changes are dropping Debian 9 and Puppet 6, which is fine for us


evgeni:
no idea why unit tests started failing, but they also do without this PR and I don't see why

PR #1122, opened 2023-06-21
Fixes #36645 - Change the default Redis cache DB to 4

ekohl:
On a typical Katello installation Dynflow uses DB 6 and Pulp uses DB 8. Other applications may also use the default DB. Using DB 4 should not conflict.

@evgeni this is something you raised yesterday.

This probably deserves a release note. We should also consider cleaning up the old DB. By default Redis is configured for persistence so it should remain even after a restart. According to the internet using `redis-cli -n 0 flushdb` should do that. I don't think it's something we want to do automatically. If users don't do this, they will expire eventually anyway and the cache is not that large.

### DIFF
--- a/README.md
+++ b/README.md
@@ -40,7 +40,7 @@ module using the parameter `rails_cache_store`. The parameter takes a hash
 containing the type and options specfic to the backend.
 
 The default is the file backend, configured via `{'type' => 'file'}`. To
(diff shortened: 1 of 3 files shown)


evgeni:
I like that!

ehelms:
What was the issue with using `0`?

evgeni:
PCP is also using 0 by default

ekohl:
Now with a Redmine issue for a changelog entry.

evgeni (approved):

PR #1130, opened 2023-09-20
correct sendmail configuration

evgeni:
the `email_` prefix is wrong, the settings have no such prefix

Fixes: 6c902a4f1f484137ebccafe4d3390881fd0b3d61


ekohl (approved):
I wonder if the tests started to fail due to a refactor in puppetlabs-postgresql, but that's not related now.

evgeni:
Yeah, I didn't look too closely, but it didn't seem related.

PR #1134, opened 2023-10-05
Fixes #36801: Make Redis the default cache type

ehelms:


ehelms:
Related is adding to status API but that should not block this change just enhance it -- https://github.com/theforeman/foreman/pull/9793

spec/classes/foreman_spec.rb
```diff
@@ -451,8 +451,6 @@
       describe 'with custom redis' do
         context 'with redis_url' do
           let(:params) { super().merge(dynflow_redis_url: 'redis://127.0.0.1:4333/') }
```
ekohl, line 453: You're removing all assertions in this test, so it's no longer doing anything. I'd still like to test that we can still run without Redis on the same server. You can do that by specifying the Redis URL:
```suggestion
          let(:params) { super().merge(dynflow_redis_url: 'redis://127.0.0.1:4333/', rails_cache_store: {type: 'redis', 'urls': '127.0.0.1:4334}) }
```

ekohl (changes requested):

ekohl (approved):

PR #1135, opened 2023-10-09
Fixes #36812 - allow setting (fc)ct_location

evgeni:


manifests/init.pp
```diff
@@ -182,6 +182,10 @@
 #
 # $register_in_foreman::          Register host in Foreman
 #
+# $provisioning_ct_location::     The location of the binary to call when transpiling CoreOS templates.
```
evgeni, line 185: I wasn't sure we really need the `provisioning` prefix here

templates/settings.yaml.erb
```diff
@@ -114,3 +114,10 @@
     :namespace: foreman
 <%   end -%>
 <% end -%>
+
```
ekohl, line 117: This will result in an empty line at the end and restarts for everyone. Can we avoid this?

evgeni: by teaching Puppet to diff YAML by content, and not by checksum? sure ;-)

other than that: there is surely a weird `<%--+-!` invocation to force this, but I can never remember how exactly in ERB.
I can also just move the next line one up

evgeni: also, will it really? the line above is `<% end -%>` and I would think the `-%>` should eat up any empty newlines?

ekohl (approved):

PR #1140, opened 2023-11-13
Include settings header via concat

ekohl:
Rather than including this directly via ERB (which raises a deprecation warning), this uses more native methods.


evgeni (approved):

ekohl:
CentOS 8 timed out (which is more common than I'd like), but this is part of https://github.com/theforeman/puppet-foreman/pull/1139 and that passed as well as Debian 11 & 20.04 so I'm merging this.

PR #1142, opened 2023-11-13
Use JSON to parse Foreman API responses

ekohl:
The PSON parser is deprecated and Foreman doesn't actually send PSON; it sends JSON.

See https://www.puppet.com/docs/puppet/8/http_api/pson.html


evgeni (approved):

ekohl:
CentOS 8 timed out (which is more common than I'd like), but this is part of https://github.com/theforeman/puppet-foreman/pull/1139 and that passed as well as Debian 11 & 20.04 so I'm merging this.

PR #1145, opened 2023-11-22
Add hammer-cli-foreman-rh-cloud package

ShimShtein:


ekohl (changes requested):
Please add it to https://github.com/theforeman/puppet-foreman/blob/master/spec/classes/cli_plugins_spec.rb & https://github.com/theforeman/puppet-foreman/blob/master/spec/acceptance/foreman_cli_plugins_spec.rb so it gets tested.

spec/acceptance/foreman_cli_plugins_spec.rb
```diff
@@ -28,6 +28,7 @@ class { 'foreman::cli':
         include foreman::cli::tasks
         include foreman::cli::templates
         include foreman::cli::webhooks
+        include foreman::cli::rh_cloud
```
ekohl, line 31: This is only packaged for Red Hat, so I'd expect it to be similar to azure, kubevirt and openscap. However, it's even only packaged for Katello, whcih we don't set up. In that sense I suppose it's similar to `foreman::cli::katello` that we don't have acceptance tests for it. Just unit tests.

ekohl (changes requested):

ShimShtein:
@ekohl Refactored the acceptance tests to incorporate the plugin the same way Katello does.
The tests are all green now!

ekohl (approved):
Thanks! https://github.com/theforeman/foreman-plugin-overview/blob/014863c694f1d1642ed8751fa193a9c3a7a1b04a/config.yaml#L394 also needs an update after this, though perhaps only after we add it to `foreman-installer`.

PR #1151, opened 2024-02-16
Ruby 3 - Replace YAML.load with YAML.safe_load

tuxmea:


ekohl:
I wonder why this is needed. In Ruby 3 load and safe_load are the same. Is it just to get the same behavior on both 2 & 3?

ekohl:
@tuxmea could you take a look at my question?

PR #1154, opened 2024-02-23
Fixes #37211 - Add ProxyPass for /images to support avatars

dgoetz:
See https://community.theforeman.org/t/user-avatar-support-in-the-foreman-keep-it-or-drop-it/32271

### DIFF
--- a/manifests/config/apache.pp
+++ b/manifests/config/apache.pp
@@ -99,7 +99,7 @@ class foreman::config::apache (
   Pattern['^(https?|unix)://'] $proxy_backend = 'unix:///run/foreman.sock',
   Boolean $proxy_add_headers = true,
   Hash $proxy_params = { 'retry' => '0' },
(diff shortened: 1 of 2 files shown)


ekohl (commented):
Digging through the source I can see we set this as the avatar path: https://github.com/theforeman/foreman/blob/c31702337015ecdf76aaa3053924b57007b8867f/app/models/auth_sources/auth_source_ldap.rb#L238-L240

And interestingly enough, we don't use this here because it's private: https://github.com/theforeman/foreman/blob/c31702337015ecdf76aaa3053924b57007b8867f/app/models/user.rb#L291-L292

Can you open a Redmine issue for this?

dgoetz:
Redmine issue was created as https://projects.theforeman.org/issues/37211, adjusted the commit message and force pushed it, adjusted also the title of the PR to reflect this.

ekohl (approved):

PR #1159, opened 2024-03-25
Refs #37296 - Drop foreman-hooks

adamruzicka:


ekohl (changes requested):
I'd say this is `Refs` at most and the PR in foreman-installer.git is the actual fix.

ehelms (approved):

PR #1161, opened 2024-04-24
Fixes #33974 - Change the pool size to threads + 4

ShimShtein:
According to an investigation described in https://community.theforeman.org/t/rails-connection-pool-size-optimizations/36675 foreman process spawns 4 additional threads that consume DB connection during the startup. Hence the amount of acctive DB connections should be the amount of puma threads + 4 additional threads.

### DIFF
--- a/manifests/config.pp
+++ b/manifests/config.pp
@@ -68,7 +68,15 @@ class foreman::config {
     'database'  => $foreman::db_database,
     'username'  => $foreman::db_username,
     'password'  => $foreman::db_password,
-    'db_pool'   => max($foreman::db_pool, $foreman::foreman_service_puma_threads_max),
+    # Set the pool size to at least the amount of puma threads + 4 threads that are spawned automatically by the process.
+    # db_pool is optional, and undef means "use default" and the second part of the max statement will be set.
+    # The number 4 is for 4 threads that are spawned internally during the execution:
+    # 1. Katello event daemon listener
+    # 2. Katello event monitor poller
+    # 3. Stomp listener (required by Katello)
+    # 4. Puma server listener thread
+    # This means for systems without Katello we can reduce the amount of the pool to puma_threads_max + 1
+    'db_pool'   => pick($foreman::db_pool, $foreman::foreman_service_puma_threads_max + 4),
   }
 
   file { '/etc/foreman/database.yml':
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -84,8 +84,8 @@
 #
 # $db_root_cert::                 Root cert used to verify SSL connection to postgres
 #
-# $db_pool::                      Database 'production' size of connection pool. When running as a reverse proxy,
-#                                 the value of `$foreman_service_puma_threads_max` is used if it's higher than `$db_pool`.
+# $db_pool::                      Database 'production' size of connection pool. If the value is not set, it will be
+#                                 set by default to the amount of puma threads + 4 (for internal system threads)
 #
 # $db_manage_rake::               if enabled, will run rake jobs, which depend on the database
 #
@@ -226,7 +226,7 @@ class foreman (
   String[1] $db_password = $foreman::params::db_password,
   Optional[String[1]] $db_sslmode = undef,
   Optional[String[1]] $db_root_cert = undef,
-  Integer[0] $db_pool = 5,
+  Optional[Integer[0]] $db_pool = undef,
   Boolean $db_manage_rake = true,
   Stdlib::Port $server_port = 80,
   Stdlib::Port $server_ssl_port = 443,
--- a/spec/classes/foreman_spec.rb
+++ b/spec/classes/foreman_spec.rb
@@ -199,7 +199,7 @@ describe 'foreman' do
             db_username: 'foreman',
             db_password: 'secret',
             db_sslmode: 'prefer',
-            db_pool: 5,
+            db_pool: 9,
             db_manage_rake: true,
             server_port: 80,
             server_ssl_port: 443,


ShimShtein:
@pablomh Thanks for all the testing and experimenting. 
What do you think, is it enough?

pablomh:
Hi @ShimShtein, according to our data, this change should be very welcome :)

These are the results of our testing that justify this change:

- Chart showing the number of registration errors with different DB pool values when performing concurrent registrations on a "medium" sized Satellite server

image

  - In it we can see where only modifying the DB pool value improves the registration success rate, getting no errors until reaching >200 concurrent registration requests with values of DB pool >= 8.

- Chart showing the number of DB pool exhaustion errors (`could not obtain a connection from the pool within 5.000 seconds (waited 5.000 seconds); all pooled connections were in use`) with different DB pool values when performing concurrent registrations on a "medium" sized Satellite server

image

  - In it we can see how all the DB pool exhaustion errors dissapear with values of DB pool >= 8.

- Chart showing the number of DB pool connection retries appearing in the logs when performing concurrent registrations on a "medium" sized Satellite server

image

  - In it we can see how all the DB pool connection retries dissapear with values of DB pool >= 9.

Based on this testing, we agree that the minimum Foreman DB pool value where all these issues are addressed is 9.

ehelms (approved):
@evgeni / @ekohl any concerns with this approach to changing the pool size?

evgeni (approved):

ShimShtein:
@ehelms Added a migration PR: https://github.com/theforeman/foreman-installer/pull/937

manifests/init.pp
```diff
@@ -226,7 +226,7 @@
   String[1] $db_password = $foreman::params::db_password,
   Optional[String[1]] $db_sslmode = undef,
   Optional[String[1]] $db_root_cert = undef,
-  Integer[0] $db_pool = 5,
+  Integer[0] $db_pool = 9,
```
ekohl, line 229: Should we make this:
```suggestion
  Optional[Integer[0]] $db_pool = undef,
```

That way, if we change it in the future it automatically gets recalculated. We would need to change the part in `config.pp` to deal with it by using `pick()` instead of `max()`.

The DB migration could then simply be `.delete('db_pool') if db_pool == 5`.

ShimShtein:
@ekohl fixed both here and the installer

manifests/config.pp
```diff
@@ -68,7 +68,9 @@
     'database'  => $foreman::db_database,
     'username'  => $foreman::db_username,
     'password'  => $foreman::db_password,
-    'db_pool'   => max($foreman::db_pool, $foreman::foreman_service_puma_threads_max),
+    # Set the pool size to at least the amount of puma threads + 4 threads that are spawned automatically by the process.
```
ekohl, line 71: It would be nice to document that the 4 is just needed for Katello.

manifests/init.pp
```diff
@@ -226,7 +226,7 @@
   String[1] $db_password = $foreman::params::db_password,
   Optional[String[1]] $db_sslmode = undef,
   Optional[String[1]] $db_root_cert = undef,
-  Integer[0] $db_pool = 5,
+  Optional[Integer[0]] $db_pool = undef,
```
ekohl, line 229: On line 87 there's documentation for this parameter that's shown in the installer when using `--full-help`. Perhaps you can update it to describe the behavior.

manifests/config.pp
```diff
@@ -68,7 +68,9 @@
...
+    # db_pool is optional, and undef means "use default" and the second part of the max statement will be set
+    'db_pool'   => max(pick($foreman::db_pool, 0), $foreman::foreman_service_puma_threads_max + 4),
```
ekohl, line 73: I'd keep this simple so you still provide the user with full control. For example if the `+ 4` part does give problems you still provide a way to override it.
```suggestion
    'db_pool'   => pick($foreman::db_pool, $foreman::foreman_service_puma_threads_max + 4),
```

manifests/init.pp
```diff
@@ -84,8 +84,8 @@
 #
 # $db_root_cert::                 Root cert used to verify SSL connection to postgres
 #
-# $db_pool::                      Database 'production' size of connection pool. When running as a reverse proxy,
-#                                 the value of `$foreman_service_puma_threads_max` is used if it's higher than `$db_pool`.
+# $db_pool::                      Database 'production' size of connection pool. If the value is not set, it will be
+#                                 set by default to the amount of puma threads + 4 (for threads that are spawn by the system)
```
ekohl, line 88: Took me a while to figure this out. Would this be clearer?
```suggestion
#                                 set by default to the amount of puma threads + 4 (for internal system threads)
```

ekohl (commented):
@evgeni Debian 12 is failing:
```
   current directory:
  /usr/share/foreman/vendor/ruby/3.1.0/gems/psych-5.1.2/ext/psych
  /usr/bin/ruby3.1 -I /usr/lib/ruby/vendor_ruby -r
  ./siteconf20240512-4799-xgqy0g.rb extconf.rb
  checking for yaml.h... no
  yaml.h not found
  *** extconf.rb failed ***
  Could not create Makefile due to some reason, probably lack of necessary
  libraries and/or headers.  Check the mkmf.log file for more details.  You may
  need configuration options.
```

But I thought https://github.com/theforeman/foreman-packaging/commit/60435ac4e9de0c5b30cc092b278abfa4a1b1dec3 would address that.

ShimShtein:
Should be a bit more clear now.

ekohl (approved):

PR #1164, opened 2024-05-14
run foreman in a container

evgeni:


manifests/service.pp
```diff
@@ -35,14 +35,38 @@
...
+      service_entry   => {
+        'TimeoutStartSec' => '900',
+      },
+      container_entry => {
+        'Image'         => 'quay.io/evgeni/foreman-rpm:latest',
+        'Volume'        => ['/etc/foreman/:/etc/foreman/'],
+        'AddCapability' => ['CAP_DAC_OVERRIDE', 'CAP_IPC_OWNER'],
+        'Network'       => 'host',
+        'HostName'      => $foreman::servername,
+        'Notify'        => true,
```
evgeni, line 64: `sd_notify` from the container, super nice.

manifests/service.pp
```diff
@@ -35,14 +35,38 @@
...
+      unit_entry      => {
+        'Description' => 'Foreman',
+      },
+      service_entry   => {
+        'TimeoutStartSec' => '900',
+      },
+      container_entry => {
+        'Image'         => 'quay.io/evgeni/foreman-rpm:latest',
+        'Volume'        => ['/etc/foreman/:/etc/foreman/'],
+        'AddCapability' => ['CAP_DAC_OVERRIDE', 'CAP_IPC_OWNER'],
```
evgeni, line 61: the `foreman` user inside the container might have a different UID than outside, fucking up file access :/

(I am not sure this is overall a good idea, but here we are for now)

manifests/config.pp
```diff
@@ -153,7 +155,13 @@
   }
 
   if $foreman::apache {
-    $listen_socket = '/run/foreman.sock'
+    if $foreman::deployment_mode == 'container' {
+      $listen_socket = 'localhost:3000/'
```
evgeni, line 159: This one was tricky. Apache (or `foreman::config::apache`) doesn't add a trailing slash to the backend, which then tries to access things like `http://localhost:3000users/login`, obviously failing

spec/acceptance/foreman_basic_spec.rb
```diff
@@ -55,4 +55,24 @@ class { 'foreman':
...
+  context 'in a Container' do
+    before(:context) { purge_foreman }
+    describe 'in a Container' do
+      it_behaves_like 'an idempotent resource' do
+        let(:manifest) do
+          <<~PUPPET
+          class { 'foreman':
+            deployment_mode => 'container',
+            db_host => 'localhost',
+            db_manage_rake => false,
```
evgeni, line 69: I wonder if these two could be made more "dynamic"?

manifests/service.pp
ekohl: Feels like a nasty hack that we need to resolve prior to going production with this.

manifests/config.pp
```diff
@@ -98,11 +98,13 @@
     )
   )
   $min_puma_threads = pick($foreman::foreman_service_puma_threads_min, $foreman::foreman_service_puma_threads_max)
-  systemd::dropin_file { 'foreman-service':
-    filename       => 'installer.conf',
-    unit           => "${foreman::foreman_service}.service",
-    content        => template('foreman/foreman.service-overrides.erb'),
-    notify_service => true,
+  if $foreman::deployment_mode == 'package' {
```
ekohl, line 101: Does it makes sense to make the deployment mode ensure absent/present?

evgeni: Yes, that's my next step :)

spec/acceptance/foreman_basic_spec.rb
ekohl: What do you mean? That `db_manage_rake` would be `undef` by default and the deployment mode determines the value?

evgeni: yeah, same for host (there is not really a socket in the container)

manifests/config.pp
ekohl: You can use systemd socket activation with podman containers. I've used that myself with gunicorn to avoid needing to deal with a firewall (and instead deal with SELinux). Any reason you don't use that now?

See https://github.com/containers/podman/blob/main/docs/tutorials/socket_activation.md#socket-activation-of-containers for more info.

evgeni: Oh cool. I didn't try that (yet), but I also think that the above is a legit bug in our current deployment if for some reason users do not want sockets.

also lol @ "and instead deal with SELinux" ;)

PR #1170, opened 2024-06-17
always log foreman-rake output

evgeni:
this is useful when debugging which migrations ran and how long they
took

### DIFF
--- a/manifests/rake.pp
+++ b/manifests/rake.pp
@@ -22,7 +22,7 @@ define foreman::rake (
     command     => "/usr/sbin/foreman-rake ${title}",
     user        => $user,
     environment => sort(join_keys_to_values({ 'HOME' => $app_root } + $environment, '=')),
-    logoutput   => 'on_failure',
+    logoutput   => true,
     refreshonly => $unless =~ Undef,
     timeout     => $timeout,
     unless      => $unless,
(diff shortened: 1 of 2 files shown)


manifests/rake.pp
```diff
@@ -22,7 +22,7 @@
     command     => "/usr/sbin/foreman-rake ${title}",
     user        => $user,
     environment => sort(join_keys_to_values({ 'HOME' => $app_root } + $environment, '=')),
-    logoutput   => 'on_failure',
```
evgeni, line 25: This was set to `on_failure` (which is default, btw) since the original implementation in 88e4c16403df1c55376de7ff7dd83420a7a000af

ekohl (approved):
I was debating whether this can log any secrets, like the initial password with seeding. That particular case isn't an issue since it's not printed when provided (https://github.com/theforeman/foreman/blob/93a16bdbd7a7a819bab9d7a136dee9986b30bc08/db/seeds.d/035-admin.rb#L59-L65).

I can't think of anything else, but have you thought about that?

evgeni:
I did so, briefly. In the installer context, "this is fine", as we secure the log properly. Outside, I couldn't find anything but the seeded password which you either WANT logged (when you do not set it) or it's not printed.

Migrations shouldn't print anything secret.

PR #1171, opened 2024-07-12
Drop foreman::compute::gce

ekohl:
With Foreman 3.5 this package was dropped. foreman::plugin::google is the replacement.


evgeni (approved):

ekohl:
Merging since we already have another backwards incompatible change.

PR #1172, opened 2024-07-25
Convert Apache IPA authentication template to EPP

ekohl:
This moves the IPA authentication from 4 separate ERB files to a single EPP file. This gives better guarantees because of data types. It's also easier to maintain a single file that gives the complete overview of the feature.

It's moved to `foreman::config::apache` together with the inclusion of required modules. This does make the `http_keytab` variable a bit uglier.

The sssd service definition is moved to avoid duplication. The `ipa_manage_sssd` block already existed and now the service lives together with the config file management.


templates/apache_ipa_authentication.epp
```diff
@@ -0,0 +1,69 @@
+<%|
+  String[1] $pam_service,
+  Stdlib::Absolutepath $keytab,
+  Boolean $gssapi_local_name,
+  Boolean $ipa_authentication_api,
+-%>
```
kenyon, line 6:
```suggestion
|-%>
```

PR #1173, opened 2024-08-08
Fix a typo in hammer_root.yml.epp

asteflova:
This fixes a small typo in `hammer_root.yml.epp`. (It was previously fixed in https://github.com/theforeman/hammer-cli-foreman/commit/11c2bf5dc37427a435898850f17372b060f0e83b.)


ekohl (changes requested):
Tests fail because we also have the statement there:
https://github.com/theforeman/puppet-foreman/blob/5f38479c870ef03a4a1565007be68b61e8950d53/spec/classes/foreman_cli_spec.rb#L55

ekohl (changes requested):
Looks like you missed one so the tests are still failing

asteflova:
Thanks @ekohl and sorry for taking so long, I was going to take a look later this week.

PR #1175, opened 2024-08-13
Allow forcing fact upload when host already exists

evgeni:


lib/puppet/provider/foreman_host/rest_v3.rb
```diff
@@ -2,7 +2,7 @@
   confine :feature => [:json, :oauth]
 
   def exists?
-    !id.nil?
+    !id.nil? && !resource[:force]
```
ekohl, line 5: I'd swap it since `id` ends up doing the HTTP call while `resource[:force]` is just reading a property.
```suggestion
    !resource[:force] && !id.nil?
```

PR #1185, opened 2024-09-18
Fixes #37761 - use ProxyPass and upgrade=websocket where possible

evgeni:
RewriteRules need special handling of some characters (esp "?"), which
    differs based on Apache version. Instead of going down that way, we can
    switch to using ProxyPass as proxying is the only thing we really need
    here, at least for HTTP.
    
    For WebSockets, we need to allow protocol upgrades, which modern
    (2.4.47+) Apache can do itself via "ProxyPass … upgrade=websocket".
    For older Apache (esp on EL8 and Ubuntu 20.04), we keep the RewriteRules
    in place. To be removed when we drop support for those targets.
    
    Co-Authored-By: Adam Ruzicka <aruzicka@redhat.com>

### DIFF
--- a/manifests/config/apache.pp
+++ b/manifests/config/apache.pp
@@ -167,8 +167,9 @@ class foreman::config::apache (
   if $suburi {
     $custom_fragment = undef
   } else {
-    # mod_env and mod_expires are required by configuration in _assets.conf.erb
+    # mod_env, mod_expires and mod_rewrite are required by configuration in _assets.conf.erb
     include apache::mod::env
+    include apache::mod::rewrite
     # apache::mod::expires pulls in a config file we don't want, like apache::default_mods
     # It uses ensure_resource to be compatible with both $apache::default_mods set to true and false
     include apache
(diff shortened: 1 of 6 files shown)


manifests/config/apache.pp
```diff
@@ -182,8 +183,23 @@
     order   => '03',
   }
 
-  include apache::mod::proxy_wstunnel
-  $websockets_backend = regsubst($_proxy_backend, 'http://', 'ws://')
+  # mod_proxy supports "ProxyPass ... upgrade=websocket" since 2.4.47
+  # EL8: 2.4.37 / EL9: 2.4.62 / Debian11: 2.4.62 / Ubuntu20.04: 2.4.41 / Ubuntu22.04: 2.4.52
+  $proxy_upgrade_websocket = !($facts['os']['family'] == 'RedHat' and $facts['os']['release']['major'] == '8') and !($facts['os']['name'] == 'Ubuntu' and $facts['os']['release']['major'] == '20.04')
+  if $proxy_upgrade_websocket {
+    $vhost_rewrites = []
```
evgeni, line 190: Should this be `undef` instead?

manifests/config/apache.pp
```diff
@@ -167,8 +167,9 @@
   if $suburi {
     $custom_fragment = undef
   } else {
-    # mod_env and mod_expires are required by configuration in _assets.conf.erb
+    # mod_env, mod_expires and mod_rewrite are required by configuration in _assets.conf.erb
     include apache::mod::env
+    include apache::mod::rewrite
```
evgeni, line 172: amazing what bugs you find once you disable other parts of the code ;)

ehelms:
I tried testing this, but I think it's user error,. I just run into "Sorry" when I try to look at the web console.

evgeni:
can you post a screenshot with the browser console open to the network tab (and after a reload, as otherwise the tab is empty)

ehelms:
> can you post a screenshot with the browser console open to the network tab (and after a reload, as otherwise the tab is empty)

image

ehelms:
> Eh, you have no cockpit on the target machine, or no ssh set up

I said user error. Cockpit is definitely set up. Let me double check SSH. Also, this error output sucks.

evgeni:
```
foreman-installer --enable-foreman-plugin-remote-execution --enable-foreman-plugin-remote-execution-cockpit --enable-foreman-proxy-plugin-remote-execution-script --foreman-proxy-plugin-remote-execution-script-cockpit-integration true --foreman-proxy-plugin-remote-execution-script-install-key true
```

this should make cockpit work against the foreman box

and yes, please complain to cockpit-ws, we do not own that

ehelms (approved):
This works for me, thanks @evgeni !

evgeni:
updated commit message etc.

PR #1187, opened 2024-10-14
add support for plugin `ensure` stage *purged*

UiP9AV6Y:
the package resource type supports the following values for the `ensure` parameter:

*    present
*    absent
*    purged
*    disabled
*    installed
*    latest
*    /./

this change adds support for *purged* as it has the same effect in the context of this module.


ekohl (approved):
Plugin removal is always a bit tricky, but REX cockpit is a one of those exceptions where it makes sense and the code reads well.

PR #1188, opened 2024-10-14
add support for managing individual plugin versions

UiP9AV6Y:
currently on the  remote_execution/cockpit plugin class has support for managing the plugin version. all other plugins are tied to the `foreman::plugin_version`. this change adds the same `ensure` parameter to all plugin classes allowing for granular state management.


UiP9AV6Y:
i kept the parameter documentation in the same style for each class, as opposed to using `@param` everywhere, as it will make it easier in the future to assimilate the documentation

PR #1190, opened 2024-10-16
try db_password to be Sensitive

evgeni:
- **properly escape quotes in passwords by calling to_ruby**
- **try db_password to be Sensitive**

### DIFF
--- a/templates/database.yml.epp
+++ b/templates/database.yml.epp
@@ -30,6 +30,6 @@
   username: <%= $username %>
 <% } -%>
 <% if $password { -%>
-  password: "<%= $password %>"
+  password: <%= stdlib::to_ruby($password) %>
 <% } -%>
   pool: <%= $db_pool %>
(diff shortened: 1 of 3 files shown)


templates/database.yml.epp
```diff
@@ -30,6 +30,6 @@
   username: <%= $username %>
 <% } -%>
 <% if $password { -%>
-  password: "<%= $password %>"
+  password: <%= stdlib::to_ruby($password) %>
```
evgeni, line 33: If I drop this, I get:
```
expected that the catalogue would contain File[/etc/foreman/database.yml] with content set to /password: "secret"/ but it is set to #<Sensitive [value redacted]>
```

so maybe `with_content` is a bad idea to begin with?

evgeni: `with_content(sensitive(…))` works for the plain case, but not once to_ruby is involved. yay

Issue #1191, opened 2024-10-22
Dynflow Sidekiq service files do not get updated if `dynflow_redis_url` has been set 


highpingblorg:
### The issue

If a user would set a custom Redis URL via `dynflow_redis_url` that is not `redis://localhost:6379/6`, the Dynflow orchestrator and worker(s) will attempt to connect to the default URL which is set when installing the package `foreman-dynflow-sidekiq`.

In the current state, implementing Redis authentication with foreman is impossible without manually modifying the state of the needed service file, which in return makes the configuration not idempotent. The solution will allow for better Redis configuration for Dynflow Sidekiq.

### The solution

Ensure that puppet is managing the `dynflow-sidekiq@.service` file and modifying the correct setting so the Redis URL is updated.

Both the orchestrator and worker(s) service files are a symlink to the static file  `dynflow-sidekiq@.service` that is installed with the `foreman-dynflow-sidekiq` package.

ekohl:
@adamruzicka are you aware of this issue and can we solve it in a nicer way?

adamruzicka:
Apparently redis url can be configured from within a rails initializer so we might be able to switch to that and have settings.yaml be the only source of truth. That being said, I'm not sure how well would interleaving the initialization of rails, dynflow and sidekiq play out. Also, by losing the option to configure dynflow-sidekiq instances separately we would close the door to having multi-executor-world deployments, which noone really does, but which is technically possible as of now.

PR #1192, opened 2024-10-22
Dynflow Sidekiq service file can be managed via puppet

highpingblorg:
Fixes issue #1191


ekohl (changes requested):
Loading the full definition is not something I'd want. Systemd has drop in files you can place. Why can't they be used to override only what you need?

highpingblorg:
@ekohl Modified my implementation to use systemd drop in files as requested.

If there is anything else I should change, let me know!

PR #1195, opened 2024-11-13
add db_extra_options parameter

### LINKED ISSUE #1194: Allow more more database.yml settings
We need to be able to set the `target_session_attrs` parameter in `database.yml` (which is a connection parameter for libpq), but this is currently not supported. I'm open to making a PR, but unsure if there is a preference for adding a `String $db_target_session_attrs` parameter, or just a generic `Hash[String, String] $db_extra_options` parameter, which we can then use to set any option.
I am aware that `target_session_attrs` is not documented in foreman or rails docs, but from https://github.com/rails/rails/blob/main/activerecord/lib/active_record/connection_adapters/postgresql_adapter.rb#L47 , any parameter not documented there, will be passed as libpq connection parameter.

dmaes:
This PR adds a `db_extra_options` parameter, to allow adding custom parameters in the database.yml config.

Fixes #1194


dmaes:
@ekohl @evgeni anything still missing to get this reviewed and merged?

templates/database.yml.epp
```diff
@@ -33,3 +34,6 @@
   password: <%= stdlib::to_ruby($password) %>
 <% } -%>
   pool: <%= $db_pool %>
+<% $extra_options.each |String $k, String $v| { -%>
+  <%= $k %>: <%= $v %>
```
ekohl, line 38: Should this also use `stdlib::to_ruby` to ensure things are quoted if needed? I don't trust unquoted strings with random input.

templates/database.yml.epp
```diff
@@ -33,3 +34,6 @@
   password: <%= stdlib::to_ruby($password) %>
 <% } -%>
   pool: <%= $db_pool %>
+<% $extra_options.each |String $k, String $v| { -%>
```
ekohl, line 37: I'd trust the header a the top to enforce the type and avoid duplication.
```suggestion
<% $extra_options.each |$k, $v| { -%>
```

manifests/init.pp
```diff
@@ -230,6 +233,7 @@
   Optional[String[1]] $db_sslmode = undef,
   Optional[String[1]] $db_root_cert = undef,
   Optional[Integer[0]] $db_pool = undef,
+  Hash[String, String] $db_extra_options = {},
```
ekohl, line 236: Are there options that are numbers? So something like this:
```suggestion
  Hash[String, Variant[String, Integer]] $db_extra_options = {},
```

dmaes:
@ekohl thanks for the feedback, I implemented the suggestions

PR #1197, opened 2024-12-11
Fixes #38085 - Cap the number of Puma workers at 100

ekohl:
Users can easily run out of PostgreSQL connections on large deployments.  This provides a ceiling to the automatically calculated Puma workers because it's unclear if that really provides performance benefits while it can break deployments.

Not noticable here, but foreman-installer defaults to max 500 PostgreSQL connections and has tuning options that to increase it to 1000.

Ideally it would be smarter, but it's really hard to account for all possible conditions.

### DIFF
--- a/manifests/config.pp
+++ b/manifests/config.pp
@@ -86,12 +86,14 @@ class foreman::config {
     content => epp('foreman/database.yml.epp', $db_context),
   }
 
+  # Capped at 100 because worst case that's 100 * (5 + 4) = 900 PostgreSQL connections for Katello or 100 * 5 = 500 for vanilla Foreman
   # CPU based calculation is based on https://github.com/puma/puma/blob/master/docs/deployment.md#mri
   # Memory based calculation is based on https://docs.gitlab.com/ee/install/requirements.html#puma-settings
   $puma_workers = pick(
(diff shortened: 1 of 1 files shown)


bastelfreak (approved):

bastelfreak (approved):

ekohl:
I really should have tested this locally, but third time is the charm.

manifests/config.pp
```diff
@@ -86,12 +86,14 @@
     content => epp('foreman/database.yml.epp', $db_context),
   }
 
+  # Capped at 100 because worst case that's 100 * (5 + 4) = 900 PostgreSQL connections for Katello or 100 * 5 = 500 for vanilla Foreman
```
ehelms, line 89: Given our XXL tuning is set to 1000, should we go lower than even 900 connections? (https://github.com/theforeman/foreman-installer/blob/develop/config/foreman.hiera/tuning/sizes/extra-extra-large.yaml#L8) Given we have other parts of the deployment also having to access Postgres.

ekohl: I wasn't quite sure and this was really the nice round number that sort of looked OK. I also think we can increase the XXL tuning above 1000.

ehelms (approved):

PR #1201, opened 2025-01-09
Add oidc config parameters

WoutResseler:
This PR adds a oidc parameters, to allow enabling oidc authentication with for example keycloak completely from config as code. The parameters get added to settings.yml config.

### DIFF
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -329,7 +329,7 @@ class foreman (
   Boolean $register_in_foreman = true,
   Optional[Stdlib::Absolutepath] $provisioning_ct_location = undef,
   Optional[Stdlib::Absolutepath] $provisioning_fcct_location = undef,
-  Boolean $authorize_login_delegation = $keycloak,
+  Boolean $authorize_login_delegation = false,
   String[1] $authorize_login_delegation_auth_source_user_autocreate = 'External',
   Optional[String[1]] $login_delegation_logout_url = undef,
   Optional[String[1]] $oidc_jwks_url = undef,


ekohl (commented):
Exposing these makes sense to me. In hindsight I'm not happy with how we (I) have implemented Keycloak integration. It relies a lot on `keycloak-httpd-client-install` and that's problematic. This lower level "expose the parameters" approach is IMHO better, but I have some implementation notes.

I'd prefer to avoid exposing those `authorize_login_delegation`. At first I thought about a more specific `Boolean $oidc`, but prefer a more generic auth parameter like `Enum['internal', 'ipa', 'keycloak', 'oidc'] $auth = 'internal'`. We have places where we could then use a `case` statement to handle it. This makes it obvious to users they need to select a single auth mechanism.

This is where we handle external auth on the  Foreman side:
https://github.com/theforeman/puppet-foreman/blob/dbfdc94aaf2adaa4c8826617ea53c7755dfb421c/manifests/config.pp#L188-L283

And this is where we configure Apache:
https://github.com/theforeman/puppet-foreman/blob/dbfdc94aaf2adaa4c8826617ea53c7755dfb421c/manifests/config/apache.pp#L239-L264
As you can see, this part relies on `keycloak-http-client-install` but I'd prefer to configure Apache's mod_auth_openidc ourselves and I think the parameters you expose can do that. You can do so by passing oidc parameters to the vhost: https://github.com/puppetlabs/puppetlabs-apache/blob/97449e4a0a9f395a2765a3100336070e8298f1dc/manifests/vhost.pp#L1649-L1653 and see https://github.com/puppetlabs/puppetlabs-apache/blob/main/types/oidcsettings.pp for the specific types. You'd probably add them very similar to `vhost_https_internal_options`.

In https://github.com/theforeman/puppet-foreman/pull/1172 I started some refactoring in the auth area that you might find interesting as well. In general I want to move to EPP templates instead of ERB but haven't fully pursued  it.

Lastly, I would really like to see some tests before we merge this, but I can imagine you first want to get the design right to avoid rewriting those all the time.

ekohl:
cc @adamruzicka this might interest you

ekohl:
An incomplete version of what I suggested is in https://github.com/theforeman/puppet-foreman/pull/1203.

PR #1202, opened 2025-01-09
Add parameter outofsync_interval

WoutResseler:
Add the parameter `outofsync_interval` to make it configurable from puppet and over writable from possible hiera config

### DIFF
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -329,7 +329,7 @@ class foreman (
   Boolean $register_in_foreman = true,
   Optional[Stdlib::Absolutepath] $provisioning_ct_location = undef,
   Optional[Stdlib::Absolutepath] $provisioning_fcct_location = undef,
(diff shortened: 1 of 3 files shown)


ekohl:
This is IMHO a duplicate of https://github.com/theforeman/puppet-foreman/pull/1007. It's true you can't set it via Hiera and it's using the CLI so it won't end up as a read-only setting, but I'd prefer to avoid adding all settings as parameters on `init.pp`. Perhaps we can add it to [foreman::settings](https://github.com/theforeman/puppet-foreman/blob/master/manifests/settings.pp) so you can set it via Hiera? That could use [foreman::settings_fragment](https://github.com/theforeman/puppet-foreman/blob/master/manifests/settings_fragment.pp).

WoutResseler:
@ekohl I moved the configuration as requested to [foreman::settings](https://github.com/theforeman/puppet-foreman/blob/master/manifests/settings.pp).

PR #1204, opened 2025-01-16
Add parameter to enable iop-advisor-engine to rh_cloud plugin

ehelms:


manifests/plugin/rh_cloud.pp
```diff
@@ -1,5 +1,11 @@
 # Installs rh_cloud plugin
-class foreman::plugin::rh_cloud {
+class foreman::plugin::rh_cloud (
+  Boolean $enable_iop_advisor_engine = false,
+) {
   foreman::plugin { 'rh_cloud':
   }
+
+  if $enable_iop_advisor_engine {
```
ekohl, line 8: Is there a way to disable this? So an `else` branch?

manifests/plugin/rh_cloud.pp
```diff
@@ -1,5 +1,11 @@
 # Installs rh_cloud plugin
-class foreman::plugin::rh_cloud {
+class foreman::plugin::rh_cloud (
+  Boolean $enable_iop_advisor_engine = false,
+) {
   foreman::plugin { 'rh_cloud':
```
ekohl, line 5: https://github.com/theforeman/foreman_rh_cloud/pull/932 introduces settings. IMHO this should create a `/etc/foreman/settings.plugins.d/${plugin}.yaml` file with that which can be done easily with the existing `config` parameter. The following keeps it simple and avoids a template by using `stdlib::to_yaml`:
```suggestion
  $config = {
    ':foreman_rh_cloud' => {
      ':use_local_advisor_engine' => $enable_iop_advisor_engine,
    },
  }

  foreman::plugin { 'rh_cloud':
    config => stdlib::to_yaml($config),
```

I'm not 100% sure on the use of symbols vs strings though. That could break.

ekohl (changes requested):
Can we add an acceptance test?

evgeni:
> Can we add an acceptance test?

rh_cloud needs Katello, so an acceptance test seems complicated to me?

manifests/plugin/rh_cloud.pp
```diff
@@ -1,5 +1,11 @@
 # Installs rh_cloud plugin
```
evgeni, line 1:
```suggestion
# @summary Installs rh_cloud plugin
#
# @param enable_iop_advisor_engine
#   Enable iop-advisor-engine integration
#
```

ekohl:
> rh_cloud needs Katello, so an acceptance test seems complicated to me?

If it's only about installing the package, it may be OK?+ I'd think we can still do a DB migration and start the service without candlepin present since that sounds similar to what we do in CI.

We already have some code to set up the repo:
https://github.com/theforeman/puppet-foreman/blob/dbfdc94aaf2adaa4c8826617ea53c7755dfb421c/spec/acceptance/foreman_cli_plugins_spec.rb#L66-L69

manifests/plugin/rh_cloud.pp
```diff
@@ -1,5 +1,15 @@
...
+#
+class foreman::plugin::rh_cloud (
+  Boolean $enable_iop_advisor_engine = false,
+) {
   foreman::plugin { 'rh_cloud':
   }
+
+  if $enable_iop_advisor_engine {
+    include iop_advisor_engine
+  }
```
evgeni, line 14:
```suggestion
  class { 'iop_advisor_engine':
    ensure => bool2str($enable_iop_advisor_engine, 'present', 'absent')
  }
```
like that?

ekohl: If we can implement that then I do think it is nice for users.

evgeni: https://github.com/theforeman/puppet-iop_advisor_engine/commit/5ae274d8aca70479f4f131c730b9d7c71a554fe5 just sneaked that in :)

manifests/plugin/rh_cloud.pp
```diff
@@ -1,5 +1,15 @@
...
+#   Enable iop-advisor-engine integration
+#
+class foreman::plugin::rh_cloud (
+  Boolean $enable_iop_advisor_engine = false,
+) {
   foreman::plugin { 'rh_cloud':
   }
+
+  class { 'iop_advisor_engine':
+    ensure => bool2str($enable_iop_advisor_engine, 'present', 'absent')
```
evgeni, line 13:
```suggestion
    ensure => bool2str($enable_iop_advisor_engine, 'present', 'absent'),
```

.fixtures.yml
```diff
@@ -3,9 +3,15 @@ fixtures:
     apache: 'https://github.com/puppetlabs/puppetlabs-apache'
     apt: 'https://github.com/puppetlabs/puppetlabs-apt'
     augeas_core: 'https://github.com/puppetlabs/puppetlabs-augeas_core'
+    certs:
+      repo: "https://github.com/ehelms/puppet-certs.git"
+      branch: "advisor-certs"
```
ekohl, line 8: https://github.com/theforeman/puppet-certs/pull/474

.fixtures.yml
```diff
@@ -3,9 +3,13 @@ fixtures:
     apache: 'https://github.com/puppetlabs/puppetlabs-apache'
     apt: 'https://github.com/puppetlabs/puppetlabs-apt'
     augeas_core: 'https://github.com/puppetlabs/puppetlabs-augeas_core'
+    certs: 'https://github.com/theforeman/puppet-certs'
     concat: 'https://github.com/puppetlabs/puppetlabs-concat'
     cron_core: 'https://github.com/puppetlabs/puppetlabs-cron_core'
     extlib: 'https://github.com/voxpupuli/puppet-extlib'
+    hashfile: 'https://github.com/southalc/hashfile'
+    iop_advisor_engine: 'https://github.com/theforeman/puppet-iop_advisor_engine'
```
evgeni, line 11: Should `iop_advisor_engine` become a real dependency of the foreman module, or "soft", given we only need it in the rh cloud case?

evgeni: https://github.com/theforeman/foreman-installer/pull/1005

spec/acceptance/foreman_rh_cloud_spec.rb
```diff
@@ -6,6 +6,14 @@
   it_behaves_like 'an idempotent resource' do
     let(:manifest) do
       <<-PUPPET
+      file { '/etc/foreman-proxy':
+        ensure => directory,
+      }
+
+      group { 'foreman-proxy':
+        ensure => present,
+      }
```
ekohl, line 15: Why do we need a foreman-proxy in puppet-foreman? That feels like a horrible violation of principles where Foreman should never read files from foreman-proxy. Is this because of certs?

ekohl: I already mentioned before that I feel advisor engine belongs closer to a Smart Proxy because that's where all our external integration lives. We really need to figure out external services in a consistent way, especially if we expect more services to be added. Perhaps our Smart Proxy model isn't scalable, but I've made a mental note we need to discuss this. For now a note above it in the acceptance test is a "good enough" to allow us to proceed?

evgeni: Yeah, we're abusing a bit of the SP architecture, even tho I still think it's not *actually* a SP thing (e.g. you can't have multiple of it, which breaks the SP paradigm).

I'll add a note to the test, yes.

spec/acceptance/foreman_rh_cloud_spec.rb
```diff
@@ -6,6 +6,16 @@
   it_behaves_like 'an idempotent resource' do
     let(:manifest) do
       <<-PUPPET
+      # iop_advisor_engine reqiures foreman-proxy certs to talk back to Foreman
```
ekohl, line 9:
```suggestion
      # iop_advisor_engine requires foreman-proxy certs to talk back to Foreman
```

manifests/plugin/rh_cloud.pp
evgeni:
```
# cat /etc/foreman/plugins/foreman_rh_cloud.yaml 
---
":foreman_rh_cloud":
  ":use_local_advisor_engine": true
```

:rage: 

Why are we using symbols again?

ekohl: Because it makes life interesting? I wonder if we should enhance https://github.com/theforeman/foreman/blob/9e5da95e3194f37b71f013794835fd1d0f9a4a30/config/settings.rb#L40 to symbolize keys.

evgeni: maybe… but I am not shooting at that can of worms today.

evgeni:
Not beautiful, but works :)

ekohl (approved):
I don't kind merging this. IMHO the 2 issues are structural and not limited to this PR.

I'm hopeful that if we fix the service discovery issue properly we don't even need the config file and the symbols in YAML issue disappears

PR #1206, opened 2025-01-23
add rh_cloud acceptance test

evgeni:


spec/acceptance/foreman_rh_cloud_spec.rb
```diff
@@ -0,0 +1,25 @@
...
+    let(:manifest) do
+      <<-PUPPET
+      yumrepo { 'katello':
+        baseurl  => "http://yum.theforeman.org/katello/nightly/katello/el${facts['os']['release']['major']}/x86_64/",
+        gpgcheck => 0,
+      }
+      yumrepo { 'candlepin':
+        baseurl  => "https://yum.theforeman.org/candlepin/4.4/el${facts['os']['release']['major']}/x86_64/",
+        gpgcheck => 0,
+      }
```
ekohl, line 16: Given the repetition I'm starting to wonder if we should have a common place for this. Perhaps use [katello::repo](https://github.com/theforeman/puppet-katello/blob/master/manifests/repo.pp), though that still doesn't give you candlepin.

In the past I've thought about expanding `foreman::repo` with this, but always got stuck on the mismatch in repo version. Once we finish rpm.t.o where the versions are aligned this should be easier.

Another thing to keep in mind is that this test leaves the repository enabled, so you're already in a weird state where it may or may not be enabled during a test. If you take that logic to the extreme, it may be OK to enabled it in the spec helper at the start of the run.

evgeni: yeah, I thought so too -- spec helper seems fine enogh.

ekohl: > moved to spec helper, as the acceptance tests here are not to prove our repoclosure ;-)

There is some aspect to this, but the installer also installs some packages in combinations that repoclosure can't quite understand. Still, I'm good with it. I even thought about moving all Katello packages into plugins to get away from the whole problem.

spec/acceptance/foreman_rh_cloud_spec.rb
```diff
@@ -0,0 +1,16 @@
+require 'spec_helper_acceptance'
+
+describe 'Scenario: install foreman with rh_cloud', if: os[:family] == 'redhat' do
```
ekohl, line 3: I still wonder if this should be `redhat` or a list? Look at the other code that is `['debian', 'ubuntu'].include?(os[:family])`.

evgeni: It gets executed on both CentOS Stream and Alma (I verified that), so 🤷‍♀️

ekohl: This is the reason we drink ;)

PR #1210, opened 2025-02-13
Add Settings Item to foreman function

C24-AK:
## This PR is providing the ability to change setting values in foreman.

if item => settings is provided, the code is now calling
`/api/settings/<search-param>`
instead of
`/api/hosts/?search=<search-param>`

### What can you do with it?
We can now change foreman settings on the puppet run without doing it manually.
foreman::foreman('settings', 'authorize_login_delegation', $per_page, $url, $admin_user, $admin_password, $timeout, { 'value' => true })

This is calling a PUT Request instead of a GET Request to the foreman api to change and persist settings

### DIFF
--- a/lib/puppet/functions/foreman/foreman.rb
+++ b/lib/puppet/functions/foreman/foreman.rb
@@ -11,8 +11,9 @@
 #        foreman_user => 'my_api_foreman_user'
 #        foreman_pass => 'my_api_foreman_pass'
 #
-# 'item' may be: environments, fact_values, hosts, hostgroups, puppetclasses, smart_proxies, subnets
+# 'item' may be: environments, fact_values, hosts, hostgroups, puppetclasses, smart_proxies, subnets, settings
 # 'search' is your actual search query.
+#          if item is 'settings' search is the setting name and will be a PUT Request
 # 'per_page' specifies the maximum number of results you'd like to receive.
 #            This defaults to '20' which is consistent with what you'd get from
 #            Foreman if you didn't specify anything.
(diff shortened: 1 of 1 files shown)


lib/puppet/functions/foreman/foreman.rb
```diff
@@ -62,9 +63,19 @@ def foreman(item, search, per_page = "20", foreman_url = "https://localhost", fo
...
     begin
-      path = "/api/#{CGI.escape(item)}?search=#{CGI.escape(search)}&per_page=#{CGI.escape(per_page.to_s)}"
+      if item == 'settings'
+        path = "/api/#{CGI.escape(item)}/#{CGI.escape(search)}"
+      else
+        path = "/api/#{CGI.escape(item)}?search=#{CGI.escape(search)}&per_page=#{CGI.escape(per_page.to_s)}"
+      end
+
+      if item == 'settings' && filter_result.is_a?(Hash)
+        req = Net::HTTP::Put.new(path)
```
ekohl, line 73: This is now how the function is supposed to behave. It's a lookup, not a resource.

ekohl (changes requested):
This whole design goes against how you manage resources in Foreman. I think you want the [foreman_config_entry](https://github.com/theforeman/puppet-foreman/blob/master/lib/puppet/type/foreman_config_entry.rb) type. Right now we only have a [cli](https://github.com/theforeman/puppet-foreman/blob/master/lib/puppet/provider/foreman_config_entry/cli.rb) provider for it, but you could write one that uses the REST API.

You can also enforce things by writing them out in YAML files. That way the user can't even change them in the UI because they're presented as read only.

lib/puppet/functions/foreman/foreman.rb
C24-AK: Ohh I didn't know this exists, I will try it and close the PR if successful! 
Thank you

ekohl:
> You can also enforce things by writing them out in YAML files. That way the user can't even change them in the UI because they're presented as read only.

I should have linked [foreman::settings_fragment](https://github.com/theforeman/puppet-foreman/blob/master/manifests/settings_fragment.pp).

PR #1214, opened 2025-02-19
use dashes, not underscores when settings HTTP headers

evgeni:
underscores are not allowed in HTTP headers and some backends reject
headers using underscores before converting incoming dashes to
underscores to pass to the app

### DIFF
--- a/manifests/config/apache.pp
+++ b/manifests/config/apache.pp
@@ -184,10 +184,10 @@ class foreman::config::apache (
   }
 
   $vhost_http_request_headers = [
-    'set X_FORWARDED_PROTO "http"',
-    'set SSL_CLIENT_S_DN ""',
-    'set SSL_CLIENT_CERT ""',
-    'set SSL_CLIENT_VERIFY ""',
+    'set X-FORWARDED-PROTO "http"',
+    'set SSL-CLIENT-S-DN ""',
+    'set SSL-CLIENT-CERT ""',
+    'set SSL-CLIENT-VERIFY ""',
   ] +
   $request_headers_to_unset.map |$header| {
     "unset ${header}"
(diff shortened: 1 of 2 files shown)


manifests/config/apache.pp
```diff
@@ -212,10 +212,10 @@
   }
 
   $vhost_https_request_headers = [
-    'set X_FORWARDED_PROTO "https"',
-    'set SSL_CLIENT_S_DN "%{SSL_CLIENT_S_DN}s"',
-    'set SSL_CLIENT_CERT "%{SSL_CLIENT_CERT}s"',
-    'set SSL_CLIENT_VERIFY "%{SSL_CLIENT_VERIFY}s"',
+    'set X-FORWARDED-PROTO "https"',
```
ekohl, line 215: Will this break https://github.com/theforeman/puppet-pulpcore/blob/8288f3736a1543417eba6322e7013d0f230bd85e/templates/settings.py.erb#L65-L66?

evgeni: No, it will make it work!

`My-Cool-Header` gets translated to `HTTP_MY_COOL_HEADER` by gunicorn. It's just in the past it also allowed `My_Cool_Header` to become `HTTP_MY_COOL_HEADER` (you remember, huh?) and this used to work for us, and now it doesn't anymore.

evgeni: Also, puppet-pulpcore should really not depend on puppet-foreman setting things up -- it can be deployed standalone after all :)

ekohl: I see you came to the same conclusion.

evgeni: https://github.com/theforeman/puppet-pulpcore/pull/376

ekohl (approved):

PR #1221, opened 2025-04-23
Ensure glibc-langpack-en is always installed on EL

evgeni:


manifests/database/postgresql.pp
```diff
@@ -7,11 +7,19 @@
 
   include postgresql::client, postgresql::server
 
+  if $facts['os']['family'] == 'RedHat' {
+    stdlib::ensure_packages(['glibc-langpack-en'])
+    $require_langpack = Package['glibc-langpack-en']
+  } else {
+    $require_langpack = undef
```
ekohl, line 14: I think this also works and is shorter:
```suggestion
    Package['glibc-langpack-en'] -> Postgresql::Server::Db[$foreman::db_database]
```

Then you can leave out the `require` below

ekohl (approved):

PR #1230, opened 2025-08-05
Pin puppet-apt since it dropped Puppet 7

ehelms:
Since we pull the main branch for testing, and it was updated to drop Puppet 7 support Debian and Ubuntu testing fails:

https://github.com/puppetlabs/puppetlabs-apt/commit/3f9afe6a4840a5804a740d924f6fde0b277db8e0

### DIFF
--- a/.fixtures.yml
+++ b/.fixtures.yml
@@ -1,7 +1,9 @@
 fixtures:
   repositories:
     apache: 'https://github.com/puppetlabs/puppetlabs-apache'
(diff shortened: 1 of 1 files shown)


ekohl:
No, it's common practice in the Puppet world to point to git versions when running unit tests. We actually rely on that in some cases when we depend on unreleased versions.

Dependabot isn't even capable of doing so and if it did, you would spend a LOT more time merging updating and polluting your git logs.

PR #1284, opened 2026-06-10
Fixes #39415 - Warn about unrecognized smart proxy features

adamruzicka:
Companion to https://github.com/theforeman/foreman/pull/11017

### DIFF
--- a/lib/puppet/provider/foreman_smartproxy/rest_v3.rb
+++ b/lib/puppet/provider/foreman_smartproxy/rest_v3.rb
@@ -33,7 +33,12 @@ Puppet::Type.type(:foreman_smartproxy).provide(:rest_v3, :parent => Puppet::Type
       raise Puppet::Error.new(error_string)
     end
 
-    validate_features!(resource[:features], features_list(JSON.load(r.body)))
+    body = JSON.load(r.body)
+    unless body
+      raise Puppet::Error.new("Proxy #{resource[:name]}: Foreman API returned empty response body after creating proxy")
+    end
+
+    validate_features!(body)
   end
 
   def destroy
@@ -84,17 +89,30 @@ Puppet::Type.type(:foreman_smartproxy).provide(:rest_v3, :parent => Puppet::Type
     # Replace proxy/feature list cache: pre-#19476 versions have limited responses, clear cache and re-fetch for them
     @proxy = body.key?('features') ? body : nil
 
-    validate_features!(resource[:features], features_list(proxy))
+    validate_features!(proxy) if proxy
   end
 
   private
 
   def features_list(proxy)
-    proxy['features'].map { |ft| ft['name'] }.sort
+    features = proxy['features']
+    unless features.is_a?(Array)
+      Puppet.err("Proxy #{resource[:name]}: Expected 'features' to be an array, got #{features.class}")
+      return []
+    end
+    features.map { |ft| ft['name'] }.compact.sort
   end
 
-  def validate_features!(expected, actual)
-    missing_features = expected - actual
-    raise Puppet::Error.new("Proxy #{resource[:name]} has failed to load one or more features (#{missing_features.join(", ")}), check /var/log/foreman-proxy/proxy.log for configuration errors") unless missing_features.empty?
+  def validate_features!(proxy_data)
+    unrecognized = Array(proxy_data['unrecognized_features'])
+    if unrecognized.any?
+      Puppet.warning("Proxy #{resource[:name]} has features not recognized by Foreman: #{unrecognized.join(', ')}. If these features come from a Smart Proxy plugin, make sure Foreman has the plugin installed too.")
+    end
+
+    actual = features_list(proxy_data)
+    missing = resource[:features] - actual
+    unless missing.empty?
+      raise Puppet::Error.new("Proxy #{resource[:name]} has failed to load one or more features (#{missing.join(', ')}), check /var/log/foreman-proxy/proxy.log for configuration errors")
+    end
   end
 end
--- a/spec/unit/foreman_smartproxy_rest_v3_spec.rb
+++ b/spec/unit/foreman_smartproxy_rest_v3_spec.rb
@@ -1,5 +1,43 @@
 require 'spec_helper'
 
+shared_examples 'unrecognized features handling' do
+  it 'warns about unrecognized features' do
+    stub_proxy_response('features' => [{'name' => 'TFTP'}, {'name' => 'Logs'}], 'unrecognized_features' => ['NewFeature'])
+    expect(Puppet).to receive(:warning).with(/Proxy proxy.example.com has features not recognized by Foreman: NewFeature/)
+    trigger
+  end
+
+  it 'does not warn when unrecognized features list is empty' do
+    stub_proxy_response('features' => [{'name' => 'TFTP'}, {'name' => 'Logs'}], 'unrecognized_features' => [])
+    expect(Puppet).not_to receive(:warning)
+    trigger
+  end
+
+  it 'does not warn when unrecognized_features key is absent' do
+    stub_proxy_response('features' => [{'name' => 'TFTP'}, {'name' => 'Logs'}])
+    expect(Puppet).not_to receive(:warning)
+    trigger
+  end
+
+  it 'does not crash when unrecognized_features is nil' do
+    stub_proxy_response('features' => [{'name' => 'TFTP'}, {'name' => 'Logs'}], 'unrecognized_features' => nil)
+    expect(Puppet).not_to receive(:warning)
+    expect { trigger }.not_to raise_error
+  end
+
+  it 'warns about multiple unrecognized features' do
+    stub_proxy_response('features' => [{'name' => 'TFTP'}, {'name' => 'Logs'}], 'unrecognized_features' => ['FeatureA', 'FeatureB'])
+    expect(Puppet).to receive(:warning).with(/FeatureA, FeatureB/)
+    trigger
+  end
+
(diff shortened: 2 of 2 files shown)


adamruzicka:
Test failures should be unrealted and should be addressed by https://github.com/theforeman/puppet-foreman/pull/1285

adamruzicka:
Rebased on top of 1285 to see how the tests look like.

lib/puppet/provider/foreman_smartproxy/rest_v3.rb
```diff
@@ -33,7 +33,12 @@ def create
       raise Puppet::Error.new(error_string)
     end
 
-    validate_features!(resource[:features], features_list(JSON.load(r.body)))
+    body = JSON.load(r.body)
+    if body.key?('unrecognized_features')
+      warn_unrecognized_features(body)
+    else
+      validate_features!(resource[:features], features_list(body))
+    end
```
ofedoren, line 39: The `if/else` makes warning and validation mutually exclusive — but they serve different purposes and should both always run.

Looking at foreman#11017, the API now always includes `unrecognized_features` as an attribute on the smart proxy response (via `main.json.rabl`). So once both PRs land, `validate_features!` will **never run** — not just when there are actual unrecognized features. A proxy that fails to load an expected feature (e.g. TFTP) will no longer raise `Puppet::Error`, it will just silently succeed.

`warn_unrecognized_features` already handles the case where the key is absent or the list is empty (returns early), so the guard is unnecessary.

Alternatively, since both methods share the proxy data, are always called together, and both deal with features — it might be worth merging them into a single `validate_features!` that takes the proxy hash directly:

```ruby
def validate_features!(proxy_data)
  unrecognized = proxy_data['unrecognized_features']
  if unrecognized && unrecognized.any?
    Puppet.warning("Proxy #{resource[:name]} has features not recognized by Foreman: #{unrecognized.join(', ')}. If these features come from a Smart Proxy plugin, make sure Foreman has the plugin installed too.")
  end

  actual = features_list(proxy_data)
  missing = resource[:features] - actual
  raise Puppet::Error, "Proxy #{resource[:name]} has failed to load one or more features (#{missing.join(', ')}), check /var/log/foreman-proxy/proxy.log for configuration errors" unless missing.empty?
end
```

That would simplify both call sites to just `validate_features!(body)` / `validate_features!(proxy)` and eliminate `warn_unrecognized_features` entirely. Up to you which approach you prefer.

```suggestion
    body = JSON.load(r.body)
    warn_unrecognized_features(body)
    validate_features!(resource[:features], features_list(body))
```

_Assisted by Claude Code_

lib/puppet/provider/foreman_smartproxy/rest_v3.rb
```diff
@@ -84,7 +89,11 @@ def refresh_features!
     # Replace proxy/feature list cache: pre-#19476 versions have limited responses, clear cache and re-fetch for them
     @proxy = body.key?('features') ? body : nil
 
-    validate_features!(resource[:features], features_list(proxy))
+    if proxy.key?('unrecognized_features')
+      warn_unrecognized_features(proxy)
+    else
+      validate_features!(resource[:features], features_list(proxy))
+    end
```
ofedoren: Same issue as in `create` — when the refresh response includes `features`, `@proxy = body` and that body will always have `unrecognized_features` after foreman#11017, so validation is permanently skipped.

When the refresh response lacks `features` (pre-#19476 proxy), the lazy `proxy` getter re-fetches from `api/v2/smart_proxies`, and that response also always includes `unrecognized_features` — same result.

```suggestion
    warn_unrecognized_features(proxy)
    validate_features!(resource[:features], features_list(proxy))
```

_Assisted by Claude Code_

lib/puppet/provider/foreman_smartproxy/rest_v3.rb
```diff
@@ -97,4 +106,10 @@ def validate_features!(expected, actual)
     missing_features = expected - actual
     raise Puppet::Error.new("Proxy #{resource[:name]} has failed to load one or more features (#{missing_features.join(", ")}), check /var/log/foreman-proxy/proxy.log for configuration errors") unless missing_features.empty?
   end
+
+  def warn_unrecognized_features(proxy_data)
+    unrecognized = proxy_data['unrecognized_features'] if proxy_data
```
ofedoren: Nit: the `if proxy_data` guard is dead code — every call site passes a parsed JSON hash, never nil. The `.nil?` check on `unrecognized` already handles the case where the key doesn't exist (e.g. older Foreman without the column).

```suggestion
    unrecognized = proxy_data['unrecognized_features']
```

_Assisted by Claude Code_

spec/unit/foreman_smartproxy_rest_v3_spec.rb
```diff
@@ -40,6 +40,22 @@
...
       provider.create
     end
+
+    it 'warns about unrecognized features' do
+      expect(provider).to receive(:request).with(:post, 'api/v2/smart_proxies', {}, kind_of(String)).and_return(
+        double(:code => '201', :body => {'features' => [{'name' => 'TFTP'}, {'name' => 'Logs'}], 'unrecognized_features' => ['NewFeature']}.to_json)
+      )
+      expect(Puppet).to receive(:warning).with(/Proxy proxy.example.com has features not recognized by Foreman: NewFeature/)
+      provider.create
+    end
```
ofedoren, line 65: There's no test for the compound case: proxy has unrecognized features **and** is missing an expected feature. With the current if/else logic this would silently pass. After fixing the if/else, consider adding:

```ruby
it 'warns about unrecognized features and still raises on missing expected features' do
  expect(provider).to receive(:request).with(:post, 'api/v2/smart_proxies', {}, kind_of(String)).and_return(
    double(:code => '201', :body => {'features' => [{'name' => 'TFTP'}], 'unrecognized_features' => ['NewFeature']}.to_json)
  )
  expect(Puppet).to receive(:warning).with(/NewFeature/)
  expect { provider.create }.to raise_error(Puppet::Error, /failed to load one or more features \(Logs\)/)
end
```

Same test should be added for `refresh_features!` (both the direct and re-fetch paths).

_Assisted by Claude Code_

spec/unit/foreman_smartproxy_rest_v3_spec.rb
```diff
@@ -128,6 +144,33 @@
...
+      end
+
+      it 'skips feature validation when unrecognized features are present' do
+        expect(provider).to receive(:id).and_return(1)
+        expect(provider).to receive(:request).with(:put, 'api/v2/smart_proxies/1/refresh').and_return(
+          double(:code => '200', :body => {'features' => [{'name' => 'TFTP'}], 'unrecognized_features' => ['Logs']}.to_json)
+        )
+        expect(Puppet).to receive(:warning).with(/Logs/)
+        expect { provider.refresh_features! }.not_to raise_error
+      end
```
ofedoren: This test asserts the current behavior where validation is skipped — but after the fix above (always run both), this test should be updated to expect both a warning **and** a validation error, since the response only includes TFTP but the resource expects `['TFTP', 'Logs']`.

_Assisted by Claude Code_

ofedoren (commented):
It seems https://github.com/theforeman/puppet-foreman/pull/1284/commits/e7dc314c41edeb1632c642e1787f463c205b26ad can be dropped now.

> **Note:** This review was assisted by Claude Code.

lib/puppet/provider/foreman_smartproxy/rest_v3.rb
ofedoren: Personally, I liked more the alternative (merging into one)

lib/puppet/provider/foreman_smartproxy/rest_v3.rb
ofedoren: This mostly depends on what we accept for https://github.com/theforeman/puppet-foreman/pull/1284#discussion_r3436823476

lib/puppet/provider/foreman_smartproxy/rest_v3.rb
adamruzicka: merged

ofedoren:
The core change looks good — all previous feedback addressed. A few non-blocking observations on the new defensive additions:

1. **`validate_features!(proxy) if proxy` in `refresh_features!`** — when the re-fetch also fails to find the proxy (returns nil), validation is silently skipped. Same behavior as master (which would crash on nil), so not a regression, but a deliberate error might be clearer than silent skip.

2. **Type check in `features_list`** — if `features` isn't an array, something is fundamentally broken in the API response. Logging an error and returning `[]` causes `validate_features!` to raise about all expected features being missing, which is misleading for what's really a malformed response. The original `NoMethodError` was arguably more honest.

3. **Nil body guard in `create`** — `JSON.load` on empty/nil input raises `JSON::ParserError`, it doesn't return nil. And the request already passed `success?(r)`, so an empty 201 body is extremely unlikely.

Points 2 and 3 are also unrelated to the PR's purpose (unrecognized features) — might be worth splitting out if you want to keep them.

> **Note:** This review was assisted by Claude Code.

ofedoren (approved):
Thanks, @adamruzicka, I personally don't have anything to add.

https://github.com/theforeman/puppet-foreman/pull/1284#issuecomment-4768952858 is mostly if we'll dig through history later for more context, etc. Not a blocker.

Also, I don't have permissions here to re-run the tests (the failure is probably not related) or to merge here, so just an ACK :)

adamruzicka:
force pushed without any changes just to trigger ci again

ofedoren (approved):
🍏 who can get this in?..

PR #1286, opened 2026-06-17
Add packit

adamruzicka:
Configure packit to build PRs in COPR against foreman-installer, using the foreman packit module workflow that patches foreman-installer to use the local checkout of puppet-foreman for the theforeman/foreman Puppetfile entry.

Based on the same setup in puppet-foreman_proxy.

### DIFF
--- a/.packit.yaml
+++ b/.packit.yaml
@@ -34,7 +34,7 @@ jobs:
       rhel-9:
         additional_repos:
           - https://yum.theforeman.org/releases/nightly/el9/x86_64/
-          - https://yum.puppet.com/puppet8/el/9/x86_64/
+          - https://yum.voxpupuli.org/openvox8/el/9/x86_64/
     module_hotfixes: true
 
 srpm_build_deps:
--- a/.packit.yaml
+++ b/.packit.yaml
@@ -1,0 +1,47 @@
+# See the documentation for more information:
+# https://packit.dev/docs/configuration/
+
+specfile_path: foreman-installer.spec
+
+# add or remove files that should be synced
+files_to_sync:
+  - foreman-installer.spec
+
+# name in upstream package repository or registry (e.g. in PyPI)
+upstream_package_name: foreman-installer
+# downstream (Fedora) RPM package name
+downstream_package_name: foreman-installer
+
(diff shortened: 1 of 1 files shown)


.packit.yaml
```diff
@@ -0,0 +1,48 @@
...
+  create-archive:
+    - bash -c "cd foreman-installer && bundle config set --local path vendor/bundle"
+    - bash -c "cd foreman-installer && bundle config set --local without development:test"
+    - bash -c "cd foreman-installer && bundle install"
+    - bash -c "cd foreman-installer && bundle exec rake pkg:generate_source"
+
+jobs:
+  - job: copr_build
+    trigger: pull_request
+    manual_trigger: true
```
evgeni: feel free to drop that line and always build an rpm

adamruzicka: Dropped

evgeni (approved):

.packit.yaml
```diff
@@ -0,0 +1,48 @@
...
+    - bash -c "cd foreman-installer && bundle exec rake pkg:generate_source"
+
+jobs:
+  - job: copr_build
+    trigger: pull_request
+    targets:
+      rhel-9:
+        additional_repos:
+          - https://yum.theforeman.org/releases/nightly/el9/x86_64/
+          - https://yum.puppet.com/puppet8/el/9/x86_64/
```
ekohl: We may need to update this in more places to use OpenVox. Not a blocker now though.

.packit.yaml
```diff
@@ -0,0 +1,48 @@
...
+    module_hotfixes: true
+
+srpm_build_deps:
+  - wget
+  - make
+  - gcc
+  - libffi-devel
```
ofedoren, line 44: ~if it doesn't build, we probably would need `libyaml-devel` companion~ nevermind...

adamruzicka: Luckily, it builds

.packit.yaml
adamruzicka: Let's not build tech debt, changed.

adamruzicka:
:green_apple:

evgeni:
:ship:

PR #1292, opened 2026-09-17
Fixes #39775 - Add registration admission control via Apache balancer

pablomh:
## Summary

- Adds a `registration_admission_max` parameter (default: 150) to limit
  concurrent connections to `/rhsm` and `/register` endpoints via an
  Apache `mod_proxy_balancer` pool
- When the limit is reached, excess requests queue in Apache's event loop
  instead of overwhelming Puma's thread pool during registration bursts
- Set to 0 to disable admission control entirely

Ports [foremanctl#495](https://github.com/theforeman/foremanctl/pull/495)
to the puppet module world.

## Problem

Under high-concurrency registration, all requests arrive at Puma
simultaneously, creating a deep backlog where tail hosts exceed
subscription-manager's 180s timeout. Without admission control, pass
rates drop sharply above ~760 concurrent registrations on a typical
foreman server.

## Solution

A `<Proxy balancer://foreman-registration>` block with `max=N` on the
`BalancerMember` directive caps how many registration requests reach Puma
concurrently. Excess requests are held in Apache's event loop
(non-blocking) until a slot opens.

The balancer config is injected via additional_includes fragments in both
HTTP and HTTPS vhosts, placed before the catch-all `ProxyPass /`, so
`/rhsm` and `/register` match first while all other traffic (Web UI,
API, Pulp) is unaffected.

## Test plan

- [ ] Verify `satellite-installer --foreman-registration-admission-max 300` applies correctly
- [ ] Confirm `/rhsm` and `/register` are routed through the balancer pool
- [ ] Confirm other endpoints (`/api`, `/hosts`, etc.) are unaffected
- [ ] Verify `--foreman-registration-admission-max 0` disables the balancer entirely
- [ ] Run registration at concurrency levels exceeding the configured max

🤖 Generated with [Claude Code](https://claude.com/claude-code)


jakduch (commented):
Before this leaves draft, I think the mechanism needs one more pass. Apache documents `max` as a per-child-process connection-pool limit; it is not coordinated globally. The values 150/600/800 therefore do not implement the global registration limit described here. Without `acquire` or a balancer wait timeout, an exhausted pool returns `SERVER_BUSY`; `timeout=900` is only the backend I/O timeout.

There is also a lifecycle issue: after changing the parameter from a positive value to `0`, Puppet stops declaring both throttle files but does not remove them, so admission control stays enabled. The disabled branch should manage both files as absent and cover that in tests.

The performance results from foremanctl#495 may still reflect an improvement from explicit worker connection reuse. I would first rerun with `max=1`, `16`, and `600` while observing actual backend connections, then align the design and description with those results.

Apache reference: https://httpd.apache.org/docs/2.4/mod/mod_proxy.html#proxypass

